PR #1, opened 2015-02-11
Convert lorax to use DNF instead of yum

bcl:
This also adds cmdline arguments so that pungi (and other tools) can call lorax directly instead of using it as a library. This allows lorax to move forward with DNF and python3 work without needing to wait on other packages.


vpodzime:
@bcl, could you please merge this as master-dnf?

PR #2, opened 2015-02-27
Move stage2 to images/install.img (#815275)

bcl:
Use the same stage2 location for all arches, put it in images with all
the other images. This only effects boot.iso, live images still use
LiveOS/squashfs.img because that's where dracut's 90dmsquashfs-live
module expects to find it.

For boot.iso anaconda-dracut handles finding stage2, looking at
images/install.img and LiveOS/squashfs.img


vpodzime:
Looks good to me.

PR #3, opened 2015-03-03
Turn off ldconfig

bcl:
We don't need to update ldconfig when running the boot.iso, disable it
to speed up the boot.


vpodzime:
Looks good to me.

PR #4, opened 2015-03-05
Add removekmod template command

bcl:
removekmod GLOB [GLOB...] --allbut KEEPGLOB [KEEPGLOB...]

This can be used to remove kernel modules from under
/lib/modules/*/kernel/ while keeping specific items. This should be
easier than constructing find arguments to select the right things to
save.

### DIFF
--- a/src/pylorax/ltmpl.py
+++ b/src/pylorax/ltmpl.py
@@ -633,36 +633,29 @@ class LoraxTemplateRunner(object):
                 raise ValueError("removekmod needs at least one GLOB after --allbut")
 
             globs = globs[:idx]
-
-            # --allbut needs the full list of files to match keepglobs against, gather up every
-            # file under each glob pattern's results.
-            filelist = set()
-            for g in globs:
-                for top_dir in rglob(self._out("/lib/modules/*/kernel/"+g)):
-                    for root, _dirs, files in os.walk(top_dir):
-                        filelist.update(root+"/"+f for f in files)
-
-            # Remove anything matching keepglobs from the list
-            matches = set()
-            for g in keepglobs:
-                globs_re = re.compile(fnmatch.translate("*"+g+"*"))
-                m = filter(globs_re.match, filelist)
-                if m:
-                    matches.update(m)
-                else:
-                    logger.debug("removekmod %s: no files matched!", g)
-            remove_files = filelist.difference(matches)
         else:
-            # Gather glob matches, the full list isn't needed since it will recursively remove
-            # directory matches.
+            # Nothing to keep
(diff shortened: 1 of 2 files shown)


src/pylorax/ltmpl.py
```diff
@@ -597,12 +597,75 @@ def removefrom(self, pkg, *globs):
...
+
+            globs = globs[:idx]
+
+            # --allbut needs the full list of files to match keepglobs against, gather up every
+            # file under each glob pattern's results.
+            filelist = set()
+            for g in globs:
+                for top_dir in rglob(self._out("/lib/modules/*/kernel/"+g)):
+                    for root, _dirs, files in os.walk(top_dir):
+                        filelist.update(root+"/"+f for f in files)
```
vpodzime: Wouldn't it be better to use the glob a bit more to provide a list of everything that's underneath (i.e. append '**/*' at the end) and then filter out dirs if that's needed instead of doing os.walk()?

src/pylorax/ltmpl.py
```diff
@@ -597,12 +597,75 @@ def removefrom(self, pkg, *globs):
...
+                    logger.debug("removekmod %s: no files matched!", g)
+            remove_files = filelist.difference(matches)
+        else:
+            # Gather glob matches, the full list isn't needed since it will recursively remove
+            # directory matches.
+            keepglobs = []
+            filelist = set()
+            for g in globs:
+                filelist.update(rglob(self._out("/lib/modules/*/kernel/"+g)))
+            remove_files = filelist
```
vpodzime: Maybe I'm missing something important here, but wouldn't it be easier to just construct two sets (remove, keep) from the globs with the "**/*" appended and then do a difference on these two sets followed by filtering out dirs if these should be kept in place?

src/pylorax/ltmpl.py
bcl: No, the directory structure may be several levels deeper, we need to use walk to gather up everything underneath.

src/pylorax/ltmpl.py
bcl: That could work too, walk everything, then apply the keepglob if it exists.

vpodzime: I'm leaving it up to you then to chose an approach that better fits your way of thinking. :)

src/pylorax/ltmpl.py
vpodzime: Hmm, seems like GitHub squashed my glob! I meant appending ``**/*``, but that won't work neither so just ignore this suggestion.

PR #7, opened 2015-03-18
Keep logitech hid drivers (#1199770)

bcl:
Some logitech mice need a driver.


vpodzime:
Looks good to me (for both master and f22-branch).

PR #8, opened 2015-03-19
Add ability for external templates to graft content into boot.iso

cgwalters:
I originally added --add-template to support doing something similar
to pungi, which injects content into the system to be used by default.
However, this causes the content to be part of the squashfs, which
means PXE installations have to download significantly more data that
they may not need (if they actually want to pull the tree data from
the network, which is not an unusual case).

What I actually need is to be able to modify *both* the runtime image
and the arch-specific content.  For the runtime, I need to change
/usr/share/anaconda/interactive-defaults.ks to point to the new
content.  (Although, potentially we could patch Anaconda itself to
auto-detect an ostree repository configured in disk image, similar to
what it does for yum repositories)

For the arch-specfic image, I want to drop my content into the ISO
root.

So this patch adds --add-arch-template and --add-arch-template-var
in order to do the latter, while preserving the --add-template
to affect the runtime image.

Further, the templates will automatically graft in a directory named
"iso-graft/" from the working directory (if it exists).

(I suggest that external templates create a subdirectory named
 "content" to avoid clashes with any future lorax work)

Thus, this will be used by the Atomic Host lorax templates to inject
content/repo, but could be used by e.g. pungi to add content/rpms as
well.

I tried to avoid code deduplication by creating a new template for the
product.img bits and this, but that broke because the parent boot.iso
code needs access to the `${imggraft}` variable.  I think a real fix
here would involve turning the product.img, content/, *and* boot.iso
into a new template.

### DIFF
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -153,7 +153,10 @@ class Lorax(BaseLoraxClass):
             installpkgs=None,
             size=2,
             add_templates=None,
-            add_template_vars=None):
+            add_template_vars=None,
+            add_arch_templates=None,
+            add_arch_template_vars=None,
+            template_tempdir=None):
 
         assert self._configured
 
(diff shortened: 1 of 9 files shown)


cgwalters:
See also https://git.fedorahosted.org/cgit/spin-kickstarts.git/commit/?id=0840fc1394830a3f970239f06cae4a2ef58ed85f

src/pylorax/__init__.py
```diff
@@ -153,7 +153,10 @@ def run(self, dbo, product, version, release, variant="", bugurl="",
             installpkgs=None,
             size=2,
             add_templates=None,
-            add_template_vars=None):
+            add_template_vars=None,
+            add_arch_templates=None,
+            add_arch_template_vars=None,
+            template_tempdir=None):
```
vpodzime: This really is a terribly long list of parameters. Seems to me like we should probably start thinking about grouping some of them into namedtuples or something like that.

bcl:
That's probably a good idea, I like namedtuple vs. something like the **kwarg passing we do in Blivet.

PR #9, opened 2015-03-20
Turn off ldconfig (#1204031)

bcl:
We don't need to update ldconfig when running the boot.iso, disable it
to speed up the boot.

(cherry picked from commit c4c90e0b5ee1f3684cd4beed4f0b935570f897b1)


bcl:
Further experimentation revealed that this causes libedit to crash. So it needs ld.so.conf (master should probably revert the version of this and add ld.so.conf too)

vpodzime:
I'm not much happy about the evolution of this, but it seems like we cannot do any better. The change looks good to me.

PR #12, opened 2015-03-24
Check that the transaction process is still alive

vpodzime:
If it terminates really badly (e.g. with SIGSEGV), it doesn't report any error,
just doesn't put anything to the queue. So instead of just blindly waiting on
the queue forever, check that the process is still alive if we don't get any
message in long time interval.


vpodzime:
Pushed.

PR #14, opened 2015-04-02
Include cryptsetup in the image (#1208214)

bcl:
blivet dropped its requirement on cryptsetup, but fedup needs it in the
upgrade.img and it is useful for rescue mode so add it back.


AdamWill:
FWIW rescue mode actually works OK without this - it can mount encrypted installs, at least, as it uses blivet to do it. Once you have the installed system mounted you can use the cryptsetup it contains.

If for whatever reason you wanted to set up or manipulate an encrypted device from the rescue environment yourself using the cryptsetup tools, though, you could not.

This should certainly fix the upgrade.img case, though - I don't know whether this is the most-correct way to do it or not, but it should certainly work, so +1 from me for that.

PR #15, opened 2015-04-09
Added Sphinx Documentation

bcl:
Run 'make docs' to update the apidocs.


docs/index.rst
```diff
@@ -0,0 +1,27 @@
+.. src documentation master file, created by
+   sphinx-quickstart on Wed Apr  8 13:46:00 2015.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to src's documentation!
```
vpodzime: Do we really want this in the index?

vpodzime:
Other than the question/comment above this looks good to me.

docs/index.rst
bcl: oops, missed a spot. that's what happens when you don't give it a proper name when you run the first pass.

bcl:
pushed

PR #16, opened 2015-04-13
Make 64-bit kernel on 32-bit firmware work for x86 efi machines

vathpela:
This enables Baytrail and similar atom CPUs that typically ship with a
32-bit firmware, but have a 64-bit capable CPU.

(This patch requires a newer shim package that is not ready yet, so don't pull this yet.)

Signed-off-by: Peter Jones <pjones@redhat.com>


vpodzime:
Looks good to me to the extent I understand what it does. :)

lwahonen:
Was this abandoned eventually?

PR #18, opened 2015-05-07
Make sure openssh-clients is installed (#1219398)

bcl:
It used to be pulled in by a dependency, which went away. It should be
explicitly installed.


dashea:
Looks fine

PR #19, opened 2015-05-09
Convert pylorax, lorax and livemedia-creator to Python3

bcl:


vpodzime:
Other than the question/request above these all look good to me.

bcl:
Pushed. gh-pages updated. https://rhinstaller.github.io/lorax/

PR #20, opened 2015-05-18
Fix bug with product DataHolder overwriting product string.

bcl:


clumens:
This works for me locally.

vpodzime:
Looks good to me.

PR #21, opened 2015-05-26
Use imgutils.copytree instead of shutil (#1221878)

bcl:
Sometimes the directory is already there, we just want to copy things
into it, so use our copytree.


vpodzime:
Looks good to me. I always wonder how many projects around the world have such function implemented in their codebase.

PR #23, opened 2015-06-10
Remove some stale entires from runtime-install

dashea:
This doesn't remove all of python2, but it takes care of a good chunk of it.

### DIFF
--- a/share/runtime-install.tmpl
+++ b/share/runtime-install.tmpl
@@ -6,7 +6,6 @@ installpkg anaconda anaconda-widgets
 installpkg kexec-tools-anaconda-addon
 ## anaconda deps that aren't in the RPM
 installpkg tmux
-installpkg iscsi-initiator-utils
 ## Other available payloads
 installpkg dnf
 installpkg rpm-ostree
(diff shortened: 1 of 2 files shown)


share/runtime-install.tmpl
```diff
@@ -91,14 +86,14 @@ installpkg device-mapper-persistent-data
 installpkg xfsdump
 
 ## needed for LUKS escrow
-installpkg python-volume_key volume_key
-installpkg python-nss
+installpkg volume_key
+installpkg nss-tools
```
vpodzime: Why do we need these two? For the kickstart tests? Because other than that libblockdev should correctly require what it needs for LUKS escrow.

share/runtime-install.tmpl
```diff
@@ -91,14 +86,14 @@ installpkg device-mapper-persistent-data
...
 
 ## SELinux support
-installpkg selinux-policy-targeted audit libsemanage-python
+installpkg selinux-policy-targeted audit
 
 ## network tools/servers
-installpkg python-ethtool ethtool openssh-server nfs-utils openssh-clients
+installpkg ethtool openssh-server nfs-utils openssh-clients
```
vpodzime: Do we still need ``ethtool``? I remember somebody mentioning it recently in relation to some deprecated/removed boot option or whatever.

share/runtime-install.tmpl
dashea: And rescue. The nss tools only really make sense for the kickstart test (I think they get pulled in anyway through some dep chain but I don't think we really need to keep the /usr/bin parts), but volume_key would be nice to keep around.

share/runtime-install.tmpl
dashea: The kickstart network command still has a - - ethtool option, and someone might be using ethtool during rescue or %pre

vpodzime:
Fair enough, thanks for the explanation @dashea.

PR #24, opened 2015-06-12
Disable systemd-tmpfiles-clean (#1202545)

bcl:
When the time changes dramatically it cleans up the files in /tmp/
causing problems with the installation.

Resolves: rhbz#1202545


vpodzime:
Wow, this is funny!  (sarcasm) ACK. (for real)

PR #26, opened 2015-06-18
Add ntp configuration file to installation

jkonecny12:
Add package ntp and remove it except ntp.conf file.

Related: rhbz#1085013

### DIFF
--- a/share/runtime-cleanup.tmpl
+++ b/share/runtime-cleanup.tmpl
@@ -45,6 +45,8 @@ removepkg logrotate
 remove /etc/logrotate.d
 ## anaconda needs this to do media check
 removefrom isomd5sum --allbut /usr/bin/checkisomd5
+## remove ntp keep only config file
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me.

bcl:
I don't understand how this solved the problem in the bug. Adding the ntp.conf to the install image isn't going to have any effect on the packages installed to their system.

jkonecny12:
@bcl you must look on main commit for this patch in this [PR](https://github.com/rhinstaller/anaconda/pull/144). 
Here I add configuration file to install environment and in that PR the servers are changed and they are copied to the installed system.

Sorry I mentioned this in the second PR and I forgot to add it here too.

bcl:
Thanks for the explanation, could you add that to the commit before pushing?

jkonecny12:
OK, I will add some information to commit message.

jkonecny12:
Merging this commit.
Thank you all for help.

PR #27, opened 2015-06-22
Explicitly add kernel-modules and kernel-modules-extra

bcl:
When it is left up to dnf to decide how to fulfill the kmod()
requirement from gfs2-utils it will pick kernel-debuginfo-* which adds
about 100M to the size of the iso.

Adding these packages first makes dnf choose them and the iso size is
back down around 450MB


vpodzime:
This seems to me like a bug in DNF -- it should use the smallest package fulfilling the requirements, I think. ACK to this change as a workaround, but we should discuss this with the DNF developers.

PR #29, opened 2015-06-30
Add a bit more overhead to the root filesystem size

bcl:
When this is too small the rootfs can run into problems when used with a
live system. Doubling it leaves enough space for the system to run
properly during the installation and since it's all compresses it
doesn't make the image noticeably bigger.


bcl:
pushed

PR #31, opened 2015-07-15
Add a custom RHEL tagger for tito

bcl:
The default tagger doesn't add Related/Resolves lines to the changelog,
resulting in builds that have to be fixed up manually.

The output from this code still needs to be edited to remove the '-'
prefix from the Related/Resolves lines because tito appends them to
every line of output from the Tagger. But this is better than having to
add them manually.

### DIFF
--- a/rel-eng/lorax_tito/tagger.py
+++ b/rel-eng/lorax_tito/tagger.py
@@ -1,0 +1,118 @@
+import re
+from tito.common import run_command
+from tito.tagger import ReleaseTagger
+
+
+class LoraxRHELTagger(ReleaseTagger):
+    """
+    Tagger which is based on ReleaseTagger and use Red Hat Enterprise Linux
+    format of Changelog:
+    - description
+      Resolves/Related: rhbz#1111
+
+    Used for:
+        - Red Hat Enterprise Linux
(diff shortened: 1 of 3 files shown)


rel-eng/lorax_tito/tagger.py
```diff
@@ -0,0 +1,118 @@
...
+                    tmp = bugstr[len(bug):]
+
+                    for c in tmp:
+                        if not c.isalpha():
+                            tmp = tmp[1:]
+                        else:
+                            break
+
+                    if len(tmp) > 0:
+                        author = tmp
```
vpodzime: It seems to me like the above string operations could be done in a nicer way using the groups or named groups in the regular expression.

vpodzime:
Looks good to me other than the suggestion above.

rel-eng/lorax_tito/tagger.py
bcl: > It seems to me like the above string operations could be done in a nicer way using the groups or named groups in the regular expression.

I agree, but for the moment I want to make sure it matches the anaconda makebumpver code. I'll probably improve both of them at some point.

PR #32, opened 2015-07-17
Create an empty selinux config file (#1243168)

bcl:
In order for selinux to properly label the system it needs to see that
the config file exists.

Also remove the old code trying to copy in a selinux config file, it
never worked -- the removepkg would remove it.


vpodzime:
This is really a stupid bug in *libselinux*. But I think we can live with it as long as it has the comment right above.

bcl:
Pushed.

PR #33, opened 2015-07-20
verify that everything in /usr/bin and /usr/sbin is still runnable after cleanup

dashea:
spice-vdagent, part of the thing that makes copy/paste work between a host and a virt guest, is missing libraries in rawhide. It isn't as noticeable when spice-vdagent breaks as it is, say, when dhclient breaks, so it's probably been broken for a little while. I would like changes like this to be less of a surprise.

I don't have any strong opinions about whether verify should be on or off by default as long as the jenkins job runs with it on.

### DIFF
--- a/share/runtime-cleanup.tmpl
+++ b/share/runtime-cleanup.tmpl
@@ -10,7 +10,8 @@ remove usr/share/i18n
     removepkg perl*
 %endif
 ## no sound support, thanks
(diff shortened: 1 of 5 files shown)


vpodzime:
Nice! Looks good to me.

bcl:
Very nice! Feel free to push when you want.

dashea:
Pushed.

PR #34, opened 2015-07-29
Fix pylorax and livemedia-creator to work with --no-virt and mock installs

bcl:


vpodzime:
Looks good to me otherwise.

PR #35, opened 2015-08-04
Return the output from failed commands in CalledProcessError

bcl:
Some callers expect CalledProcessError.output to have the output, so
pass up the stdout + stderr output.

This means failed runcmd template commands will log to program.log and
lorax.log


dashea:
Looks fine to me

vpodzime:
Very nice line of Python! :)

bcl:
> Very nice line of Python! :)

yeah, yeah, I'm now used to the X or Y shorthand. Still not going to embed if/else though :)

PR #36, opened 2015-08-04
Add a default vconsole.conf to the boot.iso

bcl:


share/config_files/common/vconsole.conf
```diff
@@ -0,0 +1,2 @@
+KEYMAP=us
+FONT=eurlatgr
```
vpodzime: Why this font? I remember using ``latarcyrheb`` in anaconda by default.

M4rtinK: Also with https://github.com/rhinstaller/anaconda/commit/515f0679276369713eaed7d33f86904f9fb062db there should always be a correct font set for the currently selected language. So isn't this change redundant ?

bcl:
This is for the default console font at boot time, and is related to https://fedoraproject.org//wiki/Changes/NewDefaultConsoleFont

Which isn't working right in f23/rawhide. Run 'setfont eurlatgr' and you'll see the font shift. adamw has tested this and confirmed that it fixes the problem.

vpodzime:
Thanks for the explanation! Looks good to me then.

bcl:
pushed.

PR #40, opened 2015-08-13
Change default releasever to 7 (#1253242)

bcl:
7 is more generic than 7.0 and doesn't need to be changed for each
release.

Resolves: rhbz#1253242


vpodzime:
Looks good to me too.

PR #41, opened 2015-08-19
Kill fedup

wgwoods:
These patches make lorax stop building `upgrade.img` and stop requiring `fedup-dracut` and friends. 

For backward-compatibility, the API and CLI have been left unchanged. `--noupgrade` is still accepted and `doupgrade` still gets passed into the templates. It just doesn't do anything.

This is intended for f23-branch and master, so it might need backporting.

_NOTE_: there's one place in `pylorax/__init__.py` that might want some tweaking:
```python
        # ppc64 cannot boot an initrd > 32MiB so remove some drivers
        if self.arch.basearch in ("ppc64", "ppc64le"):
            dracut_args.extend(["--omit-drivers", REMOVE_PPC64_DRIVERS])

            # Only omit dracut modules from the initrd so that they're kept for
            # upgrade.img
            anaconda_args.extend(["--omit", REMOVE_PPC64_MODULES])
```
As far as I can tell, though, that shouldn't hurt anything

### DIFF
--- a/docs/lorax.1
+++ b/docs/lorax.1
@@ -79,9 +79,6 @@ volume id
 \fB\-\-nomacboot\fR
 
 .TP
-\fB\-\-noupgrade\fR
-
-.TP
 \fB\-\-logfile=LOGFILE\fR
 Path to logfile
 
@@ -95,4 +92,3 @@ Martin Gracik
 Will Woods
 Brian C. Lane
 .fi
-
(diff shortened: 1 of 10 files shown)


jkonecny12:
Seems good to me but I think @bcl should see this.

BTW when the funeral will be?

wgwoods:
> BTW when the funeral will be?

Once the packages are officially dead - hopefully when the clock strikes F23 - we'll say a few words, close all the bugs, and then drink a lot of whiskey.

bcl:
Great idea :) I also like how most of the changes are in the templates.

bcl:
pushed.

PR #42, opened 2015-08-20
Add enough of shadow-utils to create new user accounts.

dashea:
Yes user accounts, some account management

### DIFF
--- a/share/runtime-cleanup.tmpl
+++ b/share/runtime-cleanup.tmpl
@@ -37,8 +37,13 @@ append etc/selinux/config ""
 removefrom fedora-release --allbut /etc/os-release /usr/lib/os-release \
                                    /usr/lib/os.release.d/*
 removepkg fedora-release-rawhide
-## no user accounts = no account management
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

bcl:
Why do we need these here? This is for the libuser replacement, right? Shouldn't any tools it is using be run from the installed system's chroot? Sorry for all the ? marks :)

dashea:
> Shouldn't any tools it is using be run from the installed system's chroot?

Sort of. The way I did the anaconda change was to run useradd etc from / with -R <sysroot>, so the tools themselves are doing the chroot instead of anaconda. Granted, we add the sysroot to the path so that could still use the tools installed to the sysroot, but I figured that shadow-utils is probably one of those base packages that the cloud folks don't install to their images, and I did this change to avoid breaking things for that case.

bcl:
Ok, ACK from me as well then, thanks.

dashea:
Pushed to f23 and master

PR #43, opened 2015-08-26
Document --timeout in livemedia-creator man page

atodorov:


docs/livemedia-creator.1
```diff
@@ -184,6 +185,10 @@ volume id
 \fB\-\-squashfs_args SQUASHFS_ARGS\fR
 additional squashfs args
 
+.TP
+\fB\-\-timeout TIMEOUT\fR
+Cancel installer after X minutes
```
vpodzime: *"Terminate the installation"* sounds better to me, but native speakers should probably tell us what is the right wording here.

vpodzime:
Looks good to me otherwise.

docs/livemedia-creator.1
atodorov: I've taken this from --help description. Will update both if wording changes.

bcl:
merged, thanks.

PR #44, opened 2015-08-26
Replace the metacity theme file.

dashea:
This replaces the Adwaita metacity-theme-2.xml file with a version that
sets has_title=False and removes the title borders for maximized
windows. This allows anaconda to run as a maximized window instead of a
fullscreen window while still running without a titlebar.

Related: rhbz#1231856


bcl:
Bleh. If we must.

PR #45, opened 2015-08-26
Update lmc docs and use the same size for / across installation types.

bcl:


vpodzime:
Looks good to me now that I understand it better. :)

bcl:
pushed

PR #46, opened 2015-08-27
Run spice-vdagentd without systemd-logind integration

dashea:
The system-wide spice guest agent, spice-vdagentd, normally uses
systemd-logind to determine which of the per-session spice-vdagent
instances it should be communicating with. On the non-live media, the X
session isn't registered with systemd-logind, so instead start
spice-vdagentd with -X to disable the systemd-logind integration and
serve instead to a single spice-vdagent.

(cherry picked from commit 96ecdd4e6eafd18fa2af2e374cae59c131301984)

Related: rhbz#1169991


vpodzime:
Looks good to me too.

dashea:
Pushed this one a while ago. Oops.

PR #47, opened 2015-08-31
Don't include early microcode in initramfs (#1258498)

bcl:
The system the image boots on will likely not match the host where lorax
was run, and in some cases this can cause systems to hang.

Resolves: rhbz#1258498


vpodzime:
Looks good to me too.

bcl:
pushed

PR #48, opened 2015-09-01
livemedia-creator: Remove random-seed from images (#1258986)

bcl:
systemd uses /var/lib/systemd/random-seed to add entropy to /dev/urandom
at boot time. During image creation this file is created, and if not
removed everything using the image will be adding the same seed.

This is only additional entropy, NOT a seed in the sense of a starting
point for a PRNG, so it will be mixed with other entropy as the system
runs. It isn't a good idea to use the same value everywhere so make sure
it is removed in %post

Resolves: rhbz#1258986

### DIFF
--- a/docs/fedora-minimized.ks
+++ b/docs/fedora-minimized.ks
@@ -37,6 +37,9 @@ part swap --size=1000
 %post
 # Remove root password
 passwd -d root > /dev/null
+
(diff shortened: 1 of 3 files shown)


vpodzime:
Looks good to me.

bcl:
pushed

PR #49, opened 2015-09-08
Add a font that supports Urdu characters (#1004717)

bcl:


vpodzime:
Looks good to me too. I think we should ask the i18n/l10n people to create and maintain a list of fonts that are needed to cover all languages supported by Fedora.

PR #50, opened 2015-09-09
Drop multiprocessing for do_transaction (#1208296)

bcl:
When running the transaction in a separate process it crashes if you use
a https repo source. There's really no need for threads or processes in
lorax so drop it.

Also switched to using the DNF TransactionProgress API for progress
reporting.


jkonecny12:
Seems good to me

vpodzime:
Looks good to me too.

marmarek:
Could this be backported to f23 branch?

PR #53, opened 2015-10-02
Add findmnt command

jkonecny12:
This is nice way how to show existing mounts.

*This wasn't tested - I had some problems with creating compose to test this but the change should be alright*


bcl:
pushed

PR #54, opened 2015-10-02
livemedia-creator: Also generate an iPXE config

cgwalters:
Currently lmc has some ad-hoc code to generate a `PXE_CONFIG`.
In practice we should probably split this out somehow, as in reality
downstream users need to edit it.  So it should be more explicitly
a template.

That aside, it turns out there's also iPXE which in addition to having
the critical `i` prefix, supports http and variable substitution in
the config file itself.

This makes it a lot more convenient to use - downstream users can more
easily edit the `set base-url` and have things work, it's easy to put
an iPXE config into a pastebin etc.

### DIFF
--- a/share/pxe-live/ipxe-config.tmpl
+++ b/share/pxe-live/ipxe-config.tmpl
@@ -1,0 +1,5 @@
+#!ipxe
+set base-url http://example.com/  # You must download the files referenced here and change this URL
+kernel <%text>${base-url}</%text>${kernel} root=live:<%text>${base-url}/</%text>${liveimg} ${addargs}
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me.

bcl:
Good idea, but I'd rather see this setup like --make-appliance and --app-template, let the user specify a template and if they don't use the default.

cgwalters:
The existing `PXE_CONFIG` file from `pxe-config.tmpl` is not a configuration option either, do you think it should be one?

bcl:
Yes. Let's setup reasonable defaults, but let them be overridden.

PR #55, opened 2015-10-02
Reduce the size of macboot.img (#952747)

bcl:
The size estimate was counting the /EFI/BOOT/ contents twice and then
doubling that. Only count things once, then double it for the
System/Library/CoreServices/ copy.

hard-links don't work. With CoreServices hardlinked to /EFI/BOOT/ the
Mac won't boot. With /EFI/BOOT/ hardlinked to CoreServices grub2 cannot
read the config file so there are 2 real copies.

This reduces the image size from 21M to about 12M


vpodzime:
Looks good to me.

PR #56, opened 2015-10-05
Disable systemd-tmpfiles-setup.

dashea:
Most of the files configured by tempfiles are not relevant when an
actual factual /etc is present. Some of the files configured by
tempfiles are actively hostile. Quit letting systemd crud up the
filesystem and let the packages take care of the files that really
matter.

### DIFF
--- a/share/runtime-postinstall.tmpl
+++ b/share/runtime-postinstall.tmpl
@@ -42,7 +42,7 @@ systemctl mask fedora-configure.service fedora-loadmodules.service \
                fedora-autorelabel.service fedora-autorelabel-mark.service \
                fedora-wait-storage.service media.mount \
                systemd-tmpfiles-clean.service systemd-tmpfiles-clean.timer \
-               ldconfig.service
+               ldconfig.service systemd-tmpfiles-setup.service
 
 ## Make logind activate anaconda-shell@.service on switch to empty VT
 symlink anaconda-shell@.service lib/systemd/system/autovt@.service


bcl:
I'd much rather systemd fixed this themselves. Barring that, consider this a very reluctant ack.

vpodzime:
Do you have any particular example of what's going wrong now, @dashea? I'm afraid this change would mean we will soon get to a point where something is not set up correctly and people around systemd will tell us we are disabling something that's supposed to do it.

dashea:
> Do you have any particular example of what's going wrong now, @dashea?

Of course! You may have noticed that the kickstart_tests have not done anything useful in quite some time, and the reason is that all the DNS lookups are failing, because NetworkManager is unable to write to /etc/resolv.conf, because systemd-tempfiles is, again, installing a broken symlink. There's been a bug open since 18 september and the response is basically that this exact behavior is necessary for weirdo embedded corner case stuff that we don't care about, and systemd is not at fault for systemd junking up the system.

> and people around systemd will tell us we are disabling something that's supposed to do it.

Yes they tell us a lot of things.

clumens:
ACK, I am tired of them blocking my work from getting done.

vpodzime:
Considering the extra info from the above, I'm also for just pushing this.

dashea:
https://github.com/rhinstaller/lorax/pull/57 may be more appropriate

PR #58, opened 2015-10-06
Keep libthread so that gdb will work correctly (#1269055)

bcl:
Related: rhbz#1269055


vpodzime:
Looks good to me too.

bcl:
Pushed.

PR #60, opened 2015-10-14
Enable gtk inspector.

dashea:
Gtk turned off the inspector keybindings by default, because they were
interfering with applications that use a lot of complicated keyboard
shortcuts. This is not a concern for anaconda, and the inspector is
pretty handy, so turn it back on.


vpodzime:
Looks good to me too.

dashea:
Pushed.

PR #62, opened 2015-10-21
Make OCI and Vagrant images

bcl:
These commits implement --make-oci and --make-vagrant

OCI is the Open Container Initiative container format (a tar with / at /rootfs/ and a couple json files). Vagrant is a tar of a qcow2 image with a metadata.json file and optional vagrantfile.


vpodzime:
Looks good to me too other than the above question.

PR #64, opened 2015-11-04
Add --virt-uefi to boot the VM using OVMF

bcl:
This requires OVMF to be setup on the system, and for the kickstart to
create a /boot/efi/ partition. You can then use it to create UEFI
bootable partitioned disk images.

### DIFF
--- a/docs/livemedia-creator.rst
+++ b/docs/livemedia-creator.rst
@@ -200,6 +200,12 @@ Example cmdline:
 
 ``sudo livemedia-creator --make-iso --no-virt --ks=./fedora-livemedia.ks``
 
+.. note::
+    Using no-virt to create a partitioned disk image (eg. --make-disk or
+    --make-vagrant) will only create disks usable on the host platform (BIOS
+    or UEFI). You can create BIOS partitioned disk images on UEFI by using
+    virt.
+
 
 AMI Images
 ----------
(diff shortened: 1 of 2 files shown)


jkonecny12:
Looks good to me.

docs/livemedia-creator.rst
```diff
@@ -479,6 +485,34 @@ This also works with ``--no-virt``, but will not work inside a mock due to its
...
+4. Copy /usr/share/edk2.git/ovmf-x64/OVMF_VARS-pure-efi.fd to /usr/share/OVMF/OVMF_VARS.fd
+
+Now you can run livemedia-creator with ``--virt-uefi`` to boot and install using UEFI::
+
+    sudo livemedia-creator --make-disk --virt-uefi --iso=/path/to/boot.iso \
+    --ks=/path/to/fedora-minimal.ks
+
+Make sure that the kickstart you are using creates a /boot/efi partition by including this::
+
+    part /boot/uefi --fstype="uefi" --size=500
```
vpodzime: Should be ``/boot/efi``, shouldn't it?

vpodzime:
Looks good to me too other than the above note/question/

docs/livemedia-creator.rst
bcl: > Should be /boot/efi, shouldn't it? The same applies to the fstype.

Whoops, yes. thanks for catching that.

bcl:
Pushed

PR #68, opened 2015-11-16
livemedia-creator: Check kickstart for shutdown (#1207959)

bcl:
When using virt with livemedia-creator the kickstart needs to include
shutdown to make sure it doesn't hang at the end of the installation.


vpodzime:
Looks good to me.

clumens:
Another option would be that if you have to include the shutdown command but the user doesn't, add it in for them.  Would there be any harm in that?

bcl:
That would complicate the code since the kickstart is passed to the vm, not the ksdata object, it then wouldn't match what they pass to us. I think it's better to encourage them to create a correct kickstart in the first place.

bcl:
pushed to master.

PR #69, opened 2015-11-17
Add support for .repo files (#1264058)

bcl:
This adds the --repo command which can be added multiple times to point
to dnf .repo files.

--enablerepo and --disablerepo can be used multiple times to control
which repos from the .repo files are actually used for the boot.iso
creation.

--repo can be used instead of --source, or in addition to it.

### DIFF
--- a/src/sbin/lorax
+++ b/src/sbin/lorax
@@ -61,7 +61,9 @@ def main(args):
     required.add_argument("-v", "--version", help="version identifier", required=True, metavar="VERSION")
     required.add_argument("-r", "--release", help="release information", required=True, metavar="RELEASE")
     required.add_argument("-s", "--source", help="source repository (may be listed multiple times)",
-                        metavar="REPOSITORY", action="append", default=[], required=True)
+                        metavar="REPOSITORY", action="append", default=[])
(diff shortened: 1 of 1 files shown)


src/sbin/lorax
```diff
@@ -302,6 +325,24 @@ def get_dnf_base_object(installroot, repositories, mirrorlists=None,
...
             return None
 
+    # Enable repos listed on the cmdline
+    for r in enablerepos:
+        repolist = dnfbase.repos.get_matching(r)
+        if not r:
+            log.warning("%s is an unknown repo, not enabling it", r)
+        else:
+            log.info("Enabled repo %s", r)
+            repolist.enable()
```
vpodzime: I think the above two lines should be swapped. Otherwise, one could e.g. get a traceback from ``repolist.enable()`` with *Enabled repo something* already in the log. The same applies to the other for cycle below.

vpodzime:
Looks good to me otherwise.

bcl:
Thanks, that actually cannot happen since it does the lookup first, but I realized I was checking r not repolist like it should have been. fixed that and moved the log anyway.

PR #70, opened 2015-11-20
A couple misc fixes for lmc

bcl:


vpodzime:
These both look good to me.

bcl:
Pushed to master.

PR #71, opened 2015-11-23
Include qemu modules in the initrd

bcl:
This includes, among other things, virtio-rng. Without it the initrd may
take longer to initialize the random non-blocking pool on KVM
installations.


vpodzime:
This looks good to me.

PR #72, opened 2015-12-01
livemedia-creator: Some smaller improvements

fabiand:
This branch will add a few improvements to livemdia-creator, nothing ground breaking.


bcl:
These all look good to me, we've been using the cache=unsafe in kickstart_tests for a bit, and the discard=unmap sounds like a good idea as well.

bcl:
pushed. Thanks!

PR #73, opened 2015-12-02
Update docs for product.img

bcl:
Make it clear that the contents of product.img and updates.img comes
from the install root, not from the build host's filesystem.


vpodzime:
Looks good to me too.

bcl:
Pushed to master.

bcl:
Pushed.

PR #77, opened 2015-12-16
Write a list of debuginfo packages to /root/debug-pkgs.log (#1068675)

bcl:
Sometimes debugging a boot.iso requires using gdb, and finding the
corresponding debuginfo packages can be difficult. This writes the
matching -debuginfo package names and full ENVR to a file on the iso.
This can then be fed to dnf to install the correct debug packages.

### DIFF
--- a/src/pylorax/imgutils.py
+++ b/src/pylorax/imgutils.py
@@ -204,8 +204,8 @@ def umount(mnt,  lazy=False, maxretry=3, retrysleep=1.0):
             count += 1
             if count == maxretry:
                 raise
-            logger.warn("failed to unmount %s. retrying (%d/%d)...",
-                         mnt, count, maxretry)
+            logger.warning("failed to unmount %s. retrying (%d/%d)...",
+                           mnt, count, maxretry)
             if logger.getEffectiveLevel() <= logging.DEBUG:
                 fuser = execWithCapture("fuser", ["-vm", mnt])
                 logger.debug("fuser -vm:\n%s\n", fuser)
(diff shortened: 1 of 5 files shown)


jkonecny12:
Looks good to me

cgwalters:
It seems like it'd be a lot saner to just stop stripping out the rpmdb.  There a a variety of tools that already know how to find debuginfo packages given that.

This solution also requires the debuginfo repository to be enabled when the ISO is being built, which shouldn't be otherwise necessary.

bcl:
That wouldn't really help, you're not trying to install them in the installer environment. The use case is with gdb remote from a different system. Keeping the rpmdb also increases the install.img size by 8M for no useful reason.

cgwalters:
I understand the gdb remote case, but one can easily query for debuginfo packages post-facto given the ENVR from the rpmdb.

Having the full rpmdb would have helped me in the past for other things as well, mostly knowing the version of `python-kickstart` etc.  So it's not "no useful reason".

One other minor negative here is it's another copy+paste of some historical lore around the naming of debuginfo repositories.

As for the size...I'm somewhat sympathetic, but the biggest win in this area would be a headless boot.iso for people who want to PXE + kickstart.  For live or other media, it's already huge.

vpodzime:
Looks good to me too.

bcl:
> Having the full rpmdb would have helped me in the past for other things as well, mostly knowing the version of python-kickstart etc. So it's not "no useful reason".

There's actually already a list of the installed packages in /root/lorax-packages.log

PR #78, opened 2015-12-19
Add --image-type and --qemu-args

bcl:
This keeps --qcow2, and adds --image-type to allow other disk image types to be created. The old --qcow2 is aliased to --image-type=qcow2 and --qemu-arg is the same as --qcow2-arg


vpodzime:
Looks good to me.

jkonecny12:
Seems good to me too.

bcl:
Pushed

PR #80, opened 2016-01-07
Add --iso-only option to --make-iso

bcl:
This option removes all the extra build artifacts from --make-iso,
leaving only the boot.iso

It also supports naming of the final iso with --image-name


vpodzime:
Looks good to me either way.

PR #81, opened 2016-01-11
Add --iso-name to use with --iso-only

bcl:
Instead of reusing --image-name add a new argument to name the iso. This
way the disk image can be given a unique name with --image-name and the
iso can be named something different.


vpodzime:
Looks good to me.

PR #86, opened 2016-02-05
Move templates to a subpackage and their own directories

bcl:


vpodzime:
Looks good to me otherwise.

bcl:
> Do we (and should we) really require some templates in all cases?

Yes, it won't run at all without templates. By requiring 'lorax-templates' and having any package that provides them also provide 'lorax-templates' we guarantee that at least the generic ones will be installed.

> What if the user has their own not-packaged templates?

Then they point to them with --sharedir, or stick them into /usr/share/lorax/templates.d/01-custom to override any others already installed.

> And what about lmc?

LMC uses the templates under the ./live/ directory. They are very similar to the boot.iso templates, but have some differences specific to live isos.

> Maybe this is a good case for the new weak dependencies functionality in RPM and DNF.

I don't think so. The templates, as long as they are uniquely named, can coexist and 99-generic will act as a fallback.

vpodzime:
Sounds reasonable to me. Thanks for the extra explanation.

bcl:
Pushed.

Frodox:
@bcl is `--conf=mylorax.conf` with
```
[lorax]
sharedir=/path/to/user/templates
```
still works fine instead of `--sharedir` option?

bcl:
Yes that still works. --sharedir is just shorthand so that another file doesn't need to be maintained if the only thing you want to override is the sharedir.

PR #87, opened 2016-02-11
livemedia-creator: Stop passing --repo to anaconda (#1304802)

bcl:
When running a no-virt installation it was parsing the kickstart url
method and passing it to anaconda using --repo which prevents it from
working with url --mirrorlist method. There is no good reason to do
this, anaconda gets the method directly from the kickstart when it isn't
on the cmdline.


jkonecny12:
Looks good to me.

bcl:
Pushed

PR #89, opened 2016-02-19
Include grub2-efi-modules on the boot.iso (#1277227)

bcl:
Some users need to be able to run grub2-install from %pre

Resolves: rhbz#1277227


vpodzime:
Looks good to me.

bcl:
Pushed

PR #90, opened 2016-02-19
Keep modules needed for ast video driver support (#1272658)

bcl:
The ast module depends on:
drm,drm_kms_helper,ttm,syscopyarea,i2c-core,sysfillrect,sysimgblt,i2c-algo-bit

This retains the syscopyarea, sysfillrect, and sysimgblt modules.

Resolves: rhbz#1272658


vpodzime:
Looks good to me.

bcl:
Pushed

PR #91, opened 2016-02-24
templates: Reinstate gpgme-pthread.so for ostree

cgwalters:
See https://github.com/GNOME/ostree/pull/190

Without this the installer ISO generation fails.


bcl:
@cgwalters is this going to change for RHEL7 too or just Fedora?

cgwalters:
It's highly likely the changes would get backported to RHEL7 as well, yeah.   We really want threadsafe GPGME.

bcl:
Pushed to master and rhel7-branch

PR #93, opened 2016-02-26
Add glibc-all-langpacks

dashea:
glibc recently split all of its locale data into subpackages, so if we
install no langpacks we get no locales. Explicitly install all of the
langpacks.

The boot.iso this spits out is about 20MB bigger than the last one from Jenkins, but I'm not sure if that's because of the langpacks or something else.


vpodzime:
Looks good to me.

dashea:
Added a bug number to the commit message and pushed.

PR #94, opened 2016-03-05
lmc: Fixing full filesystem problems

bcl:
Still need to update and test the other example kickstarts.


jkonecny12:
Looks good to me.

bcl:
Added a couple more small fixes and pushed 'em all to master.

PR #95, opened 2016-03-09
Use dnf.rpm.basearch instead of dnf.arch.basearch

lmacken:
:warning: Untested :construction_worker: 

This API was added in 1.1.7
https://dnf.readthedocs.org/en/latest/release_notes.html#release-notes


lmacken:
https://bugzilla.redhat.com/show_bug.cgi?id=1312087

bcl:
Similar patch pushed for 25.1 build.

PR #96, opened 2016-03-09
Keep /usr/bin/xrdb (#1241724)

dashea:
Some font parameters are set via Xresources, so keep the tool that does
that.


bcl:
Pushed to master for 25.1 build, feel free to push to whatever other branches need this.

dashea:
According to adamw on the bug, it's not necessary for 24, so closing.

PR #98, opened 2016-03-15
Create UDF iso when stage2 is >= 4GiB (#1312158)

bcl:
Some images are becoming REALLY large. When a file is >= 4GiB we need to
pass -allow-limited-size to mkisofs to tell it to make a UDF image. Note
that the manpage says that this may result in it not booting on all
systems.

### DIFF
--- a/share/templates.d/99-generic/aarch64.tmpl
+++ b/share/templates.d/99-generic/aarch64.tmpl
@@ -7,6 +7,16 @@ STAGE2IMG="images/install.img"
 LORAXDIR="usr/share/lorax/"
 %>
 
+## Test ${runtime_img} to see if udf is needed
+<%
+    import os
+    from pylorax.sysutils import joinpaths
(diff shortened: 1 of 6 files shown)


jkonecny12:
Looks good to me.

vpodzime:
I have the same concern that @dwlehman had yesterday on IRC -- what about blivet and UDF? (didn't manage to wait for the answer, sorry).

bcl:
Blivet doesn't care that it is UDF. It actually doesn't even mention it in the logs, and it properly detects it as a cdrom and as the backing device, setting it as protected. I can start an install with it and it gets past partitioning and running rsync, then crashes for other reasons.

Also note that this is only for live images with squashfs.img >= 4GiB so it isn't going to show up everywhere, only in spins that are really large.

vpodzime:
Works for me then. Thanks for the extra info.

bcl:
Pushed.

PR #99, opened 2016-03-16
Remove crud from boot.iso

dashea:
Remove a bunch of stuff that webkit and *-firmware pulled in. This adds a --except option to the installpkg command to filter packages from globs before they are handed to dnf.

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -20,7 +20,13 @@ installpkg pigz
 installpkg kernel kernel-modules kernel-modules-extra
 installpkg grubby
 %if basearch != "s390x":
-    installpkg *-firmware
+    ## skip the firmware for sound, video, and scanners, none of which will
+    ## do much good for the installer. Also skip uhd-firmware which is not
+    ## even a kernel firmware package.
+    installpkg *-firmware --except alsa* --except midisport-firmware \
+                          --except crystalhd-firmware --except ivtv-firmware \
+                          --except cx18-firmware --except iscan-firmware \
+                          --except uhd-firmware
     installpkg b43-openfwwf
 %endif
 
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -146,7 +146,6 @@ installpkg vlgothic-fonts
 installpkg wqy-microhei-fonts
 installpkg sil-abyssinica-fonts
 installpkg xorg-x11-fonts-misc
-installpkg gnome-icon-theme-legacy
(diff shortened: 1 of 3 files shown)


dashea:
Added one more commit after realizing that the gnome icon theme doesn't really do anything anymore. That shaves another 10MB or so off of boot.iso.

share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -20,7 +20,13 @@ installpkg pigz
 installpkg kernel kernel-modules kernel-modules-extra
 installpkg grubby
 %if basearch != "s390x":
-    installpkg *-firmware
+    ## skip the firmware for sound, video, and scanners, none of which will
+    ## do much good for the installer. Also skip uhd-firmware which is not
+    ## even a kernel firmware package.
+    installpkg *-firmware --except alsa* --except midisport-firmware \
+                          --except crystalhd-firmware --except ivtv-firmware \
```
bcl: the pkgglob needs to go at the end. Or change what you wrote in ltmpl :)
Good idea though!

dashea:
Redid the comment in ltmpl, since I liked the order I used in runtime-instal better.

src/pylorax/ltmpl.py
```diff
@@ -503,9 +503,35 @@ def installpkg(self, *pkgs):
...
 
+        excludes = []
+        while '--except' in pkgs:
+            idx = pkgs.index('--except')
+            if len(pkgs) == idx+1:
+                raise ValueError("installpkg needs an argument after --except")
+
+            excludes.append(pkgs[idx+1])
+            pkgs = pkgs[:idx] + pkgs[idx+2:]
+
```
bcl: This block expects the --except entries to all be at the start.

share/templates.d/99-generic/runtime-install.tmpl
nullr0ute: You could exclude iscan-firmware too (if you don't already)

src/pylorax/ltmpl.py
dashea: It looks for the --except entries anywhere in the list and pulls them out into the excludes list. So if you wanted you could interleave pkgglobs and --excepts.

share/templates.d/99-generic/runtime-install.tmpl
dashea: Yes, iscan-firmware is the next line down ;-) It was another that was already in runtime-cleanup that I moved to the install command.

src/pylorax/ltmpl.py
bcl: Oh geez, I was thinking of it like a for loop. ACK.

dashea:
Pushed.

PR #100, opened 2016-03-19
Make it possible to generate manpages using Sphinx

bcl:


jkonecny12:
Seems good to me.

bcl:
Pushed

PR #103, opened 2016-03-22
Fix example kickstarts

bcl:
Update for new compose URL, etc.


src/bin/image-minimizer
```diff
@@ -115,37 +115,31 @@ class ImageMinimizer:
...
             return
 
         for pkg in self.drops_rpm:
-            if self.dryrun:
-                print("erasing ", pkg)
-            else:
-                self.ts.addErase(pkg)
+            if self.verbose:
+                print("erasing: %s " % pkg)
+            self.ts.addErase(pkg)
```
jkonecny12: This will change behavior. The ``self.ts.addErase(pkg)`` is now called even in ``dryrun`` mode. Do you really want this modification?

bcl: Whoops, good catch.

bcl:
Fixed a os.rmdir that was using self.verbose instead of self.dryrun

jkonecny12:
Looks good to me then.

bcl:
Pushed.

PR #104, opened 2016-03-23
templates: Fix runtime_img check

bcl:
commit 66241f7cd72a25317dece91d15bf6e30538b9a08 added a check on
runtime_img to create UDF iso's. Ends up it is only in outroot for live,
so switch all the checks to look at it in inroot instead.


dashea:
Looks good to me. This is also reported in https://github.com/rhinstaller/lorax/issues/102 looks like

bcl:
Pushed.

PR #105, opened 2016-03-23
livemedia-creator: Catch dnf download error

bcl:
If there isn't enough space for DNF to download packages it will log:

"Not enough disk space to download the packages."

So add this to the messages in monitor that trigger an error.


jkonecny12:
Looks good to me.

bcl:
Pushed.

PR #106, opened 2016-03-24
Add compression args to installimg template command

bcl:


dashea:
Two things:

Seems like compression args without a leading - should raise a ValueError instead of being silently dropped.

Where does 3700MiB come from? I thought 32-bit processes were limited to 3G of address space to leave room for kernel addresses.

bcl:
The example comes from dgilmore, that's what they use when running pungi on them.

vpodzime:
Looks good to me.

bcl:
Pushed to master and F24, need to backport to F23 and F22.

bcl:
Pushed everywhere.

PR #107, opened 2016-03-24
livemedia-creator: Check selinux state and exit

bcl:
lmc --no-virt was switching selinux to permissive if it was enforcing
and restore it when done. This works fine when it is the only session
running, but would cause problems if it was run in parallel.

It now only checks the state and exits with an error if it isn't already
disabled or in Permissive mode.

Users will need to run setenforce 0 before running lmc.

### DIFF
--- a/src/sbin/livemedia-creator
+++ b/src/sbin/livemedia-creator
@@ -32,6 +32,7 @@ import glob
 import json
 from math import ceil
 import socket
+import selinux
 
 # Use pykickstart to calculate disk image size
 from pykickstart.parser import KickstartParser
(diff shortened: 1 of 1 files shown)


src/sbin/livemedia-creator
```diff
@@ -663,20 +664,12 @@ def novirt_install(opts, disk_img, disk_size):
     :param str disk_img: The full path to the disk image to be created
     :param int disk_size: The size of the disk_img in MiB
 
-    This method makes sure SELinux is permissive during the install, runs anaconda
-    to create the image and then based on the opts passed create a qemu disk image
-    or tarfile.
+    This method runs anaconda to create the image and then based on the opts
+    passed create a qemu disk image or tarfile.
```
dashea: "creates".  Pre-dates this change but might as well fix it as long as you're in there.

src/sbin/livemedia-creator
```diff
@@ -663,20 +664,12 @@ def novirt_install(opts, disk_img, disk_size):
...
     dirinstall_path = ROOT_PATH
 
-    # Set selinux to Permissive if it is Enforcing
-    selinux_enforcing = False
-    if selinux.is_selinux_enabled() and selinux.security_getenforce():
-        selinux_enforcing = True
-        selinux.security_setenforce(0)
-
-    # Clean up /tmp/ from previous runs to prevent stale info from being used
+        # Clean up /tmp/ from previous runs to prevent stale info from being used
```
dashea: I don't think you meant to indent this

dashea:
Looks good other than the two nit-picks

bcl:
Pushed with changes

PR #108, opened 2016-03-24
livemedia-creator: Pass -Xbcj to mksquashfs

bcl:
Some cases of mksquashfs were not using -Xbcj when it is available for
the arch. This adds a function to return the correct args based on the
arch and the cmdline args.

### DIFF
--- a/src/sbin/livemedia-creator
+++ b/src/sbin/livemedia-creator
@@ -315,6 +315,21 @@ def get_arch(mount_dir):
         return "i386"
     return kernels[0].arch
 
+def squashfs_args(opts):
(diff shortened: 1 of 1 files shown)


dashea:
squashfs_args() ignores --compress-arg values instead of only adding -Xbcj if opts.compress_args is empty. Is that intentional?

bcl:
Yes, --compress-arg is really only useful for passing to things like mkcpio since mksquashfs doesn't really support any of the same arguments as xz or gzip, etc.

dashea:
Ok, sounds good then

bcl:
Pushed to master and f24-branch

PR #110, opened 2016-03-29
livemedia-creator: Use correct suffix on default image names (#1318958)

bcl:
When an image name hasn't been passed, and the compression type is
something other than xz, the default image name should use the user
specified compression suffix.

Resolves: rhbz#1318958


jkonecny12:
Looks good to me.

PR #111, opened 2016-03-30
Copying same file shouldn't crash (#1269213)

bcl:
When using the template install command copying the same file to itself
shouldn't crash. Just log the error and continue.

Resolves: rhbz#1269213


jkonecny12:
Looks good to me.

bcl:
Pushed to all branches.

PR #112, opened 2016-04-05
livemedia-creator: Simplify cleanup for no-virt

bcl:
If an anaconda no-virt run crashes it can leave things mounted under
/mnt/sysimage. Previously anaconda-cleanup was used to handle this, but
it will also try to cleanup host mountpoints which isn't desired.


jkonecny12:
Looks good to me.

bcl:
Pushed to all branches.

PR #113, opened 2016-04-05
livemedia-creator: Make sure make-iso kickstart includes dracut-live

bcl:
iso creation requires the dracut-live package, otherwise rebuilding the
inird will crash. Since it takes a long time to discover let's fail
early.


jkonecny12:
Looks good to me.

vpodzime:
There's a typo in the commit message - *inird*. Otherwise this looks good to me too.

bcl:
Pushed to all branches.

PR #114, opened 2016-04-13
Remove Metacity override and theme (#1324890)

bcl:
This will now be handled by Anaconda because it is also needed on the
installed system for Initial Setup.

Resolves: rhbz#1324890


jkonecny12:
Looks good to me.

PR #117, opened 2016-04-27
F24 lmc ostree

bcl:
These patches allow livemedia-creator --no-virt (and --no-virt inside a mock) to be used for --make-pxe-live and --make-ostree-live


jkonecny12:
AFAICT this looks good to me.

bcl:
Pushed to master only, too much change for post-freeze F24.

PR #118, opened 2016-05-03
Create an empty selinux config file (#1332147)

bcl:
In order for selinux to properly label the system it needs to see that
the config file exists.

Also remove the old code trying to copy in a selinux config file, it
never worked -- the removepkg would remove it.

(cherry picked from commit d6584e1d7735cd181ae3cced839bf1118ec0e0ad)

Resolves: rhbz#1332147


vpodzime:
Looks good to me.

PR #119, opened 2016-05-05
Add support for creating the initramfs using the kernel version

bcl:
Previously the initramfs was needed in order to recreate it. This isn't really necessary and adds about 40M to the squashfs.img size. This adds support for recreating the initramfs using the kernel name to recreate the initramfs's name.

Doing this for --make-pxe-live also had problems running out of space when using the minimal rootfs.img so it bind mounts the results directory and /var/tmp before running dracut.

### DIFF
--- a/src/pylorax/treebuilder.py
+++ b/src/pylorax/treebuilder.py
@@ -261,28 +261,35 @@ class TreeBuilder(object):
         If backup is empty, the existing initrd files will be overwritten.
         If suffix is specified, the existing initrd is untouched and a new
         image is built with the filename "${prefix}-${kernel.version}.img"
+
+        If the initrd doesn't exist its name will be created based on the
+        name of the kernel.
         '''
         add_args = add_args or []
         dracut = ["dracut", "--nomdadmconf", "--nolvmconf"] + add_args
         if not backup:
             dracut.append("--force")
 
-        kernels = [kernel for kernel in self.kernels if hasattr(kernel, "initrd")]
-        if not kernels:
-            raise Exception("No initrds found, cannot rebuild_initrds")
+        if not self.kernels:
+            raise Exception("No kernels found, cannot rebuild_initrds")
 
         # Hush some dracut warnings. TODO: bind-mount proc in place?
         open(joinpaths(self.vars.inroot,"/proc/modules"),"w")
-        for kernel in kernels:
+        for kernel in self.kernels:
             if prefix:
-                idir = os.path.dirname(kernel.initrd.path)
+                idir = os.path.dirname(kernel.path)
                 outfile = joinpaths(idir, prefix+'-'+kernel.version+'.img')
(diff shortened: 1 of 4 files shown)


src/pylorax/treebuilder.py
```diff
@@ -261,28 +261,35 @@ def rebuild_initrds(self, add_args=None, backup="", prefix=""):
...
         '''
         add_args = add_args or []
         dracut = ["dracut", "--nomdadmconf", "--nolvmconf"] + add_args
         if not backup:
             dracut.append("--force")
 
-        kernels = [kernel for kernel in self.kernels if hasattr(kernel, "initrd")]
-        if not kernels:
+        if not self.kernels:
             raise Exception("No initrds found, cannot rebuild_initrds")
```
vpodzime: Shouldn't this be ``"No kernels found,..."``?

src/sbin/livemedia-creator
```diff
@@ -483,20 +482,27 @@ def rebuild_initrds_for_live(opts, sys_root_dir, results_dir):
...
+    mount(results_dir, opts="bind", mnt=joinpaths(sys_root_dir, "results"))
+    # Dracut runs out of space inside the minimal rootfs image
+    mount("/var/tmp", opts="bind", mnt=joinpaths(sys_root_dir, "var/tmp"))
     for kernel in kernels:
-        outfile = kernel.initrd.path + ".live"
+        if hasattr(kernel, "initrd"):
+            outfile = os.path.basename(kernel.initrd.path)
+        else:
+            # Construct an initrd from the kernel name
+            outfile = os.path.basename(kernel.path.replace("vmlinuz-","initrd-")+".img")
```
vpodzime: This line could use some spaces. :)

vpodzime:
Other than the above nitpicks this looks good to me.

jkonecny12:
There is a typo in 3rd commit message:
> can rebuild it using the **the** version from the kernel.

dashea:
Won't this require an anaconda change as well? Part of the reason it's in the stage2 of liveimg is the bootloader is installed before the initramfs is regenerated, and if the initramfs is absent grub2-mkconfig outputs a busted config.

bcl:
> Won't this require an anaconda change as well? Part of the reason it's in the stage2 of liveimg is the bootloader is installed before the initramfs is regenerated, and if the initramfs is absent grub2-mkconfig outputs a busted config.

Good question, it looks like this is fixed (at least on rawhide where I've been testing). new-kernel-pkg is run after grub2-mkconfig and the final grub.cfg is correct. I'll double check this with f24 too.

bcl:
Ends up @dashea was right, on BTRFS it ends up with the wrong kernel cmdline. https://github.com/rhinstaller/anaconda/pull/617 is the associated fix for anaconda.

PR #122, opened 2016-05-11
Add efi, product, and updates image paths to treeinfo

bcl:
Include the information about efiboot.img, macboot.img, product.img, and
updates.img in .treeinfo.


jkonecny12:
Looks good to me.

bcl:
Pushed to master only.

PR #124, opened 2016-05-17
Update lmc UEFI support to use the edk2-ovmf package

bcl:
Fedora now has a edk2 package so use the OVMF code from there. This also
adds using a copy of OVMF_VARS for each boot instead of reusing the one
provided by the package.


vpodzime:
Looks good to me.

PR #126, opened 2016-06-01
livemedia-creator: Always copy novirt logs before cleanup

bcl:
Before attempting to cleanup any dangling anaconda mounts copy the
anaconda logs to their final location.

Also, catch failures to cleanup the mounts, log it, and continue trying
the other mountpoints. A cleanup failure will result in an InstallError
instead of a CalledProcessError.

### DIFF
--- a/src/sbin/livemedia-creator
+++ b/src/sbin/livemedia-creator
@@ -690,16 +690,26 @@ def anaconda_cleanup(dirinstall_path):
     Cleanup any leftover mounts from anaconda
 
     :param str dirinstall_path: Path where anaconda mounts things
+    :returns: True if cleanups were successful. False if any of them failed.
 
     If anaconda crashes it may leave things mounted under this path. It will
     typically be set to /mnt/sysimage/
+
+    Attempts to cleanup may also fail. Catch these and continue trying the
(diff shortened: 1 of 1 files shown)


jkonecny12:
Looks good to me.

src/sbin/livemedia-creator
```diff
@@ -690,16 +690,26 @@ def anaconda_cleanup(dirinstall_path):
     Cleanup any leftover mounts from anaconda
 
     :param str dirinstall_path: Path where anaconda mounts things
+    :returns: True if cleanups were successful. False if any of them failed.
 
     If anaconda crashes it may leave things mounted under this path. It will
     typically be set to /mnt/sysimage/
+
+    Attempts to cleanup may also fail. Catch these and continue trying the
+    other mountpoints.
```
vpodzime: This paragraph sounds more like a comment to me than a part of the documentation.

bcl: No, that's telling you what it will do. It used to just crash with a CalledProcessError and exit. Now it will try all the mountpoints, even if some of them fail the cleanup attempt.

PR #127, opened 2016-06-02
Stop using undocumented DNF logging API

bcl:
Lorax already takes over the "dnf" logger and directs it to ./dnf.log so
all this was doing was generating deprecation warnings.


vpodzime:
Looks good to me.

bcl:
Pushed to master.

PR #128, opened 2016-06-02
Add Requires on kmod for modinfo tool (#1342041)

bcl:


bcl:
Ends up this wasn't the problem and kmod gets pulled in by other required things.

PR #129, opened 2016-06-15
Keep the pci utilities for use in kickstarts (#1344926)

bcl:
Resolves: rhbz#1344926


simpsonar77:
thanks for the quick look at this!  Will be much better than me copying the lspci binary in during the kickstart %pre section

PR #131, opened 2016-06-16
Keep fsfreeze in install environment (#1315468)

kellinm:
The latest POWER platform allows a host machine to configure guests
running in a different endian mode. Guests configured in this way may
have their bootloader configuration file corrupted after installation if
the file was not fully written to disk. The host machine would read the
journal and try to finish writing the file in the wrong endian mode.

Issuing an fsfreeze and unfreeze gives more assurance that the
configuration file is properly written before a reboot; this patch adds
fsfreeze to the installer runtime environment.

Related: rhbz#1315468


bcl:
Pushed to master.

bcl:
Cherry-picked into lorax-19.6.81-1

PR #132, opened 2016-06-17
Add back libraries needed by spice-vdagent

dashea:
spice-vdagent, half of the thing that lets copy/paste work across a
spice connection, added sound support, I guess, so now we need
alsa-libs.

(cherry picked from commit 89441cd6752a97ce2ef62438a015e452f1421f77)

Resolves: rhbz#1347737


dashea:
Needs acks, of course

PR #134, opened 2016-06-20
Make sure cmdline config file exists (#1348304)

bcl:
If the --config file doesn't exist lorax will fall back to its defaults,
which are probably not what you want. Now it will exit if it doesn't
exist.


vpodzime:
Looks good to me.

PR #135, opened 2016-06-21
Keep all of the kernel drivers/target/ modules (#1348381)

bcl:
The ib_srpt, and ib_isert modules depend on these.

Resolves: rhbz#1348381


bcl:
Pushed to both branches.

PR #136, opened 2016-06-22
Change installpkg default to --required

bcl:
Switch the installpkg default to --required so that missing packages will cause the boot.iso creation to fail.


vpodzime:
Looks good to me.

PR #137, opened 2016-07-08
Use the size from the kickstart, not the minimum size, for the root fs

bcl:


vpodzime:
The commit message of the second commit could refer to the ported commit instead of the original one. Otherwise this looks good to me.

PR #138, opened 2016-07-11
Installing *-firmware should be optional

bcl:


share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -23,8 +23,8 @@ installpkg grubby
     ## skip the firmware for sound, video, and scanners, none of which will
     ## do much good for the installer. Also skip uhd-firmware which is not
     ## even a kernel firmware package.
-    installpkg *-firmware --except alsa* --except midisport-firmware \
-                          --except crystalhd-firmware --except ivtv-firmware \
+    installpkg --optional *-firmware --except alsa* --except midisport-firmware \
+                           --except crystalhd-firmware --except ivtv-firmware \
```
ausil: Extra space here?

PR #139, opened 2016-07-11
Install storaged-iscsi to the runtime (#1347415)

vpodzime:
Blivet now needs it to work with iSCSI (with the iSCSI support being optional in
Blivet).

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -93,6 +93,9 @@ installpkg system-storage-manager
 installpkg device-mapper-persistent-data
 installpkg xfsdump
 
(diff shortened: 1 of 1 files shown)


bcl:
Do we really need to add this explicitly? Or is there something else that should be adding it as a requirement like Anaconda or Blivet?

vpodzime:
> Do we really need to add this explicitly? Or is there something else that should be adding it as a requirement like Anaconda or Blivet?

Hard to say, probably not as a Blivet requirement, because Blivet explicitly tries to not make this a dependency. Anaconda could make more sense, but the logic of it would be quite twisted as it's not using storaged.

vpodzime:
Pushed.

PR #140, opened 2016-07-18
Fix aarch64 installs due to missing/unsupported packages

nullr0ute:
With commit fe17f97 changing the default from optional to required there's
a few packages that aren't currently supported on aarch64 that break the
compose. In particular aarch64 currently still doesn't have kexec, with luck
that might change in the F-25 cycle but until it does we need to have an
exception.

Signed-off-by: Peter Robinson <pbrobinson@fedoraproject.org>

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -3,7 +3,9 @@
 
 ## anaconda package
 installpkg anaconda anaconda-widgets
+%if basearch != "aarch64":
 installpkg kexec-tools-anaconda-addon
+%endif
 ## anaconda deps that aren't in the RPM
 installpkg tmux
 ## for anaconda crash handling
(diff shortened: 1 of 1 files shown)


share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -115,8 +117,9 @@ installpkg bridge-utils
 ## hardware utilities/libraries
 installpkg pciutils usbutils ipmitool
 installpkg mt-st smartmontools
-%if basearch != "s390x":
-installpkg hdparm pcmciautils
+installpkg hdparm
```
bcl: I don't think hdparm is available on s390 so it still needs a wrapper, or --optional

nullr0ute:
s390x does have hdparm, I explicitly checked that before I made that change, it's been there since F-22 :-)

http://s390.koji.fedoraproject.org/koji/packageinfo?packageID=4392

bcl:
Pushed to master.

PR #141, opened 2016-07-22
livemedia-creator: Use hd:LABEL for stage2 iso (#1355882)

bcl:
The correct way with the current dracut (and it should be backwards
compatible) is to reference it as stage2=hd:LABEL=...

(cherry picked from commit 4bc4b4c5a50c595e3dbe27fe530d27b8bd46faba)
Resolves: rhbz#1355882

### DIFF
--- a/src/sbin/livemedia-creator
+++ b/src/sbin/livemedia-creator
@@ -299,7 +299,7 @@ class VirtualInstall( object ):
         if kernel_args:
             extra_args += " "+kernel_args
         if iso.liveos:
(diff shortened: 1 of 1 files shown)


vpodzime:
So ``CDLABEL=`` works too?

bcl:
> So CDLABEL= works too?

Actually it does, but it should just be LABEL. Thanks for pointing that out.

vpodzime:
Looks good to me.

PR #142, opened 2016-07-25
Use size=10 by default

cgwalters:
I'm working on
https://fedoraproject.org/wiki/Changes/WorkstationOstree and when
using lorax to make an installer ISO with content embedded, I run out
of disk space since the desktop+various apps is large.

Since this ends up being compressed anyways, let's just bump the
currently arbitrary `2` to `10` - the only real cost I can think of is
going to be a few more superblock entries.

### DIFF
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -166,7 +166,7 @@ class Lorax(BaseLoraxClass):
             isfinal=False, workdir=None, outputdir=None, buildarch=None, volid=None,
             domacboot=True, doupgrade=True, remove_temp=False,
             installpkgs=None,
-            size=2,
+            size=10,
             add_templates=None,
             add_template_vars=None,
             add_arch_templates=None,


bcl:
Increasing the size does result in somewhat larger images, even with compression, as well as longer compose time. I'd rather we add something like a --root-fs-size argument that gets passed down to it, this is how it was handled for the pxe live code in lmc.

cgwalters:
Are you sure?  That'd be surprising.  I just tried changing the size to 500 to magnify any size difference, and in the end it added 1MB (which is kind of well within the "squashfs noise" level IME):

```
[root@localhost lorax]# ls -alh installer{,.unpatched}/images/boot.iso
-rw-r--r--. 1 root root 439M Aug  5 15:35 installer/images/boot.iso
-rw-r--r--. 1 root root 438M Aug  5 15:04 installer.unpatched/images/boot.iso
```

cgwalters:
Bumping to 20GB also only results in ~1MB increase in my tests.

bcl:
Oh, that's good news then. I thought I'd seen a larger increase than that.

bcl:
Pushed.

cgwalters:
This was changed in https://github.com/rhinstaller/lorax/pull/150

PR #144, opened 2016-07-27
Don't install python3-dnf-langpacks

vpodzime:
The package has been retired [1] and anaconda no longer needs it.

[1]
http://pkgs.fedoraproject.org/cgit/rpms/dnf-langpacks.git/commit/?id=51922797be2e6157582b86580b9898e146291245

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -124,9 +124,6 @@ installpkg pcmciautils
 installpkg libmlx4 rdma
 installpkg rng-tools
 
-## translations & language packs
-installpkg python3-dnf-langpacks
-
 ## fonts & themes
 installpkg bitmap-fangsongti-fonts
 installpkg dejavu-sans-fonts dejavu-sans-mono-fonts


pnemade:
Thank you vpodzime for your quick assistance and I am sorry that I forgotten anaconda has this requirement still kept. The reason I had forgotten is dnf langpacks never worked the way its supposed to be worked like yum-langpacks. 

I requested for anaconda to add similar code like yum langpacks plugin. Thanks to David Shea for that help and then I did all minimal support for langpacks in dnf-langpacks package but the real work of pulling packages on updates or when new conditional package is installed, installing its required langpacks was not working. Again this is where dnf deliberately avoided adding [hooks](https://bugzilla.redhat.com/show_bug.cgi?id=1114422) that were present in yum.

I have been asking for dnf people to provide that hook since last 2 years now. Finally we settled with rpm weakdeps tag but then again from Live images langpacks selection cannot happen as live cannot contain all langpacks packages[[1]](https://bugzilla.redhat.com/show_bug.cgi?id=1319073). Another issues bodhi runs on epel7 and updates repository cannot generate weakdeps.

So after all these things, langpacks is still not 100% working.

Sorry for this long comment. Just wanted to put additional information here why decided to retire dnf-langpacks as it became duplication of https://fedoraproject.org/wiki/Changes/LangpacksInstallationWithRPMWeakDependencies.

PR #145, opened 2016-07-31
Allow supplying a disk image for PXE live systems

friedrich:
Fixes that the combination of command line arguments
"--make-ostree-live" and "--disk-image" lead to a program error.


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #146, opened 2016-08-12
Alt arch fixes

nullr0ute:
A few fixes for aarch64/ppc64le/s390x. In particular the ppc64le patch is needed for Fedora 25 Alpha as it's breaking the Power64 compose process.


nullr0ute:
patch borken, new PR coming in a moment

PR #147, opened 2016-08-12
templates fixes for alternate architectures

nullr0ute:
So a few minor fixes for changes in package sets for Alt Arches. The ppc64le patch is fallout from the change of default optional to requires (should be the last with luck) and is needed for Fedora 25 alpha to get a successful compose on ppc64le


bcl:
Pushed.

PR #150, opened 2016-08-23
lorax: Add --rootfs-size (#1368743)

bcl:
This controls how big the root filesystem is for the squashfs used in the boot.iso, the default is 2GiB.
    
Note that larger rootfs sizes will require more memory and may cause the build to fail.

### DIFF
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -166,7 +166,7 @@ class Lorax(BaseLoraxClass):
             isfinal=False, workdir=None, outputdir=None, buildarch=None, volid=None,
             domacboot=True, doupgrade=True, remove_temp=False,
             installpkgs=None,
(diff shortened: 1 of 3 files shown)


vpodzime:
Looks good to me.

cgwalters:
Okay, though in the big picture I'm not sure it's how sustainable it is to expect image builds to work in just 4G.  Is there not swap enabled too?

AdamWill:
@cgwalters @ausil is trying to figure out and fix the hanging issue if possible, but for now we do need to be able to work around it in order to compose Alpha; go/no-go is Thursday.

ausil:
@cgwalters 32 bit images can only work in 4G, they can not address more than that per process. even with PAE/LPAE you still only get 4G per process.

PR #152, opened 2016-08-23
Rhel7 branch rsyslog via journald

rvykydal:


bcl:
As long as this doesn't interfere with virtio logging or the --syslog option it looks fine to me.

share/config_files/common/rsyslog.conf
```diff
@@ -15,35 +32,59 @@ $InputFileStateFile anaconda-tb-statefile
...
+
+# Where to place auxiliary files
+#$WorkDirectory /var/lib/rsyslog
+
 # Use default timestamp format
 $ActionFileDefaultTemplate RSYSLOG_TraditionalFileFormat
 
+# File syncing capability is disabled by default. This feature is usually not required,
+# not useful and an extreme performance hit
+#$ActionFileEnableSync on
```
vpodzime: Do we really want this?

vpodzime:
Looks good to me too otherwise.

share/config_files/common/rsyslog.conf
rvykydal: There is no change regarding this option.
I'd leave it for follow-up consideration.

rvykydal:
Both remote logging and virtio channel logging features keep working with the patch.

PR #158, opened 2016-09-15
Make sure dir for rngd exists [BZ#1377430]

Frodox:
It is good to ensure that dir exists before doing anything with it.

Currently it works with C7.2 just because there are `firewalld` inside anaconda chroot, which creates this dir :D
Also, firewalld is pulled not by `.tmpl`, but by indirect dependencies, so, it is not very reliable.

https://bugzilla.redhat.com/show_bug.cgi?id=1377430


Frodox:
Looks like I'm not alone. Same as #156 :D (but for `rhel7-branch`)

Frodox:
bump @bcl ..?

bcl:
This can't be changed until it is approved for RHEL-7.4

bcl:
Fixed in lorax-19.6.79-1

PR #167, opened 2016-10-17
Add missing fonts (#1370118)

poncovka:
Added fonts to support Tibetan, Khmer and Burmese.


vpodzime:
Cannot help myself from thinking that somebody should create the '@AllLangFonts' package group that would contain all those fonts we need.

bcl:
Pushed to master and f25-branch.

PR #169, opened 2016-11-04
Added option to remove packages (parallel to installpkgs)

jcpunk:
The code to do removals already existed, I've just added a command line switch to manipulate it.  The end behavior is basically the same as installpkgs, only it removes them instead.

### DIFF
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -165,7 +165,7 @@ class Lorax(BaseLoraxClass):
     def run(self, dbo, product, version, release, variant="", bugurl="",
             isfinal=False, workdir=None, outputdir=None, buildarch=None, volid=None,
             domacboot=True, doupgrade=True, remove_temp=False,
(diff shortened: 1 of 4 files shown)


cgwalters:
Why?

jcpunk:
I'd use this for much the same reason as installpkgs.

The EL7 default template is almost good enough for my needs.  I just don't want this one package installed in the media.  This switch allows me to directly use the official template while automating my one small change.

cgwalters:
Can you say what that one package is?

jcpunk:
My specific usage is with subscription-manager.  For SL, we don't use it so having it on the media just results in creating some redhat.repo files in /etc/ that aren't owned by anything.

cgwalters:
Thanks!  That's the kind of thing I think belongs in commit messages, but it's ok to have here too :smile: 

The patch seems sane to me.  I'm not a lorax committer though.

PR #171, opened 2016-11-08
templates: When a subprocess fatallly errors, output its stderr directly

cgwalters:
Right now we dump all subprocess output to `program.log`.  Unfortunately,
The pungi/koji stack doesn't know how to scrape out the lorax logs.

And even when running interactively, it's annoying that *some* fatal
errors show up on stderr, but if it's from a subprocess, I need to go
over and `tail program.log`.

Let's output the subprocess stderr directly, since the user is
going to want it prominently anyways.

### DIFF
--- a/src/pylorax/ltmpl.py
+++ b/src/pylorax/ltmpl.py
@@ -518,8 +518,8 @@ class LoraxTemplateRunner(object):
             logger.debug("command finished successfully")
         except CalledProcessError as e:
             if e.output:
-                logger.debug('command output:\n%s', e.output)
-            logger.debug('command returned failure (%d)', e.returncode)
+                logger.error('command output:\n%s', e.output)
+            logger.error('command returned failure (%d)', e.returncode)
             raise
 
     def installpkg(self, *pkgs):


Frodox:
> Let's output the subprocess stderr directly, since the user is
going to want it prominently anyways.

Not completely agree. I'm often run lorax from another scripts and save it's logs. And in case of **lorax**'s fail I know which logs to view. It's convenient to have all logs for program in one place.
Can we leave error in program.log and dublicate to stderr for user convenience ?
Like adding `tail program.log` at the end :-)

cgwalters:
The error is *also* in `program.log`.

Frodox:
> The error is also in program.log.

Ok, it is fine :)

src/pylorax/ltmpl.py
```diff
@@ -518,7 +518,7 @@ def runcmd(self, *cmdlist):
             logger.debug("command finished successfully")
         except CalledProcessError as e:
             if e.output:
-                logger.debug('command output:\n%s', e.output)
+                logger.error('command output:\n%s', e.output)
             logger.debug('command returned failure (%d)', e.returncode)
```
bcl: May as well raise this up to .error as well.

cgwalters:
Updated :arrow_up:

PR #172, opened 2016-11-14
F26 alt arch fixes

nullr0ute:
Two minor fix ups. The second firmware exclude is really generic but a minor fix.


share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -27,8 +24,8 @@ installpkg grubby
     ## even a kernel firmware package.
     installpkg --optional *-firmware --except alsa* --except midisport-firmware \
                            --except crystalhd-firmware --except ivtv-firmware \
-                          --except cx18-firmware --except iscan-firmware \
-                          --except uhd-firmware
+                           --except cx18-firmware --except iscan-firmware \
```
bcl: Looks like some extra spaces crept in there :)

bcl: NM. the extra space is actually correct :)

PR #173, opened 2016-11-15
fix PYTHONDIR determenation in case of having both python2 and python3 pkgs in env

Frodox:
In case of there is several python pkgs inside, like (python2 and python3) - current variant will fail / determine wrong. It's *more reliable* to determine dir something like this..

### DIFF
--- a/share/templates.d/99-generic/runtime-postinstall.tmpl
+++ b/share/templates.d/99-generic/runtime-postinstall.tmpl
@@ -3,7 +3,10 @@
 
 <%page args="root, basearch, libdir, configdir"/>
 <%
-PYTHONDIR = sorted(glob("usr/"+libdir+"/python?.?"))[0]
+## if we have both python-2 and python-3 pkgs in environment
+import os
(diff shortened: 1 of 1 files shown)


cgwalters:
It seems like we should be able to retrieve this from the environment somehow...but I don't know if something like `sys.version` needs to be explicitly passed through.

Anyways, seems reasonable to me if you've tested it.

Frodox:
I have tested it (and using actually day to day) in my rhel7.3-env. I'll try to test it from master/F25 env...

bcl:
Your patch is a good idea, pinning it to the anaconda version. But after looking harder at this I realize that this code is no longer needed. It used to be used for moving sitecustomize.py, which we no longer do.

Frodox:
@bcl hah, you are right. I detected it in rhel7-branch, so, there was such thing. And now (master), looks like no one use `PYTHONDIR` at all. So, should/can I just remove this line?

bcl:
yes.

PR #174, opened 2016-11-16
imgutils: Don't relabel /ostree

cgwalters:
OSTree is a deduplicating hardlink store using a new file path
`/ostree`, which SELinux policy doesn't know about.  However, OSTree
has SELinux support built in, and rpm-ostree (for example) uses this
to ensure the attributes on files stored there are simply always
correct.  Relabeling it will corrupt it.

Hence, let's skip it.


cgwalters:
This is a requirement for https://pagure.io/fedora-lorax-templates/pull-request/6

PR #178, opened 2017-01-13
Do not install rsh to anaconda chroot

Frodox:
* `rsh` not used in anaconda now (can't find any mentions)
* `rsh` duplicates already existing ssh
* `rsh` not secure enough these days


bcl:
Pushed, thanks.

PR #179, opened 2017-01-13
Only cleanup libhistory from readline

ausil:
with readline 7 libreadline.so moved to libdir as a result it is
currently removed in the cleanup phase, however it is needed

Signed-off-by: Dennis Gilmore <dennis@ausil.us>


AdamWill:
big ol' ACK. 👍

bcl:
Pushed, thanks.

AdamWill:
@bcl thanks a lot - just to be sure you know, I believe we need a package build with this ASAP also. Rawhide composes are broken without the change.

PR #181, opened 2017-02-03
Add dependency for lvmdump -l command (#1255659)

jkonecny12:
This is required for pre installation log feature in Anaconda.

Required for https://github.com/rhinstaller/anaconda/pull/947 .

*Related: rhbz#1255659*

### DIFF
--- a/share/runtime-cleanup.tmpl
+++ b/share/runtime-cleanup.tmpl
@@ -59,7 +59,7 @@ removepkg avahi-autoipd coreutils-libs dash db4-utils diffutils file
 removepkg genisoimage gnome-python2 info iptables
 removepkg jasper-libs libXxf86misc
 removepkg libasyncns libhbaapi libhbalinux
-removepkg libmcpp libpcap libtiff libutempter linux-atm-libs
+removepkg libmcpp libtiff libutempter linux-atm-libs
 removepkg lvm2-libs m4 mailx makebootfat mcpp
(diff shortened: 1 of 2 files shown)


vpodzime:
Maybe rather "Do not remove dependency..."?

vpodzime:
> Maybe rather "Do not remove dependency..."?

Oh, you are also adding a package. Shouldn't this be handled by the ``lvmdump``'s dependencies?

jkonecny12:
As I understand it. It is not required by `lvmdump` it is more like optional dependency.

When you don't have `nc` package installed and you will run `lvmdump -l` you will get this message:
```WARNING: Neither socat or nc -U seems to be available...```
And this looks like expected behavior to me.

The `libpcap` library is required by `nmap-ncat`. I'll add this to commit message.

jkonecny12:
@bcl Could you please merge this PR and give me echo after a new build will be created. Thanks.

bcl:
Dangit. I just did a new build without this. How about I do another tomorrow to see if any other rawhide things show up?

jkonecny12:
Yeah sure, no problem, no need to rush, we have some time for this :). The same change is here https://github.com/rhinstaller/lorax/pull/182 for master.

bcl:
Pushed and built into lorax-19.6.80-1

jkonecny12:
Nice, thank you!

PR #182, opened 2017-02-06
Add dependency for lvmdump -l command (#1255659)

jkonecny12:
This is required for pre installation logging feature in Anaconda.

The `libpcap` library is dependency for `nmap-ncat` which is dependency for `lvmdump -l`.

Backport of PR: https://github.com/rhinstaller/lorax/pull/181


bcl:
Pushed, it will be in the next build.

PR #184, opened 2017-02-14
Fix duplicate kernel messages in /tmp/syslog (#1382611)

rvykydal:
Resolves: rhbz#1382611

The kernel messages will be read from journal so don't use additional imlkog
module to read them duplicitly.

Fixup of
commit b5d0b1fd79cb3b785f2c46be4488af9a51ad5dc9


rvykydal:
Pushed

PR #185, opened 2017-02-14
Fix duplicate kernel messages in /tmp/syslog (#1382611)

rvykydal:
Resolves: rhbz#1382611

The kernel messages will be read from journal so don't use additional imlkog
module to read them duplicitly.

Fixup of
commit 3eca8a042541b874981bc2f3ee6f95d6aba122b7


rvykydal:
Pushed

PR #189, opened 2017-02-21
Mount /dev to ensure /dev/urandom is available when executing rpm/nss

kaie:
NSS 3.29.x requires that /dev/urandom is available, which is the expected source of random data on Linux.
lorax uses rpm, rpm uses nss.
Without /dev/urandom, rpm fails.
See https://bugzilla.redhat.com/show_bug.cgi?id=1420523


ignatenkobrain:
> lorax uses rpm, rpm uses nss.

We need to note that in F27 RPM will use OpenSSL, so we can safely remove it for f27.

bcl:
Pushed a slightly modified version (mknod isn't a template command, have to run it with runcmd)

PR #190, opened 2017-02-27
add ostree to get installed in anaconda environment

dustymabe:
rpm-ostree used to have a requirement on the ostree rpm. It no
longer has that dependency, but rather requires ostree-libs. However,
we call ostree directly in some cases so we need to have it installed.

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -9,7 +9,7 @@ installpkg tmux
 installpkg gdb
 ## Other available payloads
 installpkg dnf
-installpkg rpm-ostree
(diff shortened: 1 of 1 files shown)


dustymabe:
Note this is currently causing issues for rawhide builds of Atomic Host: https://koji.fedoraproject.org/koji/taskinfo?taskID=18091547

cgwalters:
In the long term, we want rpm-ostree to lose the dependency on `/usr/bin/ostree` and just use the shared library.  We're a ways off from that for a few reasons, among them obviously that the anaconda `rpmostreepayload.py` is really just `ostreepayload.py` right now.

cgwalters:
So 👍  from me.

PR #191, opened 2017-03-08
Add --noverifyssl to lorax (#1430483)

bcl:
Previously lorax had no way to use repos with self-signed certificates.
This adds the --noverifyssl cmdline option which will ignore certificate
errors.

Resolves: rhbz#1430483


bcl:
Pushed.

PR #193, opened 2017-03-09
INSTALL_ROOT and LIVE_ROOT are not available during %post

bcl:
These were set by livecd-creator. In order to install the license files to the / of the iso they need to be copied in the arch-specific templates and added to the mkisofs graft path.


bcl:
Pushed.

PR #199, opened 2017-03-28
Add support for --repo to read yum .repo files directly (#1430479)

bcl:
This makes it easier to specify existing repos with extra args, eg.
/etc/yum.repos.d/redhat.repo generated by subscription-manager.

Resolves: rhbz#1430479

### DIFF
--- a/docs/lorax.1
+++ b/docs/lorax.1
@@ -3,7 +3,7 @@
 lorax \- Create installer boot iso
 
 .SH SYNOPSIS
-lorax -p PRODUCT -v VERSION -r RELEASE -s REPOSITORY OUTPUTDIR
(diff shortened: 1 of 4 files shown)


Frodox:
What is blocking this one?

Frodox:
I think it would be cook to documented it in man-page also

bcl:
Pushed

Frodox:
@bcl Can I use priority and yum-plugin-priority here? Does it work?

bcl:
> @bcl Can I use priority and yum-plugin-priority here? Does it work?

The best way to answer that is to try it :) I haven't tried using any plugins. But if it's available on the host and the config file is setup I'd expect it to work.

Frodox:
Well, I'm waiting for this new lorax in centos 7.[34] repos :)

PR #223, opened 2017-06-26
Make sure loop device is setup (#1462150)

bcl:
It seems that on rare occasions losetup can return before the /dev/loopX
is ready for use, causing problems with mkfs. This tries to make sure
that the loop device really is associated with the backing file before
continuing.

NOTE that using losetup --list -O to return the backing store
associated with the loop device can fail due to losetup truncating
the output filename if sysfs isn't setup. Instead of printing the full
path it will truncate it to 64 characters with a * at the end.

See util-linux lib/loopdev.c for the code that does this.

Use the existing get_loop_name function, which uses losetup -j, to lookup
the loop device associated with the backing store which should work the
same, just in the opposite direction.

### DIFF
--- a/src/pylorax/imgutils.py
+++ b/src/pylorax/imgutils.py
@@ -24,6 +24,7 @@ import os, tempfile
 from os.path import join, dirname
 from subprocess import Popen, PIPE, CalledProcessError
 import sys
+import time
 import traceback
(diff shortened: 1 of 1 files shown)


bcl:
Pushed to master.

PR #227, opened 2017-07-06
runtime-cleanup.tmpl: don't delete localedef

jlebon:
This is required in the future for anaconda to be able to inspect the
supported locales in Atomic Host installations. This is the same patch
as https://github.com/rhinstaller/lorax/pull/194 but for f26-branch.


jlebon:
Used by https://github.com/rhinstaller/anaconda/pull/1122.

bcl:
No need for 2 PR's, Just use labels on #226

PR #230, opened 2017-08-08
Include the dracut fips module in the initrd (#1341280)

bcl:
This will allow anaconda to fetch kickstarts using https when installing
with fips=1

Leave vmlinuz and .vmlinuz.hmac in /boot

dracut-fips module needs the vmlinuz.hmac file in order to boot.

Resolves: rhbz#1341280


lsiudut:
@bcl this patch breaks provisioning on virtual machines. See my comment in #1341280 (I don't know which service is better for this discussion).

PR #235, opened 2017-08-08
Add the version to the log (#1335456)

bcl:
Log the version number as INFO and add -V version display to
livemedia-creator.

Resolves: rhbz#1335456


bcl:
Pushed

PR #250, opened 2017-10-05
Fix pylint errors and warnings

atodorov:


atodorov:
I was looking at the build job in Jenkins which is against master so used that for my base. I imagine we can rebase to whatever. Just let me know where do you want these changes to appear.

(My intention is to add more tests for the lorax-compose branch later but there's no reason why master can't benefit from that)

PR #253, opened 2017-10-06
Add swaplabel command

poncovka:
It is required by libblockdev.


poncovka:
Resolves: rhbz#1500992

@bcl Could you release a new version of lorax, please?

PR #252, opened 2017-10-05
[master] Add coverage

atodorov:

PR #255, opened 2017-10-09
[lorax-composer] Add coverage

atodorov:
Initial test & coverage for the lorax-composer branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,13 @@ check:
 	@echo "*** Running pylint ***"
 	PYTHONPATH=$(PYTHONPATH):./src/ ./tests/pylint/runpylint.py
 
+test:
+	@echo "*** Running tests ***"
+	PYTHONPATH=$(PYTHONPATH):./src/ $(PYTHON) -m nose -v --with-coverage --cover-erase --cover-branches \
+                        --cover-package=pylorax --cover-package=bin --cover-package=sbin --cover-inclusive
+	coverage report -m
+	[ -f "/usr/bin/coveralls" ] && [ -n "$(COVERALLS_REPO_TOKEN)" ] && coveralls || echo
+
 clean:
 	-rm -rf build src/pylorax/version.py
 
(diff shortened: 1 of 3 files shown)


atodorov:
@bcl same changes for lorax-composer branch as in PR #252

Makefile
```diff
@@ -25,6 +25,13 @@ check:
 	@echo "*** Running pylint ***"
 	PYTHONPATH=$(PYTHONPATH):./src/ ./tests/pylint/runpylint.py
 
+test:
+	@echo "*** Running tests ***"
+	PYTHONPATH=$(PYTHONPATH):./src/ $(PYTHON) -m nose -v --with-coverage --cover-erase --cover-branches \
+                        --cover-package=pylorax --cover-package=bin --cover-package=sbin --cover-inclusive
+	coverage report -m
+	[ -f "/usr/bin/coveralls" ] && [ -n "$(COVERALLS_REPO_TOKEN)" ] && coveralls || echo
```
bcl: Where does coveralls come from? I don't see it in RHEL7 or in EPEL7.
We need to make sure that all of the tests can be run from a RHEL7 system without adding a bunch of 3rd party binary blobs (I think we should be ok using EPEL stuff, eg. pylint, but should limit it to that)

atodorov: python-coveralls from EPEL

bcl: Ok, I see python-coverage but not python-coveralls. I initially thought they were the same, but see they aren't so this looks ok.

atodorov: I see what's happening now. The python-coveralls package is built for EPEL but the last (only) available build has been removed, see:
https://koji.fedoraproject.org/koji/packageinfo?packageID=22032

Still I think if we want to use this service (I want) we can install one of the packages from Fedora or ask the EPEL maintainer to rebuild the package.

bcl: Yeah, we could also help maintain it if we're going to depend on it.

PR #264, opened 2017-10-24
Add Recipe Git repo code and tests

bcl:


bcl:
Are you running your tests with python2? It looked like you had used py3 with the buildstamp test, which I meant to mention earlier.
As for v0.py it's a work in progress and I'd rather not remove things.

atodorov:
> Are you running your tests with python2? It looked like you had used py3 with the buildstamp test, which I meant to mention earlier.

I think I might have cherry-picked this test without actually running it on Python 2. Anyway fixed in #265  now.

> As for v0.py it's a work in progress and I'd rather not remove things.

I understand your point but having these unused imports causes Jenkins to report failures and we don't actually execute the rest of the tests and have missed the above mishap. Ideally we should please pylint and have all of our testing PASS so that we know how a particular PR will affect the test results and our confidence in them. What can we do about that ?

atodorov:
All unit tests pass so +1 from me. Still need to do something about pylint.

bcl:
Pushed to lorax-composer the other day.

PR #265, opened 2017-10-24
Fix mocking on Python 2

atodorov:
- fix import of mock module
- account for io.StringIO differences in Python 2

### DIFF
--- a/tests/pylorax/test_buildstamp.py
+++ b/tests/pylorax/test_buildstamp.py
@@ -1,6 +1,6 @@
 import io
 import unittest
-from unittest.mock import patch
+from mock import patch
 
 from pylorax import buildstamp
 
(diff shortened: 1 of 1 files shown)


tests/pylorax/test_buildstamp.py
```diff
@@ -16,8 +16,12 @@ def setUp(self):
 
     def test_write_produces_file_with_expected_content(self):
         out_file = io.StringIO()
+        # in Python 2 write accepts unicode, not string
```
bcl: This is a quirk of io.StringIO, not Python2. io.StringIO was designed for py3. https://docs.python.org/2.7/library/io.html

atodorov: Fixed, I meant to document this properly but was experimenting with io.StringIO vs (c)StringIO.StringIO and forgot about it.

PR #266, opened 2017-10-24
Silence pocketlint bad-preconf-access warnings

atodorov:
Note: the problem that pocketlint is trying to warn us about lies
in YumBase._getConfig() method because it deletes the preconf
attribute from the object (self) which can cause exceptions!

- there is a YumBase.doConfigSetup() method so use it, even though
  it has been deprecated. There doesn't seem to be an alternative!
- doConfigSetup() doesn't accept releasever as parameter so we
  disable pylint for this line!

### DIFF
--- a/src/pylorax/api/yumbase.py
+++ b/src/pylorax/api/yumbase.py
@@ -14,6 +14,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
+# pylint: disable=bad-preconf-access
+
 import logging
 log = logging.getLogger("lorax-composer")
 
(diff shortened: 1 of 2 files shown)


bcl:
I'm pretty sure we had a good reason to use preconf, but I'll have to remember why. I'll look closer at this tomorrow.

atodorov:
@bcl  can we merge this one now ?

bcl:
I'd rather leave this as is (as well as the similar code in pylorax/api/yumbase.py
Once you touch yum's .conf attribute it does a bunch of setup that you cannot re-do or extend. So I think releasever will get ignored (or blow up if .preconf is removed).

atodorov:
In this case, just silence the warnings by disabling them.

PR #269, opened 2017-11-09
Add /recipes/info route to the API

bcl:


bcl:
Note this is based on PR #268

PR #273, opened 2017-11-11
Add /recipes/new and tests

bcl:


bcl:
Based on PR #270

src/pylorax/api/recipes.py
```diff
@@ -74,22 +74,22 @@ def toml(self):
         """Return the Recipe in TOML format"""
         return toml.dumps(self).encode("UTF-8")
 
-    def bump_version(self, new_version=None):
+    def bump_version(self, old_version=None):
         """semver recipe version number bump
 
-        :param new_version: An optional new version number
-        :type new_version: str
+        :param old_version: An optional new version number
```
dashea: This is an optional old version number now, right?

bcl: oops, yes. thanks.

PR #276, opened 2017-11-15
Composer service

bcl:
Based on previous PR #275

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -73,6 +73,23 @@ It also includes livemedia-creator which is used to create bootable livemedia,
 including live isos and disk images. It can use libvirtd for the install, or
 Anaconda's image install feature.
 
+%package composer
+Summary: Lorax Image Composer API Server
+# From EPEL
+Requires: python2-pytoml
+Requires: python-semantic_version
+Requires: libgit2
+Requires: libgit2-glib
+# From Distribution
+Requires: python-flask
+Requires: python-gevent
+
+%{?systemd_requires}
+BuildRequires: systemd
+
+%description composer
+lorax-composer provides a REST API for building images using lorax.
+
 %prep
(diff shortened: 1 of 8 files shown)


lorax.spec
```diff
@@ -73,6 +73,20 @@ It also includes livemedia-creator which is used to create bootable livemedia,
...
+Requires: python-semantic_version
+Requires: libgit2
+Requires: libgit2-glib
+# From Distribution
+Requires: python-flask
+Requires: python-gevent
+
+%description composer
+lorax-composer provides a REST API for building images using lorax.
+
```
dashea: According to the fedora wiki we need some scriptlets for the service, too. I think:

```
%post composer
%systemd_post lorax-composer.service

%preun composer
%systemd_preun lorax-composer.service

%postun composer
%systemd_postun_with_restart lorax-composer.service
```

plus

```
%{?systemd_requires}
BuildRequires: systemd
```

for the script Requires and the macro BuildRequires

lorax.spec
```diff
@@ -99,6 +114,11 @@ make DESTDIR=$RPM_BUILD_ROOT mandir=%{_mandir} install
 %{_datadir}/lorax/*
 %{_mandir}/man1/*.1*
 
+%files composer
+%{python_sitelib}/pylorax/api/*
+%{_sbindir}/lorax-composer
+%{_prefix}/lib/systemd/system/lorax-composer.service
```
dashea: Related to above, we can use %{_unitdir}/lorax-composer.service, which does appear to be valid for el7

lorax.spec
bcl: Thanks, for some reason I was thinking those would enable it, but it looks like they're using the presets files to make that decision.

PR #293, opened 2017-12-13
Make module/driver removal more generic (#1525417)

bcl:
Make it easier to add other arches with module and/or driver removal.
And add s390x to the list of arches to have modules and drivers removed.

Related: rhbz#1525417


bcl:
Note that the list of drivers for s390x is not finalized yet. It is there as an example.

bcl:
Currently not needed.

PR #294, opened 2017-12-14
Enable lorax-composer testing in Travis CI

atodorov:
After a comment from Martin Pitt I think we can still make use of Travis CI for things that are easier to test there and use Jenkins as a triggering system for scenarios which aren't possible in Travis or require internal infrastructure. This PR enables testing using Docker container built from the official CentOS 7 image.

Also I will need someone with more permissions to enable the builds at:
https://travis-ci.org/rhinstaller/lorax

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+sudo: required
+language: python
+services:
+  - docker
+
+script:
+  - make test-in-docker
+
+after_success:
+  - |
+
+        sudo docker create --name results-cont welder/lorax-composer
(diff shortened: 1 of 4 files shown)


Dockerfile.build
```diff
@@ -0,0 +1,21 @@
+FROM centos:7
```
bcl: Dockerfile.test would be a more accurate name for the file

atodorov:
Renamed the Dockerfile and merging.

PR #295, opened 2017-12-15
Add JWT authentication to lorax-composer

bcl:


martinpitt:
I sent this to the ML, but in case you prefer to discuss it here:

IMHO using actual PAM authentication here isn't really appropriate - it 
simultaneously requires too little (we don't want *all* users to be able to 
mess around with system-level stuff, only admins - similar to how e. g. docker
is restricted) and too much (we can't assume that we actually know or even have
a password - a lot of remote systems use SSH key auth).

I think it would be better for the daemon to create a nonce, e. g. 
/run/lorax-composer/token which is only accessible to root, or perhaps 
something like root:wheel or root:welder, and the client has to present that as
the token in order to use the API. That proves that the client has permissions
to read it off the server, is in the right groups etc., and avoids sending 
passwords around. AFAICS setup_jwt() already creates such a nonce 
(JWT_SECRET_KEY), so can this just be used directly?

PR #297, opened 2017-12-20
Switch the API to use a Unix Domain Socket

bcl:
This drops support for the TCP port and switches to using a socket at
/var/run/weldr/api.socket

Also add the start of some docs for lorax-composer.

--host and --port argument have been removed.

--group sets the group name to use for access to the socket and its
parent directory. Defaults to 'weldr'

--socket sets the full path to the socket to create. Defaults to
'/var/run/weldr/api.socket'

### DIFF
--- a/docs/lorax-composer.rst
+++ b/docs/lorax-composer.rst
@@ -1,0 +1,37 @@
+lorax-composer
+==============
+
+:Authors:
+    Brian C. Lane <bcl@redhat.com>
+
+lorax-composer is an API server that is compatible with the Weldr project's
+bdcs-api REST protocol. More information on Weldr can be found [on the Weldr
+blog](https://www.weldr.io).
+
+The server runs as root, and communication with it is via a unix domain socket
+(``/run/weldr/api.socket`` by default). The directory and socket are owned by
+root:weldr so that any user in the weldr group can use the API to control
+lorax-composer.
+
+When starting the server it will check for the correct permissions and
+ownership of pre-existing directory, or it will create a new one if none exist.
+The socket path and group owner's name can be changed from the cmdline by
+passing it the ``--socket`` and ``--group`` arguments.
+
+Logs
+----
+
+Logs are stored under ``/var/log/lorax-composer/`` and includes all console
+messages as well as extra debugging info and API requests.
+
+Quickstart
+----------
+
+1. Create a ``weldr`` group by running ``groupadd weldr``
+2. Remove any pre-existing socket directory with ``rm -rf /run/weldr/``
+   A new directory with correct permissions will be created the first time the server runs.
+3. Either start it via systemd with ``systemctl start lorax-composer`` or
+   run it directly with ``lorax-composer /path/to/recipes/``
+
+The ``/path/to/recipes/`` is where the recipe's git repo will be created, and all
+the recipes created with the ``/api/v0/recipes/new`` route will be stored.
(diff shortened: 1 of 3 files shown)


docs/lorax-composer.rst
```diff
@@ -0,0 +1,37 @@
+lorax-composer
+==============
+
+:Authors:
+    Brian C. Lane <bcl@redhat.com>
+
+lorax-composer is an API server that is compatible with the Weldr project's
+bdcs-api REST protocol. More information on Weldr can be found [on the Weldr
+blog](https://www.weldr.io)
```
martinpitt: Nitpick: Missing period at the end.

src/sbin/lorax-composer
```diff
@@ -44,10 +46,10 @@ def get_parser():
     parser = argparse.ArgumentParser(description="Lorax Composer API Server",
                                      fromfile_prefix_chars="@")
 
-    parser.add_argument("--host", default="127.0.0.1", metavar="HOST",
-                        help="Host or IP to bind to (127.0.0.1)")
-    parser.add_argument("--port", default=4000, metavar="PORT",
-                        help="Port to bind to (4000)")
+    parser.add_argument("--socket", default="/var/run/weldr/api.socket", metavar="SOCKET",
```
martinpitt: Is there a specific reason to still use the old `/var/run/` symlink? `/run` has been the [official definition](http://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#runRuntimeVariableData) for some time, and RHEL-7 supports it.

src/sbin/lorax-composer
```diff
@@ -129,6 +131,29 @@ if __name__ == '__main__':
...
+    except KeyError:
+        log.error("Missing group '%s'", opts.group)
+        sys.exit(1)
+
+    # Check the socket path to make sure it exists, and that ownership and permissions are correct.
+    socket_dir = os.path.dirname(opts.socket)
+    if not os.path.exists(socket_dir):
+        # Create the directory and set permissions and ownership
+        os.makedirs(socket_dir, 0o750)
+        os.chown(socket_dir, 0, gid)
```
martinpitt: I don't think it's necessary to restrict access to the directory. It's always a nuisance not to be able to tab-complete or ls directories when you are typing a sudo command and the like. The mere existence of the socket isn't a secret.

src/sbin/lorax-composer
```diff
@@ -149,8 +174,16 @@ if __name__ == '__main__':
...
 
-    log.info("Starting %s on %s:%d with recipes from %s", VERSION, opts.host, opts.port, opts.RECIPES)
-
-    http_server = WSGIServer((opts.host, opts.port), server, log=LogWrapper(server_log))
+    # Setup the Unix Domain Socket, remove old one, set ownership and permissions
+    if os.path.exists(opts.socket):
+        os.unlink(opts.socket)
+    listener = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    listener.bind(opts.socket)
+    os.chmod(opts.socket, 0o770)
```
martinpitt: Execute permissions are not necessary and confusing, so 0o660 is better.

src/sbin/lorax-composer
bcl: No specific reason. I'll change it.

src/sbin/lorax-composer
bcl: I think it's better to err on the side of safety. There's no good reason for anyone else to list the directory.

src/sbin/lorax-composer
bcl: Yeah, not sure why I did that.

PR #300, opened 2018-01-03
Add swaplabel command (#1530108)

poncovka:
It is required by libblockdev.

(cherry-picked from a commit 42322cc)

Resolves: rhbz#1530108


bcl:
@poncovka Does this need to go onto f27 and master? Or just f27?

poncovka:
Just F27. Master was fixed with https://github.com/rhinstaller/lorax/pull/253 .

bcl:
> Just F27. Master was fixed with #253 .

Ah, right! Thanks for the quick reply.

PR #303, opened 2018-01-15
Fix installpkg exclude operation

bcl:
Commit 8edaefd4d1726e added the ability to install specific NVR's of
packages, but it did not adjust the exclude operation to account for
this.

This patch fixes that, applying the exclude only to the name part of the
package NVR, and changes some variable names to pkgnvr/pkgnvrs to make
it more clear that the content has changed to <name>-<version>-<release>


ausil:
What would it take to get test cases written to cover these cases so we do not accidentally have issues going forward again?

bcl:
Lots of work. Master branch doesn't have much in the way of testing, and I currently don't have the time.

PR #306, opened 2018-01-24
Enable fedora-import-state service (#1537977)

rvykydal:
Used to import ifcfg files created in initramfs stage of installer.
Also see #1493479.


rvykydal:
I'd wait a bit before merging. Maybe the issue will be resolved soon properly by fixing the #1493479, ie updating the /usr/lib/systemd/system-preset/90-default.preset. Which would make sure that the fedora-import-state is enabled in installer image.

rvykydal:
This has been fixed in fedora-release package by adding fedora-import-state in default service presets.

PR #307, opened 2018-01-26
Don't try to build efi images for basearch=i386.

vathpela:
This shouldn't have been turned on when we switched to doing ia32-efi
images on x86_64; just having the file available isn't where we want
that policy decision to be.

Resolves: rhbz#1539085

Signed-off-by: Peter Jones <pjones@redhat.com>

### DIFF
--- a/share/templates.d/99-generic/live/x86.tmpl
+++ b/share/templates.d/99-generic/live/x86.tmpl
@@ -76,7 +76,7 @@ hardlink ${KERNELDIR}/initrd.img ${BOOTDIR}
 %if exists("boot/efi/EFI/*/gcdx64.efi"):
     <% efiarch64 = 'X64' %>
 %endif
-%if efiarch32 or efiarch64:
+%if efiarch32 or efiarch64 and basearch != 'i386':
     <%
     efigraft="EFI/BOOT={0}/EFI/BOOT".format(outroot)
     images = ["images/efiboot.img"]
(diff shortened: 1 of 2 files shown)


AdamWill:
Sorry, but, [didn't work](https://kojipkgs.fedoraproject.org//work/tasks/3238/24593238/mock_output.log).

Aaand, I think I know why. Assuming whatever ultimately interprets these things is Python or something else that behaves like it (isn't this a mathematical convention or something?):

    [adamw@xps13k lorax (master)]$ python3
    Python 3.6.4 (default, Jan 17 2018, 12:00:56) 
    [GCC 7.2.1 20170915 (Red Hat 7.2.1-2)] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> if 1 == 1 or 1 == 2 and 3 == 4:
    ...     print("True!")
    ... 
    True!
    >>> if (1 == 1 or 1 == 2) and 3 == 4:
    ...     print("True!")
    ... 
    >>> 

That is, this is being interpreted as:

    `%if (efiarch32) or (efiarch64 and basearch != 'i386'):`

and thus isn't doing us any good at all.

bcl:
Yeah, it's a Mako template that uses python-like syntax in the % lines. Could you throw some () in there and give it a test?

AdamWill:
Sure, I'll try and do that in a bit. Or else could just go with my initial suggestion of changing it a few lines earlier from:

    %if exists("boot/efi/EFI/*/gcdia32.efi"):

to:

    %if exists("boot/efi/EFI/*/gcdia32.efi") and $basearch != 'i386':

I suppose. (That way `efiarch32` never gets set).

AdamWill:
https://github.com/rhinstaller/lorax/pull/330

PR #309, opened 2018-02-01
Initialize the available dnf plugins

pmazzini:


bcl:
I'm not sure we really want to do this. The dnf instance used by lorax is somewhat disconnected from that of the host and I'm not sure what kinds of odd effects might show up if we start using all the host's installed plugins.

pmazzini:
Same as in #309, there are failing tests on some glusterfs dependency which seems unrelated.

bcl:
> Same as in #309, there are failing tests on some glusterfs dependency which seems unrelated.

Right, ignore that for now, I expect that the test on rawhide needs to have the expected version adjusted.

This change appears simple, but *could* potentially impact users in unexpected ways. Before simply accepting a commit I need to make sure that it either 'Just Works' or needs to be switched on with yet another cmdline argument (which I'd prefer not to do if possible).

jaymzh:
OK, so what's the path forward? What does making sure look like? Is that some tests you want to run? A survey of all lorax users? If we add a command-line flag around this (lorax doesn't have that many), would that be acceptable? I'm trying to understand the path to getting this bug fixed.

bcl:
Ok, have a look at #383 
I decided it was safer to make this opt-in, and to allow you to enable all of them, or a specific set of them. Give it a try and let me know what you think.

jaymzh:
Commented on that PR, but that's perfect, thanks. But also to echo what I said there, anytime you know what you want the changes to a PR to be, don't hesitate to just let us know, and we're happy (assuming it's not "refactor all the code !!") to do it.

bcl:
Fixed by #383

PR #313, opened 2018-02-05
Add /composer/info API route

bcl:


bcl:
Based on PR #312

bcl:

PR #315, opened 2018-02-08
Add /compose/cancel API

bcl:
Based on PR #314


bcl:
Merged into lorax-composer branch

PR #316, opened 2018-02-08
Add /compose/log API

bcl:


bcl:
Based on PR #315

bcl:
Merged into lorax-composer branch

PR #318, opened 2018-02-09
Add more tests for api.recipes

atodorov:
- test against already existing git repository
- test commit_recipe_file() handling of IOError
- update tests for commit_recipe_directory()
  - add asserts on the existing test. Not raising an exception
    isn't enough!
  - add test which exercises the method under test while it handles
    exceptions raised by other methods
- test for list_commits() when the underlying calls fail to convert
  timestamp
- test for find_name() when `name' is not on the list
- tests for get_revision_from_tag()

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git/
+.coverage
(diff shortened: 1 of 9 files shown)


atodorov:
Merging. failing Jenkins builds are because of pylint warnings which we ignore in Travis.

PR #320, opened 2018-02-10
compose tests

bcl:
Based on PR #316


bcl:

PR #323, opened 2018-02-15
Check for a source tree doc install first, not second.

bcl:
This way a source install and test run will always take precedence over
a system installed version.


atodorov:
@bcl do we still want to alter the test to verify a non 404 response and disregard the actual content ?

bcl:
> @bcl do we still want to alter the test to verify a non 404 response and disregard the actual content ?

I think so, since there's no reason to be overly strict. Check that it is a 200, looks 'long' and maybe has a few keywords in it and call that good.

PR #327, opened 2018-02-21
add system-logos dependency for syslinux

nullr0ute:
The x86.tmpl explicitly references the syslinux-splash provided in
the system-logos package and fails if it's not there so implicitly
install it on arches where syslinux is supported to ensure it's
there. Fixes rhbz #1529239

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -50,7 +50,7 @@ installpkg glibc-all-langpacks
     installpkg shim-ia32 grub2-efi-ia32-cdboot
 %endif
 %if basearch in ("i386", "x86_64"):
-    installpkg biosdevname memtest86+ syslinux
+    installpkg biosdevname memtest86+ syslinux system-logos
     installpkg grub2-tools grub2-tools-minimal grub2-tools-extra
 %endif
 %if basearch in ("ppc", "ppc64", "ppc64le"):


dustymabe:
I think this broke our builds in rawhide. Somehow the installer now wants to install both generic-logos and fedora-logos and can't figure it out. This was with `lorax-28.6-1.fc29.x86_64` (just built a few hours ago). See [koji task](https://koji.fedoraproject.org/koji/taskinfo?taskID=25243479) and [root.log](https://kojipkgs.fedoraproject.org//work/tasks/3479/25243479/root.log)

```
DEBUG util.py:439:  dnf.exceptions.DepsolveError: 
DEBUG util.py:439:   Problem: package generic-logos-18.0.0-5.fc28.noarch conflicts with fedora-logos provided by fedora-logos-28.0.2-3.fc28.x86_64
DEBUG util.py:439:    - conflicting requests
```

nullr0ute:
How does it handle fedora-release/repos etc  then?

dustymabe:
> How does it handle fedora-release/repos etc then?

I don't know. but I just tested lorax with and without this patch and it works *without* it.

AdamWill:
`fedora-release` is not actually explicitly listed in lorax; it gets pulled in as a dependency of something else (I think usually `setup`). I guess this accounts for the difference; we'd have to look at exactly what `installpkg` means.

bcl:
For the record, this is how lorax selects *-release and the corresponding *-logos:

https://github.com/rhinstaller/lorax/blob/master/src/pylorax/treebuilder.py#L112

AdamWill:
well, that's not how it *selects* a `-release` package, is it? That's how it selects a `-logos` package corresponding to a provider of `system-release` if there *is* a provider of system-release in the transaction. I don't *think* lorax has anything that explicitly pulls in a -release package at all, unless I'm missing it somewhere. I believe we just always wind up with one in our images due to at least the `setup` dependency and probably other dependency chains in the set of packages that lorax does explicitly pull in.

bcl:
It installs whatever non-generic release package provides /etc/system-release, and the corresponding logos package.

https://github.com/rhinstaller/lorax/blob/master/src/pylorax/treebuilder.py#L96

If it can't find one it will fail completely.

AdamWill:
oh yes, I missed that it actually *selects* the package too. Well, tries to. It appears not to be working, for some reason.

PR #333, opened 2018-02-23
Revert system-logos

bcl:
There appears to be some bugs with DNF, but those can wait. This should work fine as long as no updates repo is used.

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -104,6 +104,7 @@ Summary:  livemedia-creator no-virt dependencies
 Requires: lorax = %{version}-%{release}
 Requires: anaconda-core
 Requires: anaconda-tui
+Requires: system-logos
 
 %description lmc-novirt
 Additional dependencies required by livemedia-creator when using it with --no-virt
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -50,7 +50,7 @@ installpkg glibc-all-langpacks
     installpkg shim-ia32 grub2-efi-ia32-cdboot
 %endif
 %if basearch in ("i386", "x86_64"):
-    installpkg biosdevname memtest86+ syslinux system-logos
+    installpkg biosdevname memtest86+ syslinux
     installpkg grub2-tools grub2-tools-minimal grub2-tools-extra
 %endif
 %if basearch in ("ppc", "ppc64", "ppc64le"):
--- a/src/pylorax/treebuilder.py
+++ b/src/pylorax/treebuilder.py
@@ -93,7 +93,8 @@ class RuntimeBuilder(object):
         release = None
         q = self.dbo.sack.query()
         a = q.available()
-        for pkg in a.filter(provides='/etc/system-release'):
+        for pkg in a.filter(provides='system-release'):
+            logger.debug("Found release package %s", pkg)
             if pkg.name.startswith('generic'):
                 continue
             else:


AdamWill:
:+1:

dustymabe:
LGTM

AdamWill:
LGTM too.

For the record I was curious about why this doesn't work, so I dug a bit. There is a surprisingly icky codepath behind this, yay. dnf `Sack`s are subclasses of hawkey `Sack`. hawkey-the-python-module still exists, even though hawkey-the-project officially doesn't: it's now provided by [libdnf](https://github.com/rpm-software-management/libdnf) as a Python wrapper around a C++ implementation. `dnf.Sack.query()` is actually implemented in the dnf child class, but it ultimately winds up right back in hawkey/libdnf: it returns a `dnf.query.Query()`, but `dnf.query` imports `Query` from hawkey, so we're [here](https://github.com/rpm-software-management/libdnf/blob/869d3ccf4827069d6f07514da30acb36ff0b2fc1/python/hawkey/query-py.cpp), where `hawkey.Query` is implemented (technically `_hawkey.Query`, but don't worry about that). `filter` is implemented [here](https://github.com/rpm-software-management/libdnf/blob/869d3ccf4827069d6f07514da30acb36ff0b2fc1/python/hawkey/query-py.cpp#L540), which mostly just calls `filter_internal` [here](https://github.com/rpm-software-management/libdnf/blob/869d3ccf4827069d6f07514da30acb36ff0b2fc1/python/hawkey/query-py.cpp#L426), which...does a whole bunch of stuff, but I think on our path winds up calling `filter_add` and we wind up with keyname `HY_PKG_PROVIDES` [here](https://github.com/rpm-software-management/libdnf/blob/869d3ccf4827069d6f07514da30acb36ff0b2fc1/python/hawkey/query-py.cpp#L355), and call `query->addFilter(keyname, reldeplist);`. `query` there is a `HyQuery`, which via [hy-types.h](https://github.com/rpm-software-management/libdnf/blob/3f42de656a2eeb05e3f1d4f479e45a9c7b2363f5/libdnf/hy-types.h#L29) takes us [here](https://github.com/rpm-software-management/libdnf/blob/f5916d937134f2da30af76467a232c795c5d4415/libdnf/sack/query.cpp#L526) - to `libdnf::Query`, and its forest of [overloaded addFilter methods](https://github.com/rpm-software-management/libdnf/blob/f5916d937134f2da30af76467a232c795c5d4415/libdnf/sack/query.cpp#L626). I *think* we probably wind up in [this one](https://github.com/rpm-software-management/libdnf/blob/f5916d937134f2da30af76467a232c795c5d4415/libdnf/sack/query.cpp#L671), which calls [reldep_from_str](https://github.com/rpm-software-management/libdnf/blob/3f42de656a2eeb05e3f1d4f479e45a9c7b2363f5/libdnf/hy-iutil.cpp#L676), which calls [dnf_reldep_new](https://github.com/rpm-software-management/libdnf/blob/b3e38e661cf7d124d22751f3d398077dbddc192e/libdnf/dnf-reldep.cpp#L88), which calls pool_str2id, which...

YOU KNOW WHAT SCREW THIS WHAT THE HELL DNF HOW IS ANYONE SUPPOSED TO UNDERSTAND HOW THE HELL YOU WORK

bcl:
Thanks for the report Adam :)

I pushed a slightly different fix, using the provides as it is meant to be instead of file. I don't see any reason to continue using that. Tested in mock and it works.

AdamWill:
I was also curious if there was any real reason why the code uses the file name not the virtual provide. That was [introduced in this 2012 commit](https://github.com/rhinstaller/lorax/commit/910c54e3946f8eaec409e9c6732ed6e5fcb1f9b4), but I can't find a plausible reason for it. Both the file and the provide have been in Fedora since at least Fedora 10 in 2008. The file was added to RHEL in June 2009 and the virtual provide only in April 2010, but that's still two years before this commit happened. So I don't have a great theory. :P

AdamWill:
So, I was also curious about why adding the `installpkg` line caused the conflict in the first place. This was similarly difficult to debug, but I at least got *somewhere*. By hacking up lorax a bit to log stuff, I managed to determine that basically doing this:

    installpkg system-logos

*itself* adds both fedora-logos and generic-logos to the transaction. More specifically, it causes lorax to do this:

    pkgnames = [pkg for pkg in dnf.subject.Subject(p).get_best_query(self.dbo.sack).filter(latest=True)]

where `p` is `system-logos`, and what dnf gives us back there is *both* packages. So, the question I guess is whether this is what `get_best_query` is supposed to do, and whether it's what we intend it to do: is `installpkg foo` supposed to install *all possible providers* of `foo`, or just one of them? I recall that we've come up against a similar question before, but don't recall the exact context.

dustymabe:
> dnf gives us back there is both packages.

I beat you on this one by a whole [12 hours](https://github.com/rhinstaller/lorax/issues/332):

```
If I add a logger.info(pkgnvrs) at line 580 I can see ['fedora-logos-28.0.2-3.fc28', 'generic-logos-18.0.0-5.fc28'] in the output. i.e. it selects BOTH packages to install.
```

AdamWill:
hey, that quote could come from *anywhere*...

PR #335, opened 2018-03-10
Add composer-cli utility

bcl:

PR #337, opened 2018-03-15
Some docs updates for lorax-composer and composer-cli

bcl:


bcl:
Based on PR #335 so lots of other commits.

bcl:
Rebased onto PR #338

PR #339, opened 2018-03-16
Add Travis support for running make test

bcl:


atodorov:
@bcl you are missing nose as requirement

bcl:
> @bcl you are missing nose as requirement

I updated it with nose and it still fails. Even though it says nose is installed.

    Requirement already satisfied: nose in /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages (from -r test-requirements.txt (line 2))

atodorov:
hmmm, the lorax repo seems to have been deactivated on Travis. I don't have enough permissions to see why.

atodorov:
I believe this is the problem `PYTHON ?= /usr/bin/python` (at the top of Makefile). You are using system Python instead of the one from the virtualenv. Try `/usr/bin/env python` instead.

atodorov:
A slightly different version of this is included in #352. Closing.

PR #347, opened 2018-04-05
Include the anaconda-install-env-deps metapackage

M4rtinK:
Use the anaconda-install-env-deps metapackage to pull in the
Anaconda dependencies needed in the installation environment.

The anaconda-install-env-deps metapackage lists all install time
dependencies and makes it possible for packages such as
Initial Setup to depend on Anaconda without pulling all
the (mainly storage related) install time dependencies
to the installed system.

The same is applicable for dirinstall which also does
not require the install time dependencies as it is just
installing to a local folder.

Also drop the tmux and gdb dependencies from the template as
both have been added to the metapackage to make install time
dependency tracking more consistent.

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -2,11 +2,7 @@
 <%page args="basearch, product"/>
 
 ## anaconda package
-installpkg anaconda anaconda-widgets kexec-tools-anaconda-addon
-## anaconda deps that aren't in the RPM
-installpkg tmux
-## for anaconda crash handling
-installpkg gdb
+installpkg anaconda anaconda-widgets kexec-tools-anaconda-addon anaconda-install-deps
 ## Other available payloads
 installpkg dnf
 installpkg rpm-ostree ostree


M4rtinK:
For some context about this PR - it's part of an ongoing effort to separate the install-time only dependencies of anaconda, so that packages such as Initial Setup, which only use the Anaconda configuration screens can operate normally without pulling in a lot of unrelated dependencies.

This is actually pretty important on ARM devices, where in most cases Initial Setup basically takes over the role of the installer. Also storage space usage can be a limiting factor on many ARM devices.

Related PRs:

Anaconda PR: https://github.com/rhinstaller/anaconda/pull/1415
-> this PR adds the anaconda-install-env-deps metapackage

Fedora kickstarts PS: https://pagure.io/fedora-kickstarts/pull-request/375
-> includes anaconda-install-env-deps in the kickstart used to generate the Fedora live images

Related bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1491806 - make it possible to install Initial Setup without unnecessary dependencies
https://bugzilla.redhat.com/show_bug.cgi?id=1561047 - udisks2-iscsi package needs to be present in the installation environment

M4rtinK:
> Looks like a great idea!

Thanks! 

BTW, I guess we will need to do a coordinated Anaconda & Lorax release with this change once the Anaconda & Fedora kickstarts PR are merged to prevent possible breakage (eq. referring to non-existent anaconda-install-env deps or anaconda-install-env-deps not being used).

For F28 - a combined Bodhi update I guess.
For Rawhide - chainbuild ?

PR #349, opened 2018-04-19
 drop support for creating 32-bit ppc installer media

sharkcz:
We don't build any 32-bit ppc stuff for long time, so clean the unused stuff from lorax template too. Because now ppc and ppc64le template are very close, I'll try to merge them in a next step.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1566225


AdamWill:
@bcl could you do a rawhide package build with this change? Until that happens, ppc64 Rawhide composes are going to be broken. Thanks!

sharkcz:
@bcl ping, we have also the ppc64le composes broken after retiring ppc64-utils ...

bcl:
Here you go - https://koji.fedoraproject.org/koji/taskinfo?taskID=26752257

sharkcz:
Thanks, and let's wait for composes tomorrow :-)

PR #353, opened 2018-04-26
Add support for systemd socket activation

bcl:
Instead of enabling lorax-composer.service enable lorax-composer.socket
and it will start lorax-composer on first access to
/run/weldr/api.socket


bcl:
Pushed to master and lorax-composer after fixing typo in setup.py

PR #366, opened 2018-05-25
Make LogRequestHandler configurable

M4rtinK:
Make it possible to manipulate the simple and regexp
tests the LogRequestHandler class uses to check error
messages for potential error states.

This is accomplished by moving the simple and regexp test
string to contstants and exposing current test and regexp
list via class attributes. It's also now possible to set
the log request handler class for a LogMonitor.

This funactionality can then be used for example like this:

customized_log_request_handler = monitor.LogRequestHandler
customized_log_request_handler.simple_tests.remove(monitor.TEST_CALL_TRACE)
log_monitor = monitor.LogMonitor(install_log,
                                 timeout=opts.timeout,
                                 log_request_handler_class = customized_log_request_handler)

This way installation will continue even if there was a call
trace in the logs. In a similar way additional tests and regexps can be
also added.

### DIFF
--- a/src/pylorax/monitor.py
+++ b/src/pylorax/monitor.py
@@ -34,6 +34,26 @@ class LogRequestHandler(socketserver.BaseRequestHandler):
     for patterns that would indicate that the installation failed.
     self.server.log_error is set True when this happens.
     """
+
+    simple_tests = [
+        "Traceback (",
+        "traceback script(s) have been run",
+        "Out of memory:",
+        "Call Trace:",
+        "insufficient disk space:",
+        "Not enough disk space to download the packages",
+        "error populating transaction after",
+        "crashed on signal",
+        "packaging: Missed: NoSuchPackage",
+        "packaging: Installation failed",
+        "The following error occurred while installing.  This is a fatal error"
+    ]
+
+    re_tests = [
+        r"packaging: base repo .* not valid",
+        r"packaging: .* requires .*"
+    ]
+
     def setup(self):
         """Start writing to self.server.log_path"""
 
@@ -99,25 +119,13 @@ class LogRequestHandler(socketserver.BaseRequestHandler):
         """
         if "IGNORED" in line:
             return
-        simple_tests = ["Traceback (",
-                        "Out of memory:",
-                        "Call Trace:",
-                        "insufficient disk space:",
-                        "Not enough disk space to download the packages",
-                        "error populating transaction after",
-                        "traceback script(s) have been run",
-                        "crashed on signal",
-                        "packaging: Missed: NoSuchPackage",
-                        "packaging: Installation failed",
-                        "The following error occurred while installing.  This is a fatal error"]
-        re_tests =     [r"packaging: base repo .* not valid",
-                        r"packaging: .* requires .*"]
-        for t in simple_tests:
+
+        for t in self.simple_tests:
             if t in line:
                 self.server.log_error = True
                 self.server.error_line = line
                 return
-        for t in re_tests:
+        for t in self.re_tests:
             if re.search(t, line):
                 self.server.log_error = True
                 self.server.error_line = line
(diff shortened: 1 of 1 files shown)


src/pylorax/monitor.py
```diff
@@ -168,7 +191,7 @@ class LogMonitor(object):
     This needs to be running before the virt-install runs, it expects
     there to be a listener on the port used for the virtio log port.
     """
-    def __init__(self, log_path=None, host="localhost", port=0, timeout=None):
+    def __init__(self, log_path=None, host="localhost", port=0, timeout=None, log_request_handler_class=None):
```
poncovka: You can use `LogRequestHandler` instead of None.

src/pylorax/monitor.py
```diff
@@ -99,25 +134,13 @@ def iserror(self, line):
...
-                        "traceback script(s) have been run",
-                        "crashed on signal",
-                        "packaging: Missed: NoSuchPackage",
-                        "packaging: Installation failed",
-                        "The following error occurred while installing.  This is a fatal error"]
-        re_tests =     [r"packaging: base repo .* not valid",
-                        r"packaging: .* requires .*"]
-        for t in simple_tests:
+
+        for t in LogRequestHandler.simple_tests:
```
poncovka: Please, use `self.simple_tests` here to use the simple tests of the current class.

src/pylorax/monitor.py
```diff
@@ -99,25 +134,13 @@ def iserror(self, line):
...
             if t in line:
                 self.server.log_error = True
                 self.server.error_line = line
                 return
-        for t in re_tests:
+        for t in LogRequestHandler.re_tests:
```
poncovka: Same here.

src/pylorax/monitor.py
M4rtinK: Good point! Changed.

src/pylorax/monitor.py
M4rtinK: Fixed.

src/pylorax/monitor.py
M4rtinK: Fixed.

M4rtinK:
PR request updated, should be now ready for review from one of the maintainers (@bcl, etc.). :)

M4rtinK:
> I'm not sure why this would be needed. It makes it harder to maintain the lists, and I don't know why you would want to customize them, I think they're all pretty clear indicators that the installation has failed.

Lorax is used as part of the machinery that runs the [Anaconda kickstart test suite](https://github.com/rhinstaller/kickstart-tests) and we are currently hitting an issue on Rawhide, where a bogus error shows up in the logs when RPM scriptlets for the kernel package are running. While this is likely a valid error from the image generation PoV (the system might not be bootable or have other kernel related issues), it is not a problem for our kickstart tests.

We need a way to manipulate the list of strings and regexps Lorax checks for in the logs or else we would not be able to get any testing done one Rawhide at the moment. Other similar false positives (at least from the point of view of kickstart tests) are likely to show up as well.

We would also like to be able to add some custom error triggers at runtime, possibly even per test-case, to detect stuff like specific test hangs or to enable running tests that are expected not to run to the end.

Also if you think the current solution is too heavy weight, we can definitely come up with something more lightweight if needed.

bcl:
Ah, ok. That does make sense. I'd like to come up with a cleaner way to manage these strings than giving them all names and then adding them to the lists. I'll think about it for a bit.

bcl:
I think I'd like this better if we left the tests as lists of strings. Move them to be class members, and then you can override the class with your own custom set of them instead of calling remove on them. Does that make sense?

poncovka:
> I think I'd like this better if we left the tests as lists of strings. Move them to be class members, and then you can override the class with your own custom set of them instead of calling remove on them. Does that make sense?

That sounds reasonable.

M4rtinK:
OK then, I'll update the PR to use the new way. :)

M4rtinK:
Finally updated the PR to the simpler class member based version suggested by @bcl. :)

PR #379, opened 2018-06-16
Initialize DNF plugins

jaymzh:


jaymzh:
Those tests are failing on some glusterfs dependency which should be unrelated.

bcl:
This is a dupe of #309, my concerns haven't been addressed.

PR #386, opened 2018-06-23
systemd: Create necessary directories for lorax-composer

stefwalter:
Use ExecPreStart to create the necessary directories for lorax-composer
to function. These are created with appropriate permissions and group
access for lorax-composer.


bcl:
Fixed by #389

PR #387, opened 2018-06-25
Make sure /run/weldr has correct ownership and permissions

bcl:
Normally tmpfiles.d will handle this at boot time, but if you install
lorax-composer without rebooting it was ending up with root:root
ownership instead of root:weldr


wgwoods:
Looks fine. I do think that `RuntimeDirectory=/run/weldr` is _intended_ for this use case - but if `lorax-composer` isn't in a place where it can run correctly with `Group=weldr` here then this'll do.

PR #388, opened 2018-06-26
DNF 3: progress callback constants moved to dnf.transaction

AdamWill:


AdamWill:
@sgallagher

PR #390, opened 2018-06-28
More help

clumens:
Add a couple patches that make seeing the help output a little less annoying.  Basically, recognize "help" as a subcommand and also let subcommands display their own help output.


clumens:
Whoops, I opened this against the wrong branch.  I don't see any way to correct that here, so I'm just going to open a new PR.

PR #391, opened 2018-06-28
lorax-composer: Update documentation

bcl:
This updates the documentation to include composer-cli instead of curl,
warnings about SELinux and needing the RHEL 7 Optional package repo, as
well as other documentation improvements from the master branch.

### DIFF
--- a/docs/composer-cli.rst
+++ b/docs/composer-cli.rst
@@ -1,0 +1,62 @@
+composer-cli
+============
+
+:Authors:
+    Brian C. Lane <bcl@redhat.com>
+
+``composer-cli`` is used to interact with the ``lorax-composer`` API server, managing blueprints, exploring available packages, and building new images.
+
+It requires `lorax-composer <lorax-composer.html>`_ to be installed on the
+local system, and the user running it needs to be a member of the ``weldr``
+group. They do not need to be root, but all of the `security precautions
+<lorax-composer.html#security>`_ apply.
+
+composer-cli cmdline arguments
+------------------------------
+
+.. argparse::
+   :ref: composer.cli.cmdline.composer_cli_parser
+   :prog: composer-cli
+
+Edit a Blueprint
+----------------
+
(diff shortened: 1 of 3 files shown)


docs/composer-cli.rst
```diff
@@ -0,0 +1,62 @@
...
+Edit a Blueprint
+----------------
+
+Start out by listing the available blueprints using ``composer-cli blueprints
+list``, pick one and save it to the local directory by running ``composer-cli
+blueprints save http-server``. If there are no blueprints available you can
+copy one of the examples `from the test suite
+<https://github.com/weldr/lorax/tree/master/tests/pylorax/blueprints/>`_.
+
+Edit the file (it will be saved with a .toml extension) and chance the
```
clumens: s/chance/change/

docs/lorax-composer.rst
```diff
@@ -4,61 +4,198 @@ lorax-composer
...
+The ``name`` field is the name of the blueprint. It can contain spaces, but they will be converted to ``-``
+when it is written to disk. It should be short and descriptive.
+
+``description`` can be a longer description of the blueprint, it is only used for display purposes.
+
+``version`` is a `semver compatible <https://semver.org/>`_ version number. If
+a new blueprint is uploaded with the same ``version`` the server will
+automatically bump the PATCH level of the ``version``. If the ``version``
+doesn't match it will be used as is. eg. Uploading a blueprint with ``version`` set to ``0.0.1`` when the existing blueprint ``version`` is ``0.0.1`` will result in the new blueprint being stored as ``version 0.0.1``.
+
```
clumens: I think you meant this to say something like "eg. Uploading a blueprint with version set to 0.1.0 when the existing blueprint version is 0.0.1 will result in the new blueprint being stored as version 0.1.0."  That is, the first and last version numbers shouldn't be the same as the middle.

bcl: oh, ha, exactly :)

PR #392, opened 2018-07-12
Improve depsolve error reporting

bcl:
Include individual package names when things fail.
Depsolve the compose template packages at startup and include failures in /api/status in a 'msgs' field that is a list of strings.


bcl:
This will also need to be ported to the dnf based branches.

PR #393, opened 2018-07-12
Add group support to blueprints

clumens:
There's only some tests right now, but more could be added as needed.


clumens:
Pushed this the old fashioned way.

PR #399, opened 2018-07-25
Add timestamps to "compose-cli compose status" output.

clumens:


clumens:
Output looks like this:
```
[root@ibm-p8-kvm-03-guest-02 ~]# composer-cli compose status
16f9ad8d-d0d8-4de7-b09c-76edad1eb792 FINISHED webserver       0.0.2 live-iso         592445440 1530118046.63                                                                                
a3b1d944-b594-4dc0-881b-bda60bb7a7c6 FINISHED webserver       0.0.2 tar              318555432 1530047141.7  
```
I'm somewhat concerned the timestamp will get lost over next to that other number, though.  Maybe putting it between status and blueprint would look better?

PR #406, opened 2018-07-31
clean up "list" commands and add bash completion

wgwoods:
Like #405, these clean up the various composer-cli "list" commands and then add /etc/bash_completion.d/composer-cli, which does fancy tab completion to make composer-cli funner to use. Such fun. Man.

(This is the rhel7 version)


wgwoods:
HOMPH DONK BLUH HOW DO I GIT

Issue #409, opened 2018-07-31
Add queue timestamps to compose API


bcl:
Instead of one timestamp, provide 3:
  - job_created (user submitted the new compose)
  - job_started (anaconda started running)
  - job_finished (entered FAILED or FINISHED) state

clumens:
This should be fixed now - see the patches for exactly which elements you should be looking for.  If something's still busted, please open a new ticket and we'll take a look.

PR #412, opened 2018-08-02
Add more timestamps (#409).

clumens:


src/pylorax/api/queue.py
```diff
@@ -249,15 +253,16 @@ def compose_detail(results_dir):
...
     * queue_status - The final status of the composition (FINISHED or FAILED)
-    * timestamp - The time of the last status change
+    * job_created - When the user submitted the compose
+    * job_started - Anaconda started running
+    * job_finished - Job entered FINISHED or FAILED state
     * compose_type - The type of output generated (tar, iso, etc.)
     * blueprint - Blueprint name
     * version - Blueprint version
     * image_size - Size of the image, if finished. 0 otherwise.
     """
```
bcl: Could you add a note here about the timestamps being Unix UTC timestamps, and that any or all of them could be None (or null in JSON) if they haven't been created yet.

PR #413, opened 2018-08-03
Return an error if /modules/info doesn't return anything.

clumens:
Unfortunately, this isn't very useful if /modules/info is provided with
multiple modules.  yum doesn't traceback when doPackageLists is given
something that doesn't exist.  It just returns an empty list.  If
/modules/info is given just one module and yum gives us an empty list,
it's easy to say what happened.  If /modules/info is given several
modules and just one does not exist, we will not be able to detect that.

Fixing this would require doing more yum operations, which is likely to
slow things down and isn't the direction I want to be going.

### DIFF
--- a/src/pylorax/api/v0.py
+++ b/src/pylorax/api/v0.py
@@ -1560,6 +1560,11 @@ def v0_api(api):
             log.error("(v0_modules_info) %s", str(e))
             return jsonify(status=False, errors=[str(e)]), 400
 
+        if not modules:
+            msg = "one of the requested modules does not exist: %s" % module_names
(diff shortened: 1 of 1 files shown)


clumens:
Does this need to return an error id?

larskarlitski:
> Does this need to return an error id?

Yes. weldr-web needs to be able to disambiguate some error cases and that's hard to do if it has to parse human readable strings.

I'll open an issue in a bit with more examples where we would need ids. Let's discuss how best to this this on there.

clumens:
Alright, I'll deal with the error ID separately from this patch, as it's going to require work in a variety of places.  I can merge this and then just have another site to look at.

PR #415, opened 2018-08-03
Return a JSON error instead of a 404 on certain malformed URLs.

clumens:
This handles the case where a route is requested, but without a required
parameter.  So, /blueprints/info is requested instead of
/blueprints/info/http-server.  It accomplishes this via a decorator, so
a lot of these route-related functions now have quite a few decorators
attached to them.

Typo'd URLs (/blueprints/nfo for instance) will still return a 404.  I
think this is a reasonable thing to do.

### DIFF
--- a/src/pylorax/api/v0.py
+++ b/src/pylorax/api/v0.py
@@ -969,6 +969,7 @@ from flask import jsonify, request, Response, send_file
 import pytoml as toml
 
 from pylorax.sysutils import joinpaths
+from pylorax.api.checkparams import checkparams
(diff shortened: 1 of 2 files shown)


src/pylorax/api/v0.py
```diff
@@ -998,6 +999,28 @@ def take_limits(iterable, offset, limit):
     """
     return iterable[offset:][:limit]
 
+# A decorator for checking the parameters provided to the API route implementing
```
bcl: This should go into a library module so that the future v1.py can use it as well.

PR #422, opened 2018-08-08
Move disklabel and UEFI support to compose.py

bcl:
Currently we are making MBR disk images for qcow2 and partitioned disk,
so the UEFI packages aren't required at this point.

Move the clearpart command into compose.py so that in the futute it can
use clearpart --disklabel to create a GPT image, and add the required
packages to the package set.


bcl:
Pushed the 1st commit, 2nd needs to be coordinated with UI changes.

PR #433, opened 2018-08-16
Arm and other cleanups

nullr0ute:
Some changes needed for UEFI support on ARMv7 as well as some small package updates

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -30,11 +30,15 @@ installpkg glibc-all-langpacks
 
 ## arch-specific packages (bootloaders etc.)
 %if basearch == "aarch64":
-    installpkg efibootmgr grub2-efi-aa64-cdboot grubby shim-aa64
+    installpkg grub2-tools-efi
+    installpkg efibootmgr
+    installpkg grub2-efi-aa64-cdboot shim-aa64
 %endif
 %if basearch in ("arm", "armhfp"):
+    installpkg grub2-tools-efi
+    installpkg efibootmgr
+    installpkg grub2-efi-arm-cdboot
     installpkg kernel-lpae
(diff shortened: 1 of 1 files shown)


share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -68,9 +68,6 @@ installpkg plymouth
 ## extra dracut modules
 installpkg anaconda-dracut dracut-network dracut-config-generic dracut-fips
 
-## import-state.service for switchroot
-installpkg initscripts
```
bcl: NAK on this line. See https://bugzilla.redhat.com/show_bug.cgi?id=1615332

share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -30,11 +30,15 @@ installpkg glibc-all-langpacks
 
 ## arch-specific packages (bootloaders etc.)
 %if basearch == "aarch64":
-    installpkg efibootmgr grub2-efi-aa64-cdboot grubby shim-aa64
```
bcl: Are you sure grubby doesn't need to be here? Have you tested this template?

share/templates.d/99-generic/runtime-install.tmpl
nullr0ute: There was no comment from the initscripts maintainer as to why it was needed, and there's also no documentation there as to what broke, it just says "some network configuration use cases are broken" which isn't useful.

Can you merge the arm patch without the other one?

share/templates.d/99-generic/runtime-install.tmpl
nullr0ute: grubby is explicitly there for all architectures on line 16 above that

bcl:
Picked the arm patch to master.

Issue #437, opened 2018-08-17
composer-cli cannot be run from non-writable directories


larskarlitski:
Running composer-cli in a non-writable directory crashes with:
```
Traceback (most recent call last):
  File "/usr/bin/composer-cli", line 89, in <module>
    setup_logging(opts.logfile)
  File "/usr/bin/composer-cli", line 71, in setup_logging
    fh = logging.FileHandler(filename=logfile)
  File "/usr/lib64/python2.7/logging/__init__.py", line 902, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/lib64/python2.7/logging/__init__.py", line 925, in _open
    stream = open(self.baseFilename, self.mode)
```

In general, I think it's quite confusing that a log file is written into the current directory. I'd change the default of the `--log` parameter to not log at all.

Alternatively, there's a `JournalHandler` to use with python logging in `python-systemd`, but that would add a dependency.

bcl:
Yeah, I think composer-cli is now stable enough that it doesn't need to always log to a file.

PR #439, opened 2018-08-20
composer-cli should not log to a file by default

bcl:
The user can pass --log /path/to/logfile.log if they want logging
enabled.


bcl:
This should resolve #437

Issue #440, opened 2018-08-21
composer-cli status show reports wrong version on rhel7-extras branch


bcl:
It is reporting the lorax version, not the lorax-composer version. Need to add pylorax.api.version to this branch only so that it will report the correct version.

PR #441, opened 2018-08-21
Add a pylorax.api.version number

bcl:
lorax-composer is in a different package than lorax, so we need to use
its version number not lorax's when reporting the API /status/


bcl:
This will fix #440

bcl:
Merged.

PR #447, opened 2018-08-28
Minor package fixes for aarch64/ARMv7

nullr0ute:
The grub2-tools-efi is Mac specific, we still need uboot-tools in some corner cases.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -30,15 +30,15 @@ installpkg glibc-all-langpacks
 
 ## arch-specific packages (bootloaders etc.)
 %if basearch == "aarch64":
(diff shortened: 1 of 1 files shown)


AdamWill:
This is breaking Rawhide and F29 composes, so it's not terribly "minor" :) @bcl , can we get a quick merge, build for Rawhide and build + update for F29? Thanks!

https://pagure.io/dusty/failed-composes/issue/701

nullr0ute:
@bcl it was a copy paste error, and all verifiable, the first and second time, by checking koji or repos for the builds as it was all purely packaging changes. Apologies, had too many things up in the air and something slipped through. If I can assist in getting non x86 architectures available for use in CI pipelines please reach out.

Issue #450, opened 2018-08-30
lorax both has a copy of pyanaconda's simpleconfig...and imports it (without the package depending on it)


AdamWill:
As noted in https://github.com/weldr/lorax/issues/449 , https://github.com/weldr/lorax/commit/4d9ef60460f64b7b9f85e0c46fea1d69bf3f2e3e stuffed a copy of pyanaconda's `simpleconfig` into lorax (only it doesn't really work). However, another bit of lorax actually [imports it from pyanaconda](https://github.com/weldr/lorax/blob/b1e2bbfd17e7c9206837738230b4ee824ccff110/src/pylorax/api/compose.py#L43):

    from pyanaconda.simpleconfig import SimpleConfigFile

however, the RPM spec does not depend on anaconda-core despite using it in this way.

It seems that this codepah is not actually hit during Fedora composes, because if it was I'd expect them to blow up due to pyanaconda not being available. But there clearly seems to be an inconsistency here - it doesn't seem to make sense for one part of lorax to import simpleconfig from pyanaconda, and another part to use a local copy of it.

ferdnyc:
That line is in a file that's part of lorax-composer, which (according to [the spec file](https://github.com/weldr/lorax/blob/b1e2bbfd17e7c9206837738230b4ee824ccff110/lorax.spec)) does have anaconda-tui as a dependency.

(Lines 121-139:)
```specfile
%package composer
Summary: Lorax Image Composer API Server
# For Sphinx documentation build
BuildRequires: python3-flask python3-gobject libgit2-glib python3-pytoml python3-semantic_version

Requires: lorax = %{version}-%{release}
Requires(pre): /usr/bin/getent
Requires(pre): /usr/sbin/groupadd
Requires(pre): /usr/sbin/useradd

Requires: python3-pytoml
Requires: python3-semantic_version
Requires: libgit2
Requires: libgit2-glib
Requires: python3-flask
Requires: python3-gevent
Requires: anaconda-tui
Requires: qemu-img
Requires: tar
```

AdamWill:
aha, good catch - I was looking for `anaconda-core`.

Issue #449, opened 2018-08-30
The copy of simpleconfig that was taken from pyanaconda still needs other bits of pyanaconda


AdamWill:
https://github.com/weldr/lorax/commit/4d9ef60460f64b7b9f85e0c46fea1d69bf3f2e3e seems to have intended to take `simpleconfig` from pyanaconda and just dump it into lorax wholesale so lorax can use it without actually requiring `pyanaconda`, but...that doesn't really work, because this:

    from pyanaconda.core.util import upperASCII

so if you actually try and run current lorax in an environment without anaconda-core installed, [it just blows up](https://pagure.io/dusty/failed-composes/issue/716):

    DEBUG util.py:439:  Traceback (most recent call last):
    DEBUG util.py:439:    File "/usr/sbin/lorax", line 38, in <module>
    DEBUG util.py:439:      from pylorax.simpleconfig import SimpleConfigFile
    DEBUG util.py:439:    File "/usr/lib/python3.7/site-packages/pylorax/simpleconfig.py", line 24, in <module>
    DEBUG util.py:439:      from pyanaconda.core.util import upperASCII
    DEBUG util.py:439:  ModuleNotFoundError: No module named 'pyanaconda'

You *can* just take `upperASCII` and all the other bits of `pyanaconda.core.util` that are needed to back  it and dump them in the top of `pylorax/simpleconfig.py` to make it work, but that seems kinda...inelegant.

This is currently blocking Rawhide and F29 composes.

AdamWill:
https://stackoverflow.com/questions/2819696/parsing-properties-file-in-python/25493615#25493615 suggests a way to use python's own `configparser` (which lorax already uses in other places) to read files like this (ones that look more or less like .ini files, but have no section headers) by using a simple wrapper to add a dummy section header. That seems like a sane solution to me.

ferdnyc:
> You can just take upperASCII and all the other bits of pyanaconda.core.util that are needed to back it and dump them in the top of pylorax/simpleconfig.py to make it work, but that seems kinda...inelegant.

It did seem inelegant, but also fairly straightforward, so #452 contains a change which does just that, as an alternative solution. (It's really not that many bits, just three simple functions and the `_upperASCII_table`.)

I suspect #451 is the better solution, though.

PR #451, opened 2018-08-30
Ditch all use of pyanaconda's simpleconfig

AdamWill:
lorax uses pyanaconda's SimpleConfigParser in three different
places (twice with a copy that's been dumped into pylorax, once
by importing it), just to do a fairly simple job: read some
values out of /etc/os-release. The only value SimpleConfigParser
is adding over Python's own ConfigParser here is to read a file
with no section headers, and to unquote the values. The cost is
either a dependency on pyanaconda, or needing to copy the whole
of simpleparser plus some other utility bits from pyanaconda
into lorax. This seems like a bad trade-off.

This changes the approach: we copy one very simple utility
function from pyanaconda (`unquote`), and do some very simple
wrapping of ConfigParser to handle reading a file without any
section headers, and returning unquoted values. This way we can
read what we need out of os-release without needing a dep on
pyanaconda or to copy lots of things from it into pylorax.

Resolves: #449
Resolves: #450


AdamWill:
One thing worth noting I guess: the default 'fallback' value for this parser is `None`, not `""` as for `SimpleConfigParser`. (That's why I changed `os_release.get("PRETTY_NAME")` to `os_release.get("PRETTY_NAME", "")` and so on). I looked into changing this but it's actually quite complicated, as we aren't actually winding up with a `ConfigParser` instance here but an object that's for a *single section* within a config file so far as `ConfigParser` is concerned (a `SectionProxy`, IIRC). I can go into details if you like, but basically it's probably more trouble than it's worth to try and 'solve' that, it's easier to just remember to set `""` as the fallback value if that's what we want.

PR #452, opened 2018-08-30
Borrow from pyanaconda.core.util

ferdnyc:
This fixes simpleconfig.py by incorporating the other necessary bits previously imported from pyanaconda.core.util

Tested with the following code:
```python3
from simpleconfig import SimpleConfigFile

os_release = SimpleConfigFile("/etc/os-release")
os_release.read()
name = os_release.get("PRETTY_NAME")
print(name)
```
(Prints "Fedora 28 (Twenty Eight)" on my system.)

Fixes: #449

### DIFF
--- a/src/pylorax/simpleconfig.py
+++ b/src/pylorax/simpleconfig.py
@@ -21,7 +21,10 @@ import os
 import shlex
 import string  # pylint: disable=deprecated-module
 import tempfile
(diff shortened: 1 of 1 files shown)


bcl:
Thanks for this, #451 is a cleaner solution so that has been pushed to master.

PR #454, opened 2018-09-03
templates/generic: disable creation of default wired connections by NM

lkundrak:
NetworkManager by default connects all wired devices it sees. This is
not nice in the installer. Current Anaconda solution is to create
placeholder connections that don't connect by default, and that is sort
of ugly (and also seems broken).

### DIFF
--- a/share/templates.d/99-generic/config_files/common/93-anaconda-no-auto-default.conf
+++ b/share/templates.d/99-generic/config_files/common/93-anaconda-no-auto-default.conf
@@ -1,0 +1,3 @@
+# Disable creation of "Wired connection ..." for Wired devices
+[main]
+no-auto-default=*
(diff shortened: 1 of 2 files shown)


bcl:
Cherry picked as commit 044d7381a20ac520837c33ecf7fb56fd2e6bdf5b

rvykydal:
On Fedora the policy of auto-activating connections for installation has always been different from RHEL (from where Brian took the patch). On Fedora we don't disable the auto connections because we want to try to activate devices with dhcp for installation by default (even when networking is not explicitly configured).

This change will impact the cases when network is not required in installer stage1/initramfs - media installations (eg Server DVD, netinst). I am afraid that with the patch applied the installer will not be connected when installation starts, so
- for example geolocation (pre-selecting language) will not work for DVD and netinst
- for netinst the installation source will not be set from closest mirrot and user has to activate network first.

I think this would be regarded as a regression and we should revert the patch.

Also see my comment in related https://github.com/rhinstaller/anaconda/pull/1588.

rvykydal:
PR for revert:
https://github.com/weldr/lorax/pull/458

PR #455, opened 2018-09-04
Add prefixdevname support to the boot.iso

bcl:
Resolves: rhbz#1623000

NOTE: UNTESTED.


msekletar:
I don't think that proposed changes are exactly what we are after. There are a couple of expectations for the availability of the prefixdevname tool [1] in order to make sure that it covers all necessary use-cases and works as expected,

1) tool and corresponding udev rule is available in the anaconda environment (stage2)
2) tool is available in the installer's initrd (i.e. stage1)
3) tool is available on the installed system if ```net.ifnames.prefix``` option was used on the kernel command line during installation
4) tool is *not* installed in the initrd on the installed system
5) configuration files generated by the tool during system installation are present in the final initrd and on the installed system

1) is handled by marking the package for installation in ```runtime-install.tmpl```. 3) will be handled by anaconda. 4) is the current default because prefixdevname *doesn't have* any dracut module that causes the tool and udev rule to be installed in the initrd image. 5) is handled by the current dracut module.

Now the only missing part is 2). I think we have two options,

* introduce new dracut module (e.g. prefixdevname-lorax) which would include the tool in the initrd image used during installation
* change dracut invocation by lorax so that it copies over the tool and udev rule manually (I think we would need to use ```--include``` option)

[1] https://github.com/msekletar/prefixdevname
[2] https://github.com/msekletar/prefixdevname/blob/master/dracut/71prefixdevname/module-setup.sh

bcl:
The module name should not include '-lorax', there's nothing special about lorax adding it to the dracut driver list.
The module should handle installing what it needs, not lorax. ALL lorax should do is add it to the driver list as I've done in this patch. It looks like you already have a dracut module named prefixdevname, why isn't it enough to use that?

msekletar:
> It looks like you already have a dracut module named prefixdevname, why isn't it enough to use that?

Because that module installs configuration files generated by the tool to the initramfs image, not the tool itself. I **do not want** the tool present in initramfs image of the final system. Only exception is initrd.img used for system installation.

bcl:
> Because that module installs configuration files generated by the tool to the initramfs image, not the tool itself. I do not want the tool present in initramfs image of the final system. Only exception is initrd.img used for system installation.

Ah, ok. Maybe name the one that installs the tools 'prefixdevname-tools' or something like that. May want to talk to dracut folks and see if they have suggestions that are consistent with current modules.

msekletar:
Ohh, I just realized this is actually repo with Fedora config. prefixdevname is not packaged in Fedora and it is not decided whether it will ever be. It exists mostly to not cause too much distruption on the enterprise side for people migrating from RHEL/CentOS 6, so they can get their eth style NIC names (well except we don't allow eth prefix, but they can pick "net" instead).

msekletar:
I think we should close this for the time being. I will reopen in the future if necessary.

bcl:
Actually, this PR is against the rhel8-branch which is where most of the parts of lorax come from. But you could instead open a bz and track it there if you want.

msekletar:
I was looking at the README and it mentioned Fedora so that caused confusion on my side, let's continue here then.

Btw, here is new dracut module that I've introduced,

https://github.com/msekletar/prefixdevname/tree/master/dracut/71prefixdevname-tools

I think the second hunk should look like this,

```python
anaconda_args = dracut_args + ["--add", "anaconda pollcdrom qemu qemu-net prefixdevname-tools"]
```

Can you please update the PR?

PR #463, opened 2018-09-17
composer: Add VMDK support for VMWare images

stefwalter:
This is the basic functionality for creating VMWare images via the
Composer API. Once these images are tested, we may need to change the
kickstart file so they work appropriately.

But lets get this feature in, so that others can contribute those
tweaks effectively.


dashea:
closing in favor of https://github.com/weldr/lorax/pull/475

PR #465, opened 2018-09-17
Fix `make check` and add it to CI

larskarlitski:
See individual commits.

The container that's used is not rebuilt on every pull request anymore, but pulled from docker hub instead. I've used my own user (larskarlitski) for that for now, because I don't have access to the welder one. If we decide to go forward with this, I can change it.

- [ ] change docker image from `larskarlitski/lorax-tests` to `welder/lorax-tests`

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,12 +1,27 @@
 FROM registry.fedoraproject.org/fedora:rawhide
 
-RUN dnf -y install make libgit2-glib tito python3-pylint  \
-                    python3-nose python3-mako python3-flask \
-                    python3-coverage libselinux-python3 sudo \
-                    pykickstart python3-pytoml python3-sphinx \
-                    python3-semantic_version \
-                    anaconda-tui python3-gevent beakerlib && \
-    useradd weldr
+RUN dnf -y install \
+  anaconda-tui \
+  libgit2-glib \
+  libselinux-python3 \
+  make \
+  pykickstart \
+  python3-coverage \
+  python3-coveralls \
+  python3-flask \
+  python3-gevent \
+  python3-mako \
+  python3-nose \
+  python3-pocketlint \
+  python3-pylint \
+  python3-pytoml \
+  python3-semantic_version \
+  python3-sphinx \
+  sudo \
+  tito \
+  which
+
+RUN useradd weldr
 
 RUN mkdir /lorax
 COPY . /lorax
@@ -15,4 +30,4 @@ COPY . /lorax
 RUN find /lorax -name "*.pyc" -exec rm -f {} \;
 
 WORKDIR /lorax
-RUN make test
+RUN make check test
--- a/src/bin/composer
+++ b/src/bin/composer
@@ -23,9 +23,11 @@ log = logging.getLogger("composer")
 import os
 import sys
 
-from composer import vernum
-from composer.cli import main
-from composer.cli.cmdline import composer_cli_parser
+# Disable pylint warnings for these, because it cannot deal with this file and
+# the module both being called "composer"
+from composer import vernum                             # pylint: disable=import-self
+from composer.cli import main                           # pylint: disable=no-name-in-module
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -67,8 +67,14 @@ local:
 	@rm -rf /var/tmp/$(PKGNAME)-$(VERSION)
 	@echo "The archive is in $(PKGNAME)-$(VERSION).tar.gz"
 
+test-container:
+	sudo docker build -t larskarlitski/lorax-tests:latest -f Dockerfile.test .
+
+check-in-docker:
+	sudo docker run --rm -ti -v `pwd`:/lorax larskarlitski/lorax-tests make check
```
bcl: One of the problems with mounting this is that you end up with root owned artifacts in your home directory when running this locally. You also end up with SELinux problems unless you add `--security-opt label=disable`

Makefile
```diff
@@ -67,8 +67,14 @@ local:
 	@rm -rf /var/tmp/$(PKGNAME)-$(VERSION)
 	@echo "The archive is in $(PKGNAME)-$(VERSION).tar.gz"
 
+test-container:
+	sudo docker build -t larskarlitski/lorax-tests:latest -f Dockerfile.test .
```
bcl: You don't need to change this, it isn't pushed to dockerhub and comes from your local system.
It would also be nice to be able to have it run this target if the image is missing. I'm not sure if there's a good way to express that in the Makefile though.

Makefile
larskarlitski: Oh, good point. I've managed to make the mount read-only for `make check`, but not for `make dist`. The tests expect to be able to read and write from several directories in the source tree. Do you know if that's easily fixable?

Makefile
larskarlitski: I don't understand. One of the points of this PR is to use an image from dockerhub, so that
1. the builds on travis are faster with shorter logs
2. developers can download exactly the container that was used in testing to reproduce failures locally

larskarlitski:
Updated to make at least `make check` use a read-only source mount (thanks for pointing me to that @bcl). Also remove the coveralls invocation in `.travis.yml`, because we only want that to run after `make install`, not `make check`.

Makefile
bcl: Probably the document generation. It expects to be able to generate the docs so it can test serving them. Maybe we could drop that bit.

Makefile
bcl: Right, but if it isn't on dockerhub it will use the local one. So to test it you don't need to change the name of it, you just need to build a local image with the same name.

larskarlitski: Ah, you mean to always use the local one (without `larskarlitski/`) in the Makefile and then pull it from dockerhub in `.travis.yml` before running `make test-in-docker`?

bcl: No, what I mean is that if 'welder/lorax-tests:latest' doesn't exist in dockerhub it will look for a local image by that name. So you don't need (AFAIK) to do anything special other than push an image by that name. Basically, change it back to welder and it'll all get sorted out :)

Makefile
```diff
@@ -67,8 +67,23 @@ local:
...
+
+check-in-docker:
+	sudo docker run --rm -ti \
+          -v `pwd`:/lorax:ro \
+          -e top_builddir=/build \
+          -e top_srcdir=/lorax \
```
atodorov: `tests/testenv.sh` is supposed to configure these variables. Why do you have to pass them by hand ?

larskarlitski:
As discussed with @atodorov on irc, I'm splitting this into two parts. Update this PR to only keep the pylint fixes and add it to testing in the docker container.

larskarlitski:
Update again to disable warnings that were introduced by the `composer-cli` rename. Pylint doesn't seem to be able to differentiate between `composer` the module and `composer` the binary. I'm open for better ways of fixing this - maybe by telling pylint about the updated PYTHONPATH somehow?

PR #466, opened 2018-09-17
composer: Add VHD support for Azure images

stefwalter:
This is the basic functionality of creating VHD images via the Composer API. The underlying tooling is obviously already there in Lorax and Qemu.
    
However these images may not yet work with Azure. We need to identify other requirements. There may be requirements about size alignment for example. In addition we may need to change the kickstart file so that the images work appropriately. But lets get this feature in so that others can contribute those tweaks and configs effectively.


stefwalter:
More information of VHD Azure images here: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/create-upload-generic

dashea:
Closing in favor of https://github.com/weldr/lorax/pull/475

PR #467, opened 2018-09-17
Fix depsolve version globbing

bcl:
The blueprint version glob was being applied to the whole package NEVRA
by yum (it lacks a separate API for just globbing versions), so this
implements that in filterVersionGlob using fnmatchcase on the package
names, and the yum package verGT comparison on the versions for the
selected package.

Also includes tests.

Resolves: rhbz#1628114


atodorov:
+1 about the tests.

PR #469, opened 2018-09-18
Rename composer-cli to composer

larskarlitski:
Not sure if this is something we really want. Last time I asked, no-one had objections and @stefwalter just brought it up again. Let's continue the discussion here :)

I hope I got everything. Stuff in `html` is recreated when updating the docs, isn't it?

I haven't renamed `docs/composer-cli.rst` because `docs/composer.rst` already exists.

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -225,9 +225,9 @@ getent passwd weldr >/dev/null 2>&1 || useradd -r -g weldr -d / -s /sbin/nologin
 %attr(0771, weldr, weldr) %{_sharedstatedir}/lorax/composer/blueprints/*
 
 %files -n composer-cli
-%{_bindir}/composer-cli
+%{_bindir}/composer
 %{python3_sitelib}/composer/*
-%{_sysconfdir}/bash_completion.d/composer-cli
+%{_sysconfdir}/bash_completion.d/composer
 
 %changelog
 * Fri Sep 07 2018 Brian C. Lane <bcl@redhat.com> 29.15-1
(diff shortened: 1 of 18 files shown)


lorax.spec
```diff
@@ -224,10 +224,10 @@ getent passwd weldr >/dev/null 2>&1 || useradd -r -g weldr -d / -s /sbin/nologin
 %dir %attr(0771, root, weldr) %{_sharedstatedir}/lorax/composer/blueprints/
 %attr(0771, weldr, weldr) %{_sharedstatedir}/lorax/composer/blueprints/*
 
-%files -n composer-cli
```
stefwalter: This seems to be renaming the subpackage. I don't think that's what you want. If you do want that, you should probably open a separate pull request, as it involves different work.

lorax.spec
```diff
@@ -224,10 +224,10 @@ getent passwd weldr >/dev/null 2>&1 || useradd -r -g weldr -d / -s /sbin/nologin
...
-%{_bindir}/composer-cli
+%files -n composer
+%{_bindir}/composer
```
stefwalter: I think we need to ship a symlink to cover the old name. What do you think? I think we may need to do that on RHEL.

bcl:
I don't think we need a symlink. If we do, then why make the change at all?

stefwalter:
> I don't think we need a symlink. If we do, then why make the change at all?

I'm fine with that approach. Lets just make sure @atodorov and any folks working on Docs are aware that this change will land. @larskarlitski will you make sure of that?

larskarlitski:
> @larskarlitski will you make sure of that?

No objections from documentation. @atodorov is pinged here and I've also let him know on his latest PR.

Is anything else blocking this? Do we need more acks from the composer team?

Issue #460, opened 2018-09-12
composer-cli blueprints save will not error on non-existing blueprints


atodorov:
instead it will create an empty file and return exit code 0

```
# composer-cli blueprints save non-existing
[root@ibm-x3250m4-12 ~]# echo $?
0
[root@ibm-x3250m4-12 ~]# ls -l non-existing.toml 
-rw-r--r--. 1 root root 0 12 сеп  7,25 non-existing.toml
```

atodorov:
This has been resolved in 6da30793

PR #472, opened 2018-09-19
Report an error if the blueprint doesn't exist

bcl:
composer-cli uses TOML for 'blueprints save' which was returning an
empty 200 response if the blueprint didn't exist. Change this to return
a standard 400 error response if the blueprint doesn't exist.

composer-cli is already setup to handle receiving json when an error is
returned so just the toml API response for `blueprints/save` needed to
be changed.

### DIFF
--- a/src/pylorax/api/v0.py
+++ b/src/pylorax/api/v0.py
@@ -63,15 +63,23 @@ store the new blueprint on the new branch.
           "kubernetes" ],
         "total": 6 }
 
-`/api/v0/blueprints/info/<blueprint_names>`
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+`/api/v0/blueprints/info/<blueprint_names>[?format=<json|toml>]`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
   Return the JSON representation of the blueprint. This includes 3 top level
   objects.  `changes` which lists whether or not the workspace is different from
   the most recent commit. `blueprints` which lists the JSON representation of the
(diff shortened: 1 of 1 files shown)


bcl:
This fixes #460

src/pylorax/api/v0.py
```diff
@@ -1103,8 +1111,12 @@ def v0_blueprints_info(blueprint_names):
         blueprints = sorted(blueprints, key=lambda r: r["name"].lower())
 
         if out_fmt == "toml":
-            # With TOML output we just want to dump the raw blueprint, skipping the rest.
-            return "\n\n".join([r.toml() for r in blueprints])
+            if errors:
+                # If there are errors they need to be reported, use JSON and 400 for this
+                return jsonify(status=False, errors=errors), 400
```
atodorov: Wouldn't it be better to move the `if errors` block above the `if out_fmt == "toml"` block ? This will make the code a bit easier to read and also avoids these nested if conditions.

atodorov:
@bcl can you also uncomment the test in `tests/cli/test_blueprints_sanity.sh` which was merged just now.

src/pylorax/api/v0.py
bcl: That would break the JSON response which can return both errors and blueprints. TOML can't do that so we have to return the TOML recipes OR the JSON error response.

bcl:
> @bcl can you also uncomment the test in `tests/cli/test_blueprints_sanity.sh` which was merged just now.

This PR is against the wrong branch for that. We'll have to do that as a separate commit after that is merged here.

atodorov:
> This PR is against the wrong branch for that. We'll have to do that as a separate commit after that is merged here.

I guess we'll have to apply this fix to master (if not applied already) and also rhel7-branch ?

PR #476, opened 2018-09-24
cli: Clarify error message for unprivileged access

larskarlitski:
`os.path.exists("/run/weldr/api.socket")` returns False for users which have no
access. This leads to composer printing that the file does not exist, which is
misleading.

Since it's no possible to distinguish the two cases, fix this problem by
combining them and showing a single error message.


larskarlitski:
Rebase to master to include c97a6985fbbd02f885b377ed3dcfd77bc104a269, which fixes the tests.

PR #478, opened 2018-09-26
Provide a compatibility to dnf-3.5+

j-mracek:
Dnf in version 3 has a new config parser. In cases like
conf.tsflags.append('nodocs'), it only modify copy of tsflags values
(dnf-3.5.1), but there is no change in original object. To detect this
hidden problem DNF return a tuple that directly prevent from
modification (from dnf-3.6.1).

### DIFF
--- a/src/pylorax/api/dnfbase.py
+++ b/src/pylorax/api/dnfbase.py
@@ -56,7 +56,8 @@ def get_base_object(conf):
     dbc.reposdir = [repodir]
     dbc.install_weak_deps = False
     dbc.prepend_installroot('persistdir')
-    dbc.tsflags.append('nodocs')
+    # tsflags is append option therefor the action result in append operation into tsflags
+    dbc.tsflags = ('nodocs', )
 
     if conf.get_default("dnf", "proxy", None):
         dbc.proxy = conf.get("dnf", "proxy")
--- a/src/sbin/lorax
+++ b/src/sbin/lorax
@@ -212,7 +212,8 @@ def get_dnf_base_object(installroot, sources, mirrorlists=None, repos=None,
     conf.releasever = releasever
     conf.installroot = installroot
     conf.prepend_installroot('persistdir')
-    conf.tsflags.append('nodocs')
+    # tsflags is append option therefor the action result in append operation into tsflags
+    conf.tsflags = ('nodocs',)
 
     if proxy:
         conf.proxy = proxy


j-mracek:
Probably we have to change a documentation also. The new implementation really appends to tsflags.
Please see in following script:
```
base = dnf.base.Base()
base.read_all_repos()
base.fill_sack()
print("flags", base.conf.tsflags)
base.conf.tsflags = ['nodocs']
print("flags", base.conf.tsflags)
base.conf.tsflags = ('test')
print("flags", base.conf.tsflags)
base.conf.tsflags = base.conf.tsflags + ('test',)
print("flags", base.conf.tsflags)
base.close()
```

j-mracek:
The issue here is that conf is not anymore a python class but c++ class. We tried to emulate similar behavior of object like a python object. But we were unable to allow direct modification of conf object. If you do ```ts = base.conf.tsflags ``` it returns copy of values (dnf-3.5.1 list, dnf-3.6.1 tuple), therefore modification of that object cannot modify the original values in base.conf.tsflags. We change list to tuple to uncover a hidden problem with modification of copy of original data that results in lost of modified data. We are really so sorry but we tried our best in compatibility between Python and C++ objects exported by SWIG, but there are always limits.

src/pylorax/api/dnfbase.py
```diff
@@ -56,7 +56,8 @@ def get_base_object(conf):
     dbc.reposdir = [repodir]
     dbc.install_weak_deps = False
     dbc.prepend_installroot('persistdir')
-    dbc.tsflags.append('nodocs')
+    # tsflags is append option therefor the action result in append operation into tsflags
+    dbc.tsflags = ('nodocs', )
```
bcl: This should be dbc.tsflags += ('nodocs',) since we don't want to lose whatever else is set there.

src/sbin/lorax
```diff
@@ -212,7 +212,8 @@ def get_dnf_base_object(installroot, sources, mirrorlists=None, repos=None,
     conf.releasever = releasever
     conf.installroot = installroot
     conf.prepend_installroot('persistdir')
-    conf.tsflags.append('nodocs')
+    # tsflags is append option therefor the action result in append operation into tsflags
+    conf.tsflags = ('nodocs',)
```
bcl: Same here.

AdamWill:
@j-mracek this behaviour is just fundamentally *wrong* for a Python list or tuple, though. That's *not how they're supposed to work*. If this setting is really this weird "vector string that appends when set" thing, then that needs to be honestly represented in Python. Pretending it's a Python list or tuple, native types with expected behaviours that are completely different, is not going to fly.

I think this needs rethinking. See https://bugzilla.redhat.com/show_bug.cgi?id=1595917#c4

AdamWill:
@j-mracek there's some more on this in https://github.com/rpm-software-management/dnf/pull/1226 and https://github.com/rpm-software-management/libdnf/issues/597 ...

PR #480, opened 2018-09-27
Don't try to append to DNF config value that can't take it

AdamWill:
See https://bugzilla.redhat.com/show_bug.cgi?id=1595917 and
https://github.com/rpm-software-management/dnf/pull/1200 for
more on this. Briefly, DNF before 3.0 presented this config
value as a list...and mutating it worked. DNF from 3.0 until
3.6 presented it as a list...mutating it didn't work, but also
didn't *fail*, so this has actually not been doing anything on
DNF 3.x but we haven't noticed.

In DNF 3.6 values like this are presented as tuples instead of
lists, to try and catch usages like this, and it worked! We
need to change this one.

There is an additional weirdness here. tsflags is actually, in
libdnf terms, an OptionStringListAppend option: that means that
when something tries to *set* its value, the new value is just
appended to the existing list of values. This is very weird
behaviour when you're interacting with it like this, but
happens to be quite useful, as we can just 'set' the value to
a list like this and it will actually get appended (which is
what we want), and this one syntax happens to work correctly in
DNF 2.x, 3.0 through 3.5.1, and 3.6.

### DIFF
--- a/src/pylorax/api/dnfbase.py
+++ b/src/pylorax/api/dnfbase.py
@@ -56,7 +56,10 @@ def get_base_object(conf):
     dbc.reposdir = [repodir]
     dbc.install_weak_deps = False
     dbc.prepend_installroot('persistdir')
-    dbc.tsflags.append('nodocs')
+    # this is a weird 'AppendOption' thing that, when you set it,
+    # actually appends. Doing this adds 'nodocs' to the existing list
+    # of values, over in libdnf, it does not replace the existing values.
+    dbc.tsflags = ['nodocs']
 
     if conf.get_default("dnf", "proxy", None):
         dbc.proxy = conf.get("dnf", "proxy")
--- a/src/sbin/lorax
+++ b/src/sbin/lorax
@@ -212,7 +212,10 @@ def get_dnf_base_object(installroot, sources, mirrorlists=None, repos=None,
     conf.releasever = releasever
     conf.installroot = installroot
     conf.prepend_installroot('persistdir')
-    conf.tsflags.append('nodocs')
+    # this is a weird 'AppendOption' thing that, when you set it,
+    # actually appends. Doing this adds 'nodocs' to the existing list
+    # of values, over in libdnf, it does not replace the existing values.
+    conf.tsflags = ['nodocs']
 
     if proxy:
         conf.proxy = proxy


bcl:
For those reading along at home, it looks like dnf is going to fix this on their end so I'm holding off pushing this unless it's *really* necessary.

atodorov:
> For those reading along at home, it looks like dnf is going to fix this on their end so I'm holding off pushing this unless it's really necessary.

@bcl do we have any ETA for this fix? As of now the dnf change is breaking the tests so we're kind of blocked doing anything on the master branch, aren't we ?

src/pylorax/api/dnfbase.py
```diff
@@ -56,7 +56,10 @@ def get_base_object(conf):
     dbc.reposdir = [repodir]
     dbc.install_weak_deps = False
     dbc.prepend_installroot('persistdir')
-    dbc.tsflags.append('nodocs')
+    # this is a weird 'AppendOption' thing that, when you set it,
+    # actually appends. Doing this adds 'nodocs' to the existing list
+    # of values, over in libdnf, it does not replace the existing values.
+    dbc.tsflags = ['nodocs']
```
atodorov: For the record: internally changing the type of something coming from a 3rd party looks like a bad idea. I know it works b/c both are iterators but imagine going to dnf sources/docs where this is a tuple and then seeing it used as a list.

Not sure if it can lead to more severe issues in the future but definitely looks odd.

AdamWill: It's not actually anything like that simple. It's not as simple as saying any of these things "are" lists or tuples, and it doesn't exactly work "because both are iterators", it works [because dnf is explicitly set up to accept them being set as either lists or tuples](https://github.com/rpm-software-management/dnf/blob/master/dnf/conf/config.py#L242). Note that's in the middle of a rather complicated `__getattr__` which, yes, in DNF 3.6 ultimately returns a tuple (but in all previous DNFs, you got a list).

We could set it as a tuple, sure, but it's not at all obviously "better". I went with list because it's more compatible with ancient yum/dnf which would not accept setting these as tuples, though that's probably not a major concern.

bcl:
I have pushed this to master in the interest of making rawhide actually be useful for development until dnf fixes this.

atodorov:
Closing based on the previous comment, this is indeed in master.

PR #486, opened 2018-10-02
Write a rootpw line if no root customizations in the blueprint

bcl:
Anaconda requires the root password to be set or locked, so if there
isn't anything setting it we write out 'rootpw --lock'

Also adds tests for this.

Resolves: rhbz#1626122

### DIFF
--- a/src/pylorax/api/compose.py
+++ b/src/pylorax/api/compose.py
@@ -120,6 +120,8 @@ def write_ks_root(f, user):
     :type f: open file object
     :param user: A blueprint user dictionary
     :type user: dict
+    :returns: True if it wrote a rootpw command to the kickstart
+    :rtype: bool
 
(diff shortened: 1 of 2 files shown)


src/pylorax/api/compose.py
```diff
@@ -135,9 +139,13 @@ def write_ks_root(f, user):
         if any(user["password"].startswith(prefix) for prefix in ["$2b$", "$6$", "$5$"]):
             log.debug("Detected pre-crypted password")
             f.write('rootpw --iscrypted "%s"\n' % user["password"])
+            wrote_rootpw = False
```
dashea: shouldn't this (and the other one) be `wrote_rootpw = True` ?

bcl: Doh.

PR #488, opened 2018-10-03
Rename composer-cli to composer

bcl:
Resolves: rhbz#1635760


dashea:
The paths for the binary name and the bash_completion file need to be changed in %files

bcl:
> The paths for the binary name and the bash_completion file need to be changed in %files

fixed, thanks.

PR #490, opened 2018-10-04
Make no-virt generated images sparser

dashea:
At the end of disk image installs, use fstrim on the generated filesystem to
discard any blocks that were allocated during the install and are now unused.
This will allow tools such as qemu-img to create images that do not include
deleted data.

For raw disk images that do not go through qemu-img, use fallocate --dig-holes
to create sparse holes in place of the unused blocks.


dashea:
virt installs already do something like this, or at least close enough.

Here's the results of `du -s` on some before and after images:

|                |raw    |vhd    |vmdk   |
|----------------|-------|-------|-------|
|original        |1565432|1573916|1573812|
|fstrim          |1292732|1254356|1328492|
|fstrim+fallocate|1262956| N/A   | N/A   |

PR #493, opened 2018-10-05
Add and enable cloud-init for ami and vhd

### LINKED ISSUE #492: Unable to authenticate on EC2 to images that I've built with lorax
So I built an AMI file, uploaded it S3 and launched it on EC2.
Judging from the screenshot on EC2, it boots successfully, but I'm unable to connect to it using the PEM key that EC2 instructs me to use under the Connect menu. When I connect to the instance using SSH, it falls back to user/pass authentication (but that won't work either, because there is not user/pass on the image)

larskarlitski:
Images don't work at all on these clouds without cloud-init.

Fixes #492


dashea:
Sounds good for AMI. For VHD, though, Microsoft explicitly says not to use cloud-init (https://docs.microsoft.com/en-us/azure/virtual-machines/linux/redhat-create-upload-vhd#prepare-a-red-hat-based-virtual-machine-from-kvm). Is there something else we need to do with waagent?

larskarlitski:
> For VHD, though, Microsoft explicitly says not to use cloud-init 

Ah, good point. Let's do that in a separate patch after @andreasn tried it. Thanks.

Updated to remove the VHD part.

PR #495, opened 2018-10-05
Kickstart tweaks

dashea:
Add back one of the useful parts of fedora-kickstart's %post, and make cloud-init mandatory for a couple more image types.


dashea:
Removed the qcow2 cloud-init change, since https://github.com/weldr/lorax/pull/496 takes care of that

PR #497, opened 2018-10-05
Add compose localimage command

bcl:


bcl:
> This makes the API non-remotable and effectively turns these internals to public API:

It actually does the opposite, it gives you a way to get to the image file without needing to know internals. The alternative is to explore the filesystem under /var/lib/lorax/composer which has been suggested in some posts.

The API server exposes it's local path so that the composer-cli doesn't have to import anything from pylorax.api which would be bad :)

> 
>     1. images have specific names, which depend on the image type

Actually, that's a problem with this PR. I realized that I'm only returning the directory, not the full path to the output image. It'd probably be better to change all this so that there is a new API endpoint that just returns the local path to the image instead of trying to construct it in the composer-cli


> I assume this is to avoid copying images over the socket?

Right, there have been some posts suggesting using the /var/lib/ path to the results, so I figured it might be nice to make an easier way to do that than exploring the filesystem.

> 
> More specifically to these patches:
> 
> `localimage` sounds like a counterpart to `image`. I'd expect it to return the full path the the image.

Agreed.

> 
> composer should throw an error for non-existing images:
> 
> ```
> $ composer compose localimage foo
> /var/lib/lorax/composer/results/foo
> ```

Correct. Thanks.

larskarlitski:
> It actually does the opposite, it gives you a way to get to the image file without needing to know internals.

The results directory wasn't really API before, but you're right, people documenting it as such effectively made it so. `localimage` returning the full path to the image (which seems to have been your intention) is a good pragmatic solution to that.

The only thing I'd add is a caveat in the documentation: "returns the path to the local image, if it exists". In case we ever want to make this API work form images that are not available on a local file system.

PR #498, opened 2018-10-06
Reproducible image building

marmarek:
As part of [Reproducible Builds effort](https://reproducible-builds.org/), make the images produced by lorax reproducible, given the same set of inputs (packages, configuration, [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) variable).
See individual commits for explanation of specific changes.

The last commit require matching anaconda change, as it change image layout. It should be possible to make anaconda support both old and new format. Is there any actual reason why ext4 image is packaged into squashfs instead of using squashfs directly? I imagine historically it could be lack of overlayfs in vanilla kernel and the need to use dm-snapshot. But it is no longer the case. Anything else?

One thing not solved here is `efiboot.img`, because I didn't managed to reproducibly build FAT filesystem. Even after eliminating obvious metadata differences (volume ID, file mtimes, files order), there are still some differences near the end of the image, which I didn't identified.

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -18,6 +18,7 @@ RUN dnf -y install \
   python3-semantic_version \
   python3-sphinx \
   python3-rpmfluff \
+  python3-librepo \
   beakerlib \
   sudo \
   tito \
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -362,3 +362,19 @@ removepkg cdparanoia-libs opus libtheora libvisual flac-libs gsm avahi-glib avah
 
 ## metacity requires libvorbis and libvorbisfile, but enc/dec are no longer needed
 removefrom libvorbis --allbut /usr/${libdir}/libvorbisfile.* /usr/${libdir}/libvorbis.*
+
+## make the image more reproducible
+
+## make machine-id empty but present to avoid systemd populating /etc with
+## preset settings
+remove /etc/machine-id
+append /etc/machine-id ""
+## journalctl message catalog, non-deterministic
+remove /var/lib/systemd/catalog/database
+## non-reproducible caches
+remove /var/cache/ldconfig/aux-cache
+remove /etc/pki/ca-trust/extracted/java/cacerts
+
+## sort groups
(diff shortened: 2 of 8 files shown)


bcl:
Thanks for posting these, I'll try to take a look at them sometime this week.

bcl:
Overall this looks pretty simple. ISTR there was a good reason we didn't switch to plain squashfs for the install.img but cannot remember exactly why. @wgwoods may remember though.
Changing the on-disk layout needs to be coordinated with Fedora and especially Anaconda so you may want to join the fedora-devel or anaconda-devel mailing lists and introduce yourself there.
I'm not convinced that reproducible builds are worth the extra effort, in general, compared with good gpg signatures on everything, but these patches are pretty clean so I'm not opposed to merging them in the future if you can get the needed changes into Anaconda.

marmarek:
Apart from making the build reproducible, this also reduce size and complexity of the image. Not requiring `dmsquash-live` dracut module (which include dmsetup etc)  make initramfs noticeably smaller. Normally it isn't a big deal, but if you need to put that + kernel into efiboot.img, it's important. Because boot image on ISO9660 is limited to 32MB.
I'll write on fedora-devel about this.
If you want, I can extract this one commit to a separate PR.

marmarek:
Asked here: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/BJF3SLPPIXJER7FHK3HXYBL7TLK5T2AF/

bcl:
> If you want, I can extract this one commit to a separate PR.

Yeah, I think the others can be taken as-is so that would be good.

wgwoods:
> Is there any actual reason why ext4 image is packaged into squashfs instead of using squashfs directly? I imagine historically it could be lack of overlayfs in vanilla kernel and the need to use dm-snapshot. But it is no longer the case. Anything else?

Correct, the problem with bare squashfs images was that anaconda still needs its root filesystem to be writeable (for `updates.img` mainly), and squashfs doesn't support the write() system call at all, so there's no way to make a squashfs filesystem writeable using device-mapper overlays. And overlayfs wasn't in the mainstream kernel at the time (Fedora 15-16 / kernel 2.6-3.1) - that only became an option in Fedora 22 / kernel 4.0.

I've always thought the ext4-inside-squashfs image payload was needlessly complex and would be happy to see a simpler solution.

marmarek:
> If you want, I can extract this one commit to a separate PR.

Done: #507

PR #505, opened 2018-10-12
Fix vhd images

poncovka:
Images don't work without these fixes:

* Enable Network Manager.
* Disable cloud-init.
* Add Hyper-V modules into initramfs.

Fixes specific for RHEL:

* Create ifcfg-eth0 required by waagent.
* Install python3 and net-tools required by waagent.

Recommended changes:

* Use recommended kernel boot args.
* Disable kdump.

### DIFF
--- a/share/composer/vhd.ks
+++ b/share/composer/vhd.ks
@@ -22,10 +22,10 @@ shutdown
 # System timezone
 timezone  US/Eastern
 # System bootloader configuration
-bootloader --location=mbr --append="no_timer_check console=ttyS0,115200n8 console=tty1 net.ifnames=0"
+bootloader --location=mbr --append="no_timer_check console=ttyS0,115200n8 earlyprintk=ttyS0,115200 rootdelay=300 net.ifnames=0"
 
 # Basic services
-services --enabled=sshd,chronyd,waagent,cloud-init
+services --enabled=sshd,chronyd,waagent
 
 %post
 # Remove random-seed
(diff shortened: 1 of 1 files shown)


share/composer/vhd.ks
```diff
@@ -43,10 +64,9 @@ kernel
 grub2
 
 chrony
-cloud-init
 
-# Uninstall NetworkManager, install WALinuxAgent
--NetworkManager
 WALinuxAgent
+python3
```
stefwalter: Lets add a comment here above these two lines to say these are required by WALinuxAgent

share/composer/vhd.ks
```diff
@@ -34,6 +34,27 @@ rm /var/lib/systemd/random-seed
...
+PEERDNS=yes
+IPV6INIT=no
+EOF
+
+# Add Hyper-V modules into initramfs
+cat > /etc/dracut.conf.d/10-hyperv.conf << EOF
+add_drivers+=" hv_vmbus hv_netvsc hv_storvsc "
+EOF
+
+dracut -f -v --persistent-policy by-uuid
```
stefwalter: Do you know why we're doing this? Can we add a comment?

share/composer/vhd.ks
```diff
@@ -34,6 +34,27 @@ rm /var/lib/systemd/random-seed
 # Clear /etc/machine-id
 rm /etc/machine-id
 touch /etc/machine-id
+
+# Set up network.
```
stefwalter: Is this NetworkManager compatible? I assume so. If so, can you say that in the comment above, for the next person who comes around wondering this?

share/composer/vhd.ks
poncovka: Fixed.

share/composer/vhd.ks
poncovka: Fixed.

share/composer/vhd.ks
poncovka: Fixed.

share/composer/vhd.ks
bcl: If WALinuxAgent requires it, why isn't it actually requiring it?

share/composer/vhd.ks
bcl: We really shouldn't run dracut from post. Ideally these extra drivers would be added by a package, or maybe (ugh) a kickstart dracut command. I guess it's fine for the moment, but we should find a better way to do this.

stefwalter:
I've validated that this works.

bcl:
Thanks, I'm going to merge this and we can work on better solutions in a new PR.

PR #506, opened 2018-10-12
Update depsolving with suggestions from dnf (#1636239)

bcl:
The previous method worked, but wasn't exactly ideomatic. This is more
correct, and appears to work the same (templates depsolve, version globs
work, multiple repos work).

Note that this does use a private dnf attribute ._goal, but the word is
that this is going to become a public api soon, so yes it is there on
purpose.

(cherry picked from commit 38504254818f93f7e6c544d52e0e1a1f661ea9c9)

Resolves: rhbz#1638683


dashea:
> ideomatic

idiomatic

PR #507, opened 2018-10-12
Drop inner rootfs.img layer

marmarek:
Make runtime directly into squashfs image. This reduces largely
unreproducible ext4 layer, but requires anaconda's dracut module
modification to properly mount the image.

Split from #498 
Discussion: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/BJF3SLPPIXJER7FHK3HXYBL7TLK5T2AF/


bcl:
Added in #509 with a switch.

PR #509, opened 2018-10-19
Add --squashfs-only option to drop inner rootfs.img layer

bcl:
Make runtime directly into squashfs image. This reduces largely
unreproducible ext4 layer, but requires anaconda's dracut module
modification to properly mount the image.


bcl:
Test failure here is fixed by #508

PR #512, opened 2018-10-24
Create manpages for composer-cli and lorax-composer

bcl:


bcl:
Test failure here is because we need to bump the tmux version for rawhide.

PR #513, opened 2018-10-25
New test: build and deploy images on AWS

atodorov:
Add new script to automate the test process for AWS.  Takes around 25 minutes to execute.

Currently nothing executes this automatically and that is something we need to figure out how to do.

Note: if you want to start this manually either start it from the source code directory or modify the CLI variable to /usr/bin/composer-cli if you want to use the system-wide executable.

CC @stefwalter

### DIFF
--- a/tests/cli/test_build_and_deploy_aws.sh
+++ b/tests/cli/test_build_and_deploy_aws.sh
@@ -1,0 +1,197 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+#####
+#
+# Make sure we can build an image and deploy it inside AWS!
+#
+#####
+
+. /usr/share/beakerlib/beakerlib.sh
+
+CLI="./src/bin/composer-cli"
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        if [ -z "$AWS_ACCESS_KEY_ID" ]; then
+            rlFail "AWS_ACCESS_KEY_ID is empty!"
+        else
+            rlLogInfo "AWS_ACCESS_KEY_ID is configured"
+        fi
+
+        if [ -z "$AWS_SECRET_ACCESS_KEY" ]; then
+            rlFail "AWS_SECRET_ACCESS_KEY is empty!"
+        else
+            rlLogInfo "AWS_SECRET_ACCESS_KEY is configured"
+        fi
+
+        AWS_BUCKET="${AWS_BUCKET:-composerredhat}"
+        AWS_REGION="${AWS_REGION:-us-east-1}"
+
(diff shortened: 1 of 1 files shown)


tests/cli/test_build_and_deploy_aws.sh
```diff
@@ -0,0 +1,187 @@
...
+            rlRun -t -c "dnf -y install python3-pip"
+            rlAssertRpm python3-pip
+        fi
+
+        rlRun -t -c "pip3 install awscli"
+
+        # aws configure
+        mkdir ~/.aws/
+
+        cat > ~/.aws/config << __EOF__
```
bcl: I think we need to make sure that tests can either be run as a user, without destroying the user's environment, or add protections to stop the test if it isn't running inside a test environment like docker, podman, etc. If I accidentally ran this locally I'd be fairly well annoyed that my personal aws credentials had been wiped.
(And this may not be the only case of this happening, need to look at the rest of the tests)

atodorov: Updated to not overwrite the files if they exist

tests/cli/test_build_and_deploy_aws.sh
```diff
@@ -0,0 +1,198 @@
...
+        fi
+
+        rlRun -t -c "pip3 install awscli"
+
+        # aws configure
+        [ -d ~/.aws/ ] || mkdir ~/.aws/
+
+        if [ -f ~/.aws/config ]; then
+            rlLogInfo "Reusing existing ~/.aws/config"
+            rlRun -t -c "cat ~/.aws/config"
```
bcl: Does the cat really need to be there? Or was that just for debugging?

atodorov: More for debugging and easy access to information from the test log. However if you think we can expose information in this way I can remove it. My understanding is that `.aws/config` doesn't contain anything like access credentials and passwords.

bcl: I think it's better to remove it, you can put secrets in there and it'd be pretty bad for them to end up in logs.

stefwalter:
@atodorov Is this a good candidate for integrating into the CI here in GitHub? If so, is that something you'd like to do? Or should we figure out who could do the work?

atodorov:
> @atodorov Is this a good candidate for integrating into the CI here in GitHub? If so, is that something you'd like to do? Or should we figure out who could do the work?

I don't think this can be integrated with Travis CI because of the time it takes to execute and the fact that building images inside a Docker container kind of doesn't work. It is on my todo list to work on integrating these test scripts into the development workflow sometime in the future but not this week.

PR #516, opened 2018-10-31
Some more tests

bcl:
And a better way to run them with docker/podman


atodorov:
It looks like we've lost coverage data with this PR, otherwise looks good.

bcl:
> It looks like we've lost coverage data with this PR, otherwise looks good.

Yeah, I don't understand what's going on there. It's running coverage but not picking it up for some reason.

PR #517, opened 2018-11-02
WIP: Bring back upgrade initrd images

vinzenz:
While fedup isn't supporting upgrade images anymore a new approach is going to require it for more generic situations. Therefore we'd like to have it back. I mostly attempted to revert the changes made by Will Woods originally while trying to adjust it to the parts that have been changed.

rhel8_branch label suggested

### DIFF
--- a/share/templates.d/99-generic/aarch64.tmpl
+++ b/share/templates.d/99-generic/aarch64.tmpl
@@ -29,6 +29,9 @@ mkdir ${KERNELDIR}
     ## normal aarch64
     installkernel images-${basearch} ${kernel.path} ${KERNELDIR}/vmlinuz
     installinitrd images-${basearch} ${kernel.initrd.path} ${KERNELDIR}/initrd.img
+    %if doupgrade:
+        installupgradeinitrd images-${basearch} ${kernel.upgrade.path} ${KERNELDIR}/upgrade.img
+    %endif
 %endfor
 
 #FIXME: this will need adjusted when we have a real bootloader.
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -369,6 +369,21 @@ class Lorax(BaseLoraxClass):
         logger.info("anaconda args = %s", anaconda_args)
         treebuilder.rebuild_initrds(add_args=anaconda_args)
 
+        if doupgrade:
+            # Build upgrade.img. It'd be nice if these could coexist in the same
+            # image, but that would increase the size of the anaconda initramfs,
+            # which worries some people (esp. PPC tftpboot). So they're separate.
+            try:
+                # If possible, use the 'redhat-upgrade-tool' plymouth theme
+                themes = runcmd_output(['plymouth-set-default-theme', '--list'],
+                                       root=installroot)
+                if 'leapp-upgrade-tool' in themes.splitlines():
+                    os.environ['PLYMOUTH_THEME_NAME'] = 'leapp-upgrade-tool'
+            except RuntimeError:
+                pass
+            upgrade_args = dracut_args + ["--add", "system-upgrade"]
+            treebuilder.rebuild_initrds(add_args=upgrade_args, prefix="upgrade")
+
         logger.info("populating output tree and building boot images")
         treebuilder.build()
 
--- a/share/templates.d/99-generic/arm.tmpl
+++ b/share/templates.d/99-generic/arm.tmpl
@@ -32,9 +32,18 @@ mkdir ${KERNELDIR}
         platforms = platforms + delimiter + kernel.flavor
         delimiter = ','
 %>
+        %if doupgrade:
+            ## install upgrade image
+            installupgradeinitrd images-${kernel.flavor}-${basearch} ${kernel.upgrade.path} ${KERNELDIR}/upgrade-${kernel.flavor}.img
+        %endif
     %else:
         installkernel images-${basearch} ${kernel.path} ${KERNELDIR}/vmlinuz
         installinitrd images-${basearch} ${kernel.initrd.path} ${KERNELDIR}/initrd.img
+
+        %if doupgrade:
+            ## install upgrade image
+            installupgradeinitrd images-${basearch} ${kernel.upgrade.path} ${KERNELDIR}/upgrade.img
+        %endif
     %endif
 %endfor
 
--- a/share/templates.d/99-generic/s390.tmpl
+++ b/share/templates.d/99-generic/s390.tmpl
@@ -29,6 +29,11 @@ replace @INITRD_LOAD_ADDRESS@ ${INITRD_ADDRESS} generic.ins
 installkernel images-${basearch} ${kernel.path} ${KERNELDIR}/kernel.img
 installinitrd images-${basearch} ${kernel.initrd.path} ${KERNELDIR}/initrd.img
 
+%if doupgrade:
+    ## upgrader image
+    installupgradeinitrd images-${basearch} ${kernel.upgrade.path} ${KERNELDIR}/upgrade.img
+%endif
+
 ## s390 needs some extra boot config
 createaddrsize ${INITRD_ADDRESS} ${outroot}/${BOOTDIR}/initrd.img ${outroot}/${BOOTDIR}/initrd.addrsize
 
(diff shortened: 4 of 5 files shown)


src/pylorax/__init__.py
```diff
@@ -369,6 +369,21 @@ def run(self, dbo, product, version, release, variant="", bugurl="",
         logger.info("anaconda args = %s", anaconda_args)
         treebuilder.rebuild_initrds(add_args=anaconda_args)
 
+        if doupgrade:
+            # Build upgrade.img. It'd be nice if these could coexist in the same
+            # image, but that would increase the size of the anaconda initramfs,
+            # which worries some people (esp. PPC tftpboot). So they're separate.
+            try:
+                # If possible, use the 'redhat-upgrade-tool' plymouth theme
+                themes = runcmd_output(['plymouth-set-default-theme', '--list'],
```
bcl: That's not going to do what you want it to do. runcmd runs in the context of the host system doing the build, not the packages being installed. You need to find a better way to accomplish this.

share/templates.d/99-generic/aarch64.tmpl
```diff
@@ -29,6 +29,9 @@ mkdir ${KERNELDIR}
     ## normal aarch64
     installkernel images-${basearch} ${kernel.path} ${KERNELDIR}/vmlinuz
     installinitrd images-${basearch} ${kernel.initrd.path} ${KERNELDIR}/initrd.img
+    %if doupgrade:
```
bcl: doupgrade needs to be added to the <%page section's list of variables. It also needs to be passed into the template parser. eg. https://github.com/weldr/lorax/blob/27e611629f2c654939b0894086512e9d4f366113/src/pylorax/treebuilder.py#L258

src/pylorax/__init__.py
vinzenz: Actually I don’t even know what this does, this is restored code from the rhel7 branch
I will have to give this a try anyway and need to get also the system upgrade dracut module back.

I should maybe haven’t already created the PR it was probably too early

wgwoods:
I'd like to take a second to review the message from the commit (a3e1a27) where I removed `upgrade.img`:
 
> For the record, performing upgrades using an initramfs from the new
> system turns out to be fragile and hard to support:
> 
> * dracut initramfs isn't generic enough to handle booting all systems
>   (e.g. missing vconsole.conf means you get keymaps wrong, so users
>   can't unlock encrypted disks)
> 
> * The ABI differences between the two versions of plymouth, systemd,
>   etc. requires nasty workarounds at best and causes nightmarish
>   systemd crashes at worst

So.. have these problems been addressed since then?

More specifically, I'd like some answers to the following questions before _I'd_ feel comfortable supporting `upgrade.img`: 

1. The systemd developers explicitly rejected supporting switch-roots from a newer version of systemd to an older one - which is the main cause of the nightmarish crashes mentioned above, since that's what happens when `upgrade.img` tries to switch into the existing system to mount its disks. Have they agreed to support this for the (I'm assuming) RHEL7 -> RHEL8 use case? What about plymouth ABI differences?

2. How do you plan to support upgrades on systems that require third-party disk drivers that aren't present in `upgrade.img`, or have unusual/custom disk setups that `upgrade.img` might not handle correctly? Who will be responsible for tracing and fixing those problems when customers encounter them?

3. What's the plan for handling SELinux policy during the upgrade? Does the upgrade run with the old policy in effect or does it use the new system's policy? If the latter, what's the plan for getting the system to switch policy? Who's responsible for supporting that code and handling SELinux-related bugs?

4. What's the plan for getting the correct/required config files (e.g. `vconsole.conf`) from the old system into `upgrade.img` so it can boot correctly? What happens if it turns out we're missing a required file - will that require a new build of lorax and new `upgrade.img`?

vinzenz:
> I'd like to take a second to review the message from the commit ([a3e1a27](https://github.com/weldr/lorax/commit/a3e1a275ff838fccb4fa7e7ac040b773dfdd799b)) where I removed `upgrade.img`:
> 
> > For the record, performing upgrades using an initramfs from the new
> > system turns out to be fragile and hard to support:
> > 
> > * dracut initramfs isn't generic enough to handle booting all systems
> >   (e.g. missing vconsole.conf means you get keymaps wrong, so users
> >   can't unlock encrypted disks)
> > * The ABI differences between the two versions of plymouth, systemd,
> >   etc. requires nasty workarounds at best and causes nightmarish
> >   systemd crashes at worst
> 
> So.. have these problems been addressed since then?
> 
> More specifically, I'd like some answers to the following questions before _I'd_ feel comfortable supporting `upgrade.img`:
> 
> 1. The systemd developers explicitly rejected supporting switch-roots from a newer version of systemd to an older one - which is the main cause of the nightmarish crashes mentioned above, since that's what happens when `upgrade.img` tries to switch into the existing system to mount its disks. Have they agreed to support this for the (I'm assuming) RHEL7 -> RHEL8 use case? What about plymouth ABI differences?
We are going with a RHEL8 kernel and RHEL8 systemd without doing switch root and are going to perform the 'switch' by using systemd-nspawn and perform the upgrade in a  'change root' where the upgrade transaction happens. The old system won't start at all. 
So this shouldn't have any impact on any of those previous problems.

> 2. How do you plan to support upgrades on systems that require third-party disk drivers that aren't present in `upgrade.img`, or have unusual/custom disk setups that `upgrade.img` might not handle correctly? Who will be responsible for tracing and fixing those problems when customers encounter them?
upgrade.img is supposed to be used for generic cases where no disk encryption is in place and no 3rd party drivers are in place. This is meant for most generic workloads to speed up the process.
In the first phase 3rd party drivers etc aren't supported and would be supported by a custom built
initram disk on the system before we reboot, by using a containerized build environment that would
take care of generating the initram disk with drivers compiled for the new kernel. However that requires that we can build them ofc, but that's a different topic in total.

> 3. What's the plan for handling SELinux policy during the upgrade? Does the upgrade run with the old policy in effect or does it use the new systems's policy? If the latter, what's the plan for getting the system to switch policy? Who's responsible for supporting that code and handling SELinux-related bugs?
SELinux is going to be set into permissive mode during the initram upgrade phase. We are in touch with the SELinux developers for checking the source system policies for potential problems before performing the upgrade and apply automatic conversions or removals for invalid/outdated policies when applicable. The relabeling should happend after the upgrade transaction has been applied, at this point it is just scheduled by touching /.autorelabel on the system before we reboot which will cause the relabelling to be executed and another reboot will happen afterwards. I think that part is well known and doesn't need any further explanation.

> 4. What's the plan for getting the correct/required config files (e.g. `vconsole.conf`) from the old system into `upgrade.img` so it can boot correctly? What happens if it turns out we're missing a required file - will that require a new build of lorax and new `upgrade.img`?
That part would be then the use case for a custom initram disk which is planned for a later release however that shouldn't be affecting lorax except we did something wrong there, but those custom things should be taken care of in a different manner by creating a specialized initram disk. 


I would like to point out here that they original dracut hooks that have been used in the 6 => 7 scenario aren't applicable here anymore, we are changing the way how it works and we don't do the double switch root for example and other things. The dracut hooks need still some polishing so I realized that after the pull request that we have to make things available first.

And another reason why we really can't do this system-update approach that is used by the dnf system-upgrade plugin is because of the problems you mentioned above since we require the new kernel to be present with the new systemd and we can't allow it to make a switch root to a 20 versions older systemd that is not supported and would be a problem not to speak of the mentioned ABI problems of any kind of application that is involved.

Another thing, please beware this should go to the rhel8 branch only, for Fedora I wouldn't want to change this as of now as the system-update approach is still sufficient.

share/templates.d/99-generic/aarch64.tmpl
bcl: Err, oops. I didn't see it in your commit so I assumed it also needed to be restored. Apparently it's already there so nevermind :)

src/pylorax/__init__.py
bcl: That's probably a good idea. Hack up your local copy of lorax and make sure things are all working. Add tests if you can, then do a new PR. We also need to make sure that whatever new dracut modules are needed are already available before we merge this.

vinzenz:
Closing for now - until the requested things are resolved

PR #518, opened 2018-11-06
New test: build and deploy images on vSphere

atodorov:
Uses our own fork of vmware/pyvmomi-community-samples, which contains scripts and patches to do what we need. 

Then follows the same steps as the script for AWS images:
- builds an image
- uploads it to vCenter
- creates a VM from that image
- verifies the VM
- clean up

### DIFF
--- a/tests/cli/test_build_and_deploy_vmware.sh
+++ b/tests/cli/test_build_and_deploy_vmware.sh
@@ -1,0 +1,156 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+#####
+#
+# Make sure we can build an image and deploy it inside vSphere!
+#
+#####
+
+. /usr/share/beakerlib/beakerlib.sh
+
+CLI="./src/bin/composer-cli"
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        if [ -z "$V_HOST" ]; then
+            rlFail "V_HOST is empty!"
+        else
+            rlLogInfo "V_HOST=$V_HOST"
+        fi
+
+        if [ -z "$V_USERNAME" ]; then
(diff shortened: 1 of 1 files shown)


tests/cli/test_build_and_deploy_vmware.sh
```diff
@@ -0,0 +1,155 @@
...
+        SAMPLES="./pyvmomi-community-samples/samples/"
+        if [ ! -d ./pyvmomi-community-samples/ ]; then
+            rlRun -t -c "git clone https://github.com/weldr/pyvmomi-community-samples"
+            pushd ./pyvmomi-community-samples/ && git checkout composer_testing && popd
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose start"
+        if [ ! -f ~/.ssh/id_rsa.pub ]; then
+            rlRun -t -c "ssh-keygen -t rsa -N '' -f ~/.ssh/id_rsa"
```
bcl: If you run this on a local checkout it's going to clobber your ssh key! Better to try and put the artifacts into a /tmp/ directory and pull them from there (the same goes for vmware.toml and the git clone.

atodorov: How come? Isn't this why there's an if-file-doesn't-exist statement above ?

I agree putting the rest of the artifacts in temporary directory but for SSH keys would like to reuse existing keys if possible.

bcl: > How come? Isn't this why there's an if-file-doesn't-exist statement above ?

Sigh. How did I miss that? Sorry for the noise, looks good to me.

PR #519, opened 2018-11-07
Update docs with info about ssh keys

atodorov:
it looks like the key value expects the entire content of ~/.ssh/id_rsa.pub

Note: I haven't updated the resulting HTML files. Do I need to do it?


atodorov:
CC @stefwalter

PR #521, opened 2018-11-09
For OpenStack build image with rng-tools installed

atodorov:
appears to be a decent workaround for slow boot likely caused
by lack of entropy for ssh-keygen, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1639326


atodorov:
@bcl I have created a JIRA task for your comment. Will merge as-is and then try to work on improving the integration with the host if at all possible.

PR #523, opened 2018-11-13
Remove checks for SELinux permissive

bcl:
Also removed unnecessary sefiles call used by lorax.


bachradsusi:
I think that setfiles could be necessary on build systems where SELinux is disabled or when a new image would use a different policy than the system where it's built on. But I need to confirm that.

andreasn:
I still get the message:
`2018-11-14 14:29:03,311: selinux is enabled and in Enforcing mode`
Is that intended? If it doesn't affect the service any more, maybe it can be dropped?

andreasn:
Set SELinux to Enabled, started lorax-composer and built an image without any issues. Thanks for fixing this!

bcl:
> I think that setfiles could be necessary on build systems where SELinux is disabled
I can give running in disabled a try, I'm not sure if we've ever said that we support that though. If you're making a selinux enabled image you should at least use Permissive.

I'm not sure how to try situations where the policy might be different. Or how to handle that if it is an issue.

> 2018-11-14 14:29:03,311: selinux is enabled and in Enforcing mode
That's expected, I added it so that we know for sure what mode the system is running in.

Also, this isn't perfect, under RHEL8 a tar build is failing because of the useradd/groupadd denial. Works fine with qcow2 though. The difference is with tar it is running on /mnt/sysimage/ directly and with qcow2 there is a disk partition mounted there. Any suggestions?

bcl:
And in Fedora 29 I see the useradd/groupadd denials, but the tar install finished without failing. Maybe a difference in anaconda or useradd?

bachradsusi:
@bcl I'm looking for answers to your questions but I need some time for that. I'll answer as soon as I know more.

bcl:
> I'm looking for answers to your questions but I need some time for that. I'll answer as soon as I know more.

Thanks. I retried f29/rhel8 today to try and nail down the differences and it ends up BOTH fail on tar when you add a new user. If there are no [[customizations.user]] entries then it works ok. RHEL 7.6 works because it uses libuser instead of useradd/groupadd like the newer versions of anaconda.

bachradsusi:
I think I've found the problem, it can be reproduced by the following command run during compose:

```
# useradd -R /mnt/sysimage -g 100 -G users,wheel -d /home/widgeta/ -m -s /usr/bin/bash -u 1202 -c "Administrator account" adminal
useradd: cannot set SELinux context for home directory /home/widgeta/
```

bachradsusi:
It would be this bug https://bugzilla.redhat.com/show_bug.cgi?id=1321375 . A workaround is to use useradd in chroot instead of useradd with -R

```
^_^ useradd -R /mnt/sysimage -g 100 -G users,wheel -d /home/adminal2 -m -s /usr/bin/bash -u 1203 -c "Administrator account" adminal2
useradd: cannot set SELinux context for home directory /home/adminal2
```
vs
```
^_^ chroot /mnt/sysimage useradd -g 100 -G users,wheel -d /home/adminal2 -m -s /usr/bin/bash -u 1203 -c "Administrator account" adminal2
```

PR #524, opened 2018-11-14
Fix reporting of coverage results to coverall.io

### LINKED ISSUE #522: Missing coverage report in coveralls
In a few of my recent PRs the coverage report sent to coveralls didn't seem to work. I assumed it would sort itself out when I merged things onto master. But it didn't.

https://coveralls.io/builds/20059957

I must have broken something. But I'm not sure what, or even how to debug it. The coverage report is displayed when running the test, so it's not like it is totally missing.

https://travis-ci.org/weldr/lorax/builds/454128780

bcl:
When I re-arranged the test-in-docker I didn't realize how .travis.yml
was extracting the results. This should fix it.

Fixes #522

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__/
 .coverage
 pylint-log
 .pytest_cache/
+.test-results/
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,7 @@ script:
 after_success:
   - |
 
-        sudo docker create --name results-cont welder/lorax /bin/echo
-        sudo docker cp results-cont:/lorax/.coverage .coverage.docker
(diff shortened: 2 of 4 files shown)


atodorov:
```
Submitting coverage to coveralls.io...
No source for /lorax-test/src/pylorax/__init__.py
```

Now the problem is that when you copy the coverage data to the host the sources are located in a different directory. See this example (from anaconda) how to configure where coverage looks for files:

```
$ cat .coveragerc 
# used by coverage combine
# when combining Jenkins and kickstart_tests data
# doesn't affect combination of pyanaconda tests data
[paths]
source = .
        /usr/lib*/python*/site-packages/
        /usr/sbin/
```

PR #525, opened 2018-11-14
Rhel7 extras selinux enforcing

bcl:
Unlike the rhel8-branch version this one works with tar.


bcl:
This has been merged. rhbz#1654795

PR #527, opened 2018-11-17
Install grubby-deprecated package for ARMv7

martinezjavier:
Most bootloaders used in Fedora already have BootLoaderSpec but extlinux
doesn't yet. So when using extlinux the old grubby must to be installed.

Resolves: rhbz#1649778

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>


atodorov:

PR #529, opened 2018-11-21
Use --privileged mode for Docker and enable more tests

atodorov:
in addition make it possible to re-use `test_cli.sh` for setup/teardown/report if a list of other test scripts has been passed as parameters.


atodorov:
@bcl building a partitioned-disk image fails even with --privileged, see Travis log. Do you have any hints for an easy fix? If not I will try enabling this test (possibly ext4 one as well) in Jenkins instead of Travis.

bcl:
Looks simple enough.

PR #530, opened 2018-11-22
Add make targets for cloud image tests executed  by Jenkins

atodorov:
this is based on #529 and is still WIP until I figure out how to properly configure all the necessary creds in Jenkins.


atodorov:
@bcl this should be ready for review. I am still getting an error where the image running in AWS doesn't accept ssh logins with *supposedly* the correct key. ATM I have no idea why that is so maybe the last commit is not necessary.

atodorov:
OK, figured it out. For fedora images I can login as the `fedora` user instead of `ec2-user`, while on RHEL (IIRC) I was able to login as  `ec2-user`.  Is this something coming from cloud-init or lorax ? Doesn't seem to be related to our code.

bcl:
> OK, figured it out. For fedora images I can login as the `fedora` user instead of `ec2-user`, while on RHEL (IIRC) I was able to login as `ec2-user`. Is this something coming from cloud-init or lorax ? Doesn't seem to be related to our code.

If you're using the same recipe for both then the user should be whatever is set in [[customizations.user]], so I'd guess this is a cloud-init difference.

PR #532, opened 2018-11-29
Teach AWS test to distinguish between RHEL and Fedora

atodorov:


atodorov:
From these very same docs:

>  For a Fedora AMI, the user name is ec2-user **or** fedora.
>  For a RHEL AMI, the user name is ec2-user **or** root.

So a preference is given to ec2-user first and then something else second. That differes only for CentOS, Debian and Ubuntu. 

> If we change these to be ec2-user, we're creating images that differ from the officially released ones.

We are not, we are still complying with the expected defaults, see the **or** above!

larskarlitski:
> We are not, we are still complying with the expected defaults, see the or above!
 
Yeah, we're still complying with the defaults indeed. However, my point is that composer has no business in changing configuration of the generated images. They should be as close to the ones that can be downloaded from official mirrors as possible. If we think the defaults in `/etc/cloud/cloud.cfg` are wrong, we ought to change them in the distributions themselves.

Doubly so, because we seem to be making this change only for our own tests.

atodorov:
> Doubly so, because we seem to be making this change only for our own tests.

There's a difference in cloud-init defaults between RHEL and Fedora. On RHEL the default from `cloud.cfg` is to create a `cloud-init` user while on Fedora it defaults to creating the `fedora` user. Because of the RHEL defaults we have the sed line in ami.ks to make sure we're building an AMI where the default username is `ec2-user`.   Why that should not be the case for Fedora, given that AWS docs prefer their own username ? 

Also personally everyone I know who uses AWS knows ec2-user exists but we should examine our customers expectations here.

Look at https://bugzilla.redhat.com/show_bug.cgi?id=1654628 for the response from cloud-init maintainers. Not sure what they have to say but my feeling is this needs to be fixed (or documented) on our side.

Tests have just exposed this difference and the fix is not intended to satisfy them. We could just as easily update the test script to detect whether it is running on RHEL or Fedora and try to login with a different username.

bcl:
I think that for now we should match what the existing images are using. If that's different between distributions then the tests need to account for that. Also, one small note about the wording of the bz, this isn't 'breaking' anything in lorax-composer. It does make the tests more complex though.

PR #538, opened 2018-12-05
Use the correct path when pushing vmware.toml during testing

atodorov:


atodorov:
@bcl there are still a few failures but if it looks good to you we can merge:

1) OpenStack fails - unable to login via SSH, not sure why they ssh key isn't getting accepted
2) Azure fails - the storage account for uploading VHD blobs was missing so I recreated it but somehow I'm failing to match all of the previous settings. @larskarlitski maybe you know something specific that I don't ?

PR #545, opened 2018-12-10
WIP: Backport cloud image tests for RHEL 8

atodorov:
I made a git diff between master and rhel8-branch and applied the changes that affect cloud image tests. There cloud be things that I am still missing and they will be resolved in subsequent commits.

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,18 +1,37 @@
-FROM fedora:28
-
-RUN dnf -y install make libgit2-glib tito python3-pylint  \
-                    python3-nose python3-mako python3-flask \
-                    python3-coverage libselinux-python3 sudo \
-                    pykickstart python3-pytoml python3-sphinx \
-                    python3-semantic_version python3-rpmfluff \
-                    anaconda-tui python3-gevent beakerlib && \
-    useradd weldr
-
-RUN mkdir /lorax
-COPY . /lorax
-# remove byte-compiled files to avoid issues between Python 2/3
-# this can happen when you switch between rhel7 and master branches
-RUN find /lorax -name "*.pyc" -exec rm -f {} \;
-
-WORKDIR /lorax
-RUN make test
+FROM registry.fedoraproject.org/fedora:28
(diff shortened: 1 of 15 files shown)


Dockerfile.test
```diff
@@ -1,18 +1,37 @@
...
-
-RUN mkdir /lorax
-COPY . /lorax
-# remove byte-compiled files to avoid issues between Python 2/3
-# this can happen when you switch between rhel7 and master branches
-RUN find /lorax -name "*.pyc" -exec rm -f {} \;
-
-WORKDIR /lorax
-RUN make test
+FROM registry.access.redhat.com/rhel8-beta
```
bcl: Is this expected to work outside the internal lan? If not, we need to stick with f28 until centos8 is released.

atodorov: docker pull works. From what I can tell these images are publicly available but they don't have any subscriptions enabled. Fedora 28 sounds like an easier workaround. I will probably go ahead with that and open a task to not forget to update the image at a later point.

atodorov:
@bcl: with Fedora 28 base docker image pylint gives me this:

```
*** Running pylint ***
PYTHONPATH=:./src/ ./tests/pylint/runpylint.py
************* Module /lorax/src/pylorax/dnfhelper.py
E1101(no-member): 97,21: LoraxRpmCallback.progress: Module 'dnf.transaction' has no 'PKG_INSTALL' member
E1101(no-member):105,23: LoraxRpmCallback.progress: Module 'dnf.transaction' has no 'TRANS_POST' member
```

bcl:
Ugh. So it looks like there are more dnf variations floating around than I thought, and the F28 version is now different from the RHEL8 version. So I guess F28 isn't suitable for running tests and pylint on anymore, and we don't have EPEL8 available yet, so some of these packages aren't available.
I don't see any way to reliably run these on Travis anymore.

Dockerfile.test
atodorov: Switched to Fedora 28 for now.

PR #548, opened 2018-12-13
[Request for Comments] Split image creation into modules

larskarlitski:
NOTE: This is a very crude first version of this idea to validate it and
gather early feedback. I've done just enough to make this work, with
lots of code commented out and (of course) nothing working like it did
before. It's quite a change in architecture and I'd like to discuss
before putting more work into it.

Image builds are now pipelines of modules. Each module is an executable
that takes a json object as configuration on stdin. This object contains
the path to a directory containing an os tree. The module can do
anything to that tree.

The first three modules are:

- io.weldr.rpm: installs rpm packages from arbitrary sources
- io.weldr.anaconda: runs anaconda with a kickstart file on that tree,
assuming that all packages are already installed (needs skippackages
patch in anaconda)
- io.weldr.fsimage: create an ext4 image out of that tree

The pipeline is hard coded in compose.py right now. The idea is that
there will be different pipelines for various input/output formats that
live as configuration in the file system.

Going forward, this allows us to:

- test more thoroughly, as we have smaller testable units
- diff the file system tree after each module is run (useful for
debugging, auditing, figuring out what anaconda is doing exactly)
- allow more experimentation, because the pipeline of creating images
becomes configuration
- modules are pluggable and can come from other sources - this is why
they're are prefixed with reverse domain names

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,9 @@ data_files.append(("/usr/sbin", ["src/sbin/lorax", "src/sbin/mkefiboot",
 data_files.append(("/usr/bin",  ["src/bin/image-minimizer",
                                  "src/bin/mk-s390-cdboot",
                                  "src/bin/composer-cli"]))
+data_files.append(("/usr/lib/lorax", ["src/modules/io.weldr.rpm",
+                                      "src/modules/io.weldr.anaconda",
+                                      "src/modules/io.weldr.fsimage"]))
 
 # get the version
 sys.path.insert(0, "src")
--- a/src/modules/io.weldr.anaconda
+++ b/src/modules/io.weldr.anaconda
@@ -1,0 +1,25 @@
+#!/usr/bin/python3
+
+from contextlib import contextmanager
+from types import SimpleNamespace
+import json
+import os
+import subprocess
+import sys
+import tempfile
+
+options = SimpleNamespace(**json.loads(sys.stdin.read()))
+
+with tempfile.NamedTemporaryFile('w', suffix='.ks') as f:
+    f.write(options.kickstart)
+    f.flush()
+
+    cmd = [
+        'anaconda',
+        '-d', '--cmdline',
+        '--skippackages',
+        '--kickstart', f.name,
+        '--dirinstall', options.tree
+    ]
+
+    subprocess.run(cmd, check=True)
--- a/src/modules/io.weldr.fsimage
+++ b/src/modules/io.weldr.fsimage
@@ -1,0 +1,26 @@
+#!/usr/bin/python3
+
+from contextlib import contextmanager
+from tempfile import TemporaryDirectory
+from types import SimpleNamespace
+import json
+import os
+import subprocess
+import sys
+
+options = SimpleNamespace(**json.loads(sys.stdin.read()))
+
+@contextmanager
+def mountloop(image, path):
+    try:
+        subprocess.run(['mount', '-o', 'loop', image, path], check=True)
+        yield
+    finally:
+        subprocess.run(['umount', path], check=True)
+
+
+subprocess.run(['truncate', '--size', '2G', options.target], check=True)
+subprocess.run(['mkfs.ext4', '-F', options.target], check=True)
+
+with TemporaryDirectory() as d, mountloop(options.target, d):
+    subprocess.run(['cp', '-a', os.path.join(options.tree, '.'), d], check=True)
--- a/src/modules/io.weldr.rpm
+++ b/src/modules/io.weldr.rpm
@@ -1,0 +1,44 @@
+#!/usr/bin/python3
+
+from configparser import ConfigParser
+from contextlib import contextmanager
+from types import SimpleNamespace
+import json
+import os
+import subprocess
+import sys
+
+options = SimpleNamespace(**json.loads(sys.stdin.read()))
+
+@contextmanager
+def bindmount(source, dest):
+    os.makedirs(source, 0o755, True)
+    os.makedirs(dest, 0o755, True)
+    subprocess.run(['mount', '--bind', source, dest], check=True)
+    try:
+        yield
+    finally:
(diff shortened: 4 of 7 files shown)


larskarlitski:
This depends on https://github.com/rhinstaller/anaconda/pull/1726

setup.py
```diff
@@ -24,6 +24,9 @@
 data_files.append(("/usr/bin",  ["src/bin/image-minimizer",
                                  "src/bin/mk-s390-cdboot",
                                  "src/bin/composer-cli"]))
+data_files.append(("/usr/lib/lorax", ["src/modules/io.weldr.rpm",
```
bcl: I'm not a fan of reverse domain names, when the project moves you end up either changing them and breaking things, or leaving them and confusing people. I'd rather use something like the way git does it, define a core set of commands, and a contribution set, and people just need to make sure not to stomp on each other's names.
eg. /usr/share/lorax/core/ and /usr/share/lorax/contrib or something similar.

Also, I think we should avoid calling them modules to prevent confusion with modularity. These will be stand-alone programs, or extensions, to lorax.

src/modules/io.weldr.anaconda
```diff
@@ -0,0 +1,25 @@
...
+from types import SimpleNamespace
+import json
+import os
+import subprocess
+import sys
+import tempfile
+
+options = SimpleNamespace(**json.loads(sys.stdin.read()))
+
+with tempfile.NamedTemporaryFile('w', suffix='.ks') as f:
```
bcl: I try to follow the convention of prefixing lorax's temporary things with lorax-something so that when they inevitably get left laying around the filesystem it's easy to identify them and remove them :)

src/modules/io.weldr.anaconda
```diff
@@ -0,0 +1,25 @@
...
+
+    cmd = [
+        'anaconda',
+        '-d', '--cmdline',
+        '--skippackages',
+        '--kickstart', f.name,
+        '--dirinstall', options.tree
+    ]
+
+    subprocess.run(cmd, check=True)
```
bcl: (I know this is a prototype so take theses comments in that context).
We should be using pylorax tools to do this. If novirt_install doesn't provide the right pieces, break it up into ones that do. Not running anaconda directly from here.

src/modules/io.weldr.fsimage
```diff
@@ -0,0 +1,26 @@
...
+@contextmanager
+def mountloop(image, path):
+    try:
+        subprocess.run(['mount', '-o', 'loop', image, path], check=True)
+        yield
+    finally:
+        subprocess.run(['umount', path], check=True)
+
+
+subprocess.run(['truncate', '--size', '2G', options.target], check=True)
```
bcl: Similar comment here, we have lots of tools in pylorax to do these things.

src/modules/io.weldr.rpm
```diff
@@ -0,0 +1,44 @@
...
+     bindmount('/dev', f'{options.tree}/dev'), \
+     bindmount('/sys', f'{options.tree}/sys'):
+
+    cmd = ['dnf', '-yv',
+           f'--installroot={options.tree}',
+           '--releasever=29',
+           '--setopt=reposdir=None',     # don't use repos from the host
+           'install', 'chrony', 'firewalld'] + options.packages
+
+    subprocess.run(cmd, check=True)
```
bcl: Use pylorax, etc. which also allows for applying templates, etc.

larskarlitski:
> I like the general idea here, but I think it needs to be done differently:
> 
>  * New extensions should import pylorax modules and use them to compose their features.

Of course. My intention is that these modules (I agree that the name "modules" is unfortunate) can be any executable pulling in whatever libraries it wants. The initial modules would use all the things that already exists in pylorax. I didn't use those features simply because I'm not too familiar with lorax and this is only meant to show off the idea for now.

>  * lorax doesn't call extensions, extensions use pylorax, so I really don't like the idea of the pipeline there in compose.py

The pipeline concept is the core of this proposal: to have modules that are plain executables with a simple API (take in a json config and an os tree and do something to that tree). I believe that using these modules for every part of the build process and allowing for 3rd party drop-in modules allows for greater flexibility and experimentation. 

>  * If pylorax doesn't supply parts in the right form we can add and adjust things so that it does. eg. add a json parser that then calls the pylorax functions with the right args.

Yep! I'd use as much from pylorax as possible in these modules (that's the tried and tested code). My idea was to replace the hard-coded installation procedure in `novirt_install`. 

>  * existing tools should still work.

Yeah, I very much agree. This is why I'm suggesting this here instead of starting something new. It is quite the change. I do understand if you feel like this goes too much against the current way that lorax does things.

> The general idea being that pylorax provides the libraries allowing people to construct useful apps from it. livemedia-creator has, somewhat unsuccessfully, tried to be composable. eg. the --image-only option and then passing in the image to make an iso with. I've also thought it would be useful to be able to pass in a directory tree, but haven't implemented that yet.

Of course. However, I believe that we gain a lot if we additionally allow for 3rd party modules that don't use pylorax. It would be great if I could just plug in a very simple script that uses rpm-ostree to build the image, or download an existing cloud image and modify that, or use something experimental like Harald's [FedoraBook](https://pagure.io/Fedorabook) — all without having to learn about pylorax, but being able to leverage the other modules it provides.

setup.py
larskarlitski: They're reverse domain name to allow drop-in from different packages. I've been bitten too many times by conflicting names for these kinds of things :)

"modules" is an unfortunate name, I agree. How about "stages"?

src/modules/io.weldr.anaconda
larskarlitski: Oh, that's a great idea!

bcl:
Closing this here, discussion continues at https://github.com/larskarlitski/osbuild

PR #550, opened 2018-12-14
Scripts for removing old virtual machines from Amazon EC2 and OpenStack

jstodola:


atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

atodorov:

tests/cleanup/remove_old_vms_aws.sh
```diff
@@ -0,0 +1,71 @@
...
+        else
+            rlLogInfo "AWS_SECRET_ACCESS_KEY is configured"
+        fi
+
+        # VMs older than HOURS_LIMIT will be deleted
+        HOURS_LIMIT="${HOURS_LIMIT:-24}"
+        export TIMESTAMP=`date -u -d "$HOURS_LIMIT hours ago" '+%FT%T'`
+        export AWS_REGION="${AWS_REGION:-us-east-1}"
+
+        rlLogInfo "HOURS_LIMIT=$HOURS_LIMIT"
```
atodorov: will probably make sense to log the TIMESTAMP as well. If we need to debug something we can see the local timestamp from the machine running this script and then compare with timestamps logged from Ansible.

tests/cleanup/remove_old_vms_aws.sh
```diff
@@ -0,0 +1,71 @@
...
+        rlLogInfo "AWS_REGION=$AWS_REGION"
+
+        for package in ansible python3-boto3; do
```
atodorov: You can also do `pip install ansible[amazon]` to fetch all dependencies. 

@bcl given that we trash the system running this script afterwards which one do you prefer: pip or dnf ?

tests/cleanup/remove_old_vms_aws.sh
jstodola: Yes, it can be logged here. On the other hand, TIMESTAMP is printed later in case you need it:
rlLogInfo "Removing VMs created before $TIMESTAMP"

tests/cleanup/remove_old_vms_aws.sh
bcl: dnf. reduces the chances for differences resulting in weird problems.

atodorov:
@jstodola I tried to AWS script and it worked for me, more notes:

1) Probably it will be best to split the two scripts into 2 separate pull requests to make reviewing/updating easier given the next items
2) The script should cycle through all AWS regions, not only us-east-1. It is true that us-east-1 is probably the most commonly used one but we do have people in Europe and a few times I've seen VMs created in other regions
3) Script should also deregister any AMI images that are older than timestamp
4) Remove the volumes for these deregistered AMI images (or maybe even all volumes that are older)
5) Remove key pairs with name starting with "composer-" (not sure if that supports regions)
6) Remove any files on S3 that are older (S3 is supported globally, no regions here)

atodorov:
> Closing this PR, new PRs will be created for each cloud provider.

FYI next time leave one of the items in the existing PR and create a new one only for the other cloud env.

PR #557, opened 2018-12-21
WIP, DO NOT MERGE, DEBUG: log server and AWS time

atodorov:
so we can debug a RequestTimeTooSkewed error when uploading to S3

### DIFF
--- a/tests/cli/test_build_and_deploy_aws.sh
+++ b/tests/cli/test_build_and_deploy_aws.sh
@@ -63,6 +63,14 @@ aws_secret_access_key = $AWS_SECRET_ACCESS_KEY
 __EOF__
         fi
 
+        rlLogInfo "**** server date"
+        rlRun -t -c "date"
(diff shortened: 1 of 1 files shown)


atodorov:
@bcl for the record this executes on a RHEL 7.5 host and what I see is that files under /etc get modified when we start creating the compose. In particular /etc/localtime is modified which skews the local time on the system and causes Amazon to refuse the upload.

I'm able to reproduce manually on the 7.5 VMs we have in OpenStack and couldn't do this on 7.6 provisioned outside OS.

I saw this happening in the beginning of anaconda provisioning the image. Not sure if a bug that has been already fixed or the particular combination of packages that we have on the Jenkins slave.

atodorov:
I think we've been hitting a bug where running anaconda somehow updated files under /etc on the host. I've been able to reproduce the same behavior manually on 7.5.

I've managed to find a good 7.6 image which works with our Jenkins config and that seems to proceed past the AWS upload issue so I'm closing this PR and will open another one with updates related to tests.

PR #561, opened 2019-01-06
Pass dnf's ssl options to anaconda to enable RHEL subscription content

larskarlitski:
This makes the [reposync workaround](https://weldr.io/Running-Composer-on-RHEL/) unnecessary when creating images from RHEL systems.

See individual commits for explanation.

Depends on https://github.com/dcantrell/pykickstart/pull/250 and https://github.com/rhinstaller/anaconda/pull/1745/files

Before merging, let's
- [ ] figure out if we can talk to subscription-manager without being root

### DIFF
--- a/src/sbin/lorax-composer
+++ b/src/sbin/lorax-composer
@@ -244,7 +244,8 @@ if __name__ == '__main__':
 
     start_queue_monitor(server.config["COMPOSER_CFG"], uid, gid)
 
-    # Drop root privileges on the main process
+    # Change user and group on the main process.  Note that this still happens even if
+    # --user and --group were passed in, but changing to the same user should be fine.
     os.setgid(gid)
     os.setuid(uid)
     log.debug("user is now %s:%s", os.getresuid(), os.getresgid())
--- a/lorax.spec
+++ b/lorax.spec
@@ -47,7 +47,7 @@ Requires:       kpartx
 # Python modules
 Requires:       libselinux-python3
 Requires:       python3-mako
-Requires:       python3-kickstart
+Requires:       python3-kickstart >= 3.19
 Requires:       python3-dnf >= 3.2.0
 Requires:       python3-librepo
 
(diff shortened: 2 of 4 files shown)


src/sbin/lorax-composer
```diff
@@ -240,10 +240,11 @@ if __name__ == '__main__':
 
     start_queue_monitor(server.config["COMPOSER_CFG"], uid, gid)
 
-    # Drop root privileges on the main process
```
bcl: We should just cherry-pick over the rhel8 commit dbc4d08de40521ca8b1d64c3c0f36e6fc67be03c

atodorov:
@larskarlitski tests failing with

```
2019-01-08 13:34:45,882: Non interactive installation failed: Problems in request:
2019-01-08 13:34:45,882: missing packages: krb5-libs-1.16.1-23.fc29.x86_64.
```

I can't find kerberos or krb5-libs defined in the blueprints so something must be pulling that as a dependency. Not sure if we need to bump the version somewhere in the code or package is missing on the mirrors.

src/sbin/lorax-composer
larskarlitski: Ah that's much better. Done.

larskarlitski:
> I can't find kerberos or krb5-libs defined in the blueprints so something must be pulling that as a dependency. Not sure if we need to bump the version somewhere in the code or package is missing on the mirrors.

I don't think these changes pull those in. The only difference is the version bump for pykickstart.

By the way, that version of pykickstart hasn't been released yet. We should probably wait with merging this.

wzzrd:
Tried this out after talking to Lars last Friday, and this patch works for me.

larskarlitski:
The pykickstart change landed in `3.19`. I've updated the spec file to reflect that and rebased on master. I think this is ready to go.

bcl:
Thanks, I think we need to wait for pykickstart to hit the repos first, otherwise installing it will fail. It isn't on dl.fedoraproject.org yet.

bcl:
pykickstart 3.20 is in rawhide, merging this onto master.

PR #567, opened 2019-01-15
Add script for removing old virtual machines from OpenStack

jstodola:
Previous PR: https://github.com/weldr/lorax/pull/550

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ test_vmware:
 
 clean_cloud_envs:
 	sudo -E ./tests/cleanup/remove_old_objects_aws.sh
+	sudo -E ./tests/cleanup/remove_old_objects_openstack.sh
 
 clean:
 	-rm -rf build src/pylorax/version.py
--- a/tests/cleanup/remove_old_objects_openstack.sh
+++ b/tests/cleanup/remove_old_objects_openstack.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+# Script removes virtual machines and other artifacts older than HOURS_LIMIT (24 hours by default) from OpenStack
+
+. /usr/share/beakerlib/beakerlib.sh
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        if [ -z "$OS_AUTH_URL" ]; then
+            rlFail "OS_AUTH_URL is empty!"
+        else
+            rlLogInfo "OS_AUTH_URL=$OS_AUTH_URL"
+        fi
+
+        if [ -z "$OS_USERNAME" ]; then
+            rlFail "OS_USERNAME is empty!"
+        else
+            rlLogInfo "OS_USERNAME is configured"
+        fi
+
+        if [ -z "$OS_PASSWORD" ]; then
+            rlFail "OS_PASSWORD is empty!"
+        else
+            rlLogInfo "OS_PASSWORD is configured"
+        fi
+
+        # VMs older than HOURS_LIMIT will be deleted
+        HOURS_LIMIT="${HOURS_LIMIT:-24}"
+        export TIMESTAMP=`date -u -d "$HOURS_LIMIT hours ago" '+%FT%T'`
+
+        rlLogInfo "HOURS_LIMIT=$HOURS_LIMIT"
+
+        for package in ansible python3-openstacksdk python3-openstackclient; do
+            if ! rlCheckRpm "$package"; then
+                rlRun -t -c "dnf -y install $package"
+                rlAssertRpm "$package"
+            fi
+        done
+
+        PLAYBOOK_DELETE_VMS=`mktemp`
+        PLAYBOOK_DELETE_IMAGES=`mktemp`
+    rlPhaseEnd
+
+    rlPhaseStartTest "Delete old VMs"
+# The openstack_servers variable used in the playbook bellow is set by the os_server_facts ansible module.
+# The variable contains details about all discovered virtual machines.
+# See https://docs.ansible.com/ansible/latest/modules/os_server_facts_module.html
+        cat > $PLAYBOOK_DELETE_VMS << __EOF__
+- name: Delete old VMs
+  hosts: localhost
+  gather_facts: False
+  tasks:
+  - name: Get VMs
+    os_server_facts:
+
+  - name: List all VMs
+    debug:
+      var: openstack_servers
+
+  - name: Delete old VMs
+    os_server:
+      name: "{{item.id}}"
+      state: absent
+    loop: "{{openstack_servers}}"
+    when: item.created < lookup('env','TIMESTAMP')
+    loop_control:
+      label: "{{item.name}} (id: {{item.id}} created: {{item.created}})"
+__EOF__
+
+        rlLogInfo "Removing VMs created before $TIMESTAMP"
+        rlRun -t -c "ansible-playbook $PLAYBOOK_DELETE_VMS"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Delete old images"
+# The openstack_image variable used in the playbook bellow is set by the os_image_facts ansible module.
+# The variable contains details about all discovered images.
+# See https://docs.ansible.com/ansible/latest/modules/os_image_facts_module.html
+        cat > $PLAYBOOK_DELETE_IMAGES << __EOF__
+- name: Delete old images
+  hosts: localhost
+  gather_facts: False
+  tasks:
+  - name: Get images
+    os_image_facts:
+
+  - name: Delete old images
+    os_image:
+      name: "{{item.name}}"
+      id: "{{item.id}}"
+      state: absent
+    loop: "{{openstack_image}}"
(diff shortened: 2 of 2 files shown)


tests/cleanup/remove_old_vms_openstack.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+# Script removes virtual machines older than HOURS_LIMIT (24 hours by default) from OpenStack
+
+# TODO
+# * remove old images
```
atodorov: The name pattern is `Composer-$UUID-Automated-Import`.

Also need to remove volumes b/c sometimes they may not be removed automatically after removing the VM.

jstodola:
Makefile with the clean_cloud_envs target has been added. Also some small changes in the script have been included to match with other cleanup scripts being prepared.

tests/cleanup/Makefile
```diff
@@ -0,0 +1,2 @@
+clean_cloud_envs:
+	find -type f -executable -exec {} \;
```
atodorov: This target needs to be in the master Makefile, Jenkins knows to look only in there.

tests/cleanup/remove_old_artifacts_openstack.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+# Script removes virtual machines and other artifacts older than HOURS_LIMIT (24 hours by default) from OpenStack
+
+# TODO
+# * remove old images (The name pattern is Composer-$UUID-Automated-Import)
+# * remove volumes b/c sometimes they may not be removed automatically after removing the VM
```
atodorov: Are these 2 todo items going to land in this PR or no ?

Makefile
```diff
@@ -65,6 +65,9 @@ test_openstack:
 test_vmware:
 	sudo -E ./tests/test_cli.sh tests/cli/test_build_and_deploy_vmware.sh
 
+clean_cloud_envs:
+	find ./tests/cleanup/ -type f -executable -exec {} \;
```
atodorov: Errr, nope. Why not just list the exact script files we want to execute so it is evident what is going on?

jstodola:
Updated Makefile to run each cleanup script separately and using sudo (because of installation of packages in the script).
Previous changes in the cleanup script were: using temporary files for ansible playbook files and change of the script name.

atodorov:
> Updated Makefile to run each cleanup script separately and using sudo (because of installation of packages in the script).
Previous changes in the cleanup script were: using temporary files for ansible playbook files and change of the script name.

That looks good to me. What about the 2 TODO items in the script comments ?

jstodola:
The remaining 2 TODOs will be completed in a future commit, probably next week.
Removing of old volumes was blocked because of:
a) limited functionality of ansible modules (no module to list volumes exists yet)
b) openstack cli has been failing because of bug 1663043. The fix should be in stable repo later this week.

atodorov:
> Removing of old volumes was blocked because of:

What about removal of uploaded images ? Is that affected by the issues below?

> a) limited functionality of ansible modules (no module to list volumes exists yet)

Is this still a problem or does Ansible have everything you need ?

> b) openstack cli has been failing because of bug 1663043. The fix should be in stable repo later this week.

or are you going to use openstack cli instead of Ansible to remove volumes? 

If using the openstack cli I think it will be possible for the script to enable the updates-testing repository and install from there (probably needs to assert minimum rpm version). This is to avoid blocking work on this item. While development and review completes the RPM will probably be in the stable repo anyway.

jstodola:
Removal of uploaded images is not blocked, but will require careful testing to avoid accidental removal of images not uploaded by composer-jenkins.
Ansible module for listing volumes does not exist, so I'm going to use openstack cli.
I would like to avoid using testing repository in the script, but I can work on the script with the testing repo enabled on my system.
Both of these features could be ready for review next week if you prefer to wait for the script to be feature complete.

atodorov:
> Removal of uploaded images is not blocked, but will require careful testing to avoid accidental removal of images not uploaded by composer-jenkins.

OpenStack doesn't allow me to delete images I didn't upload myself so between that and the search by name I think we're pretty safe.

> I would like to avoid using testing repository in the script, but I can work on the script with the testing repo enabled on my system.

This is what I meant: no need to slow down development work just by waiting for a package to get promoted through the Fedora infrastructure.

For the record: b/c the package will probably land in the stable repo by Monday we can wait for it but if that was not the case I would go for the `updates-testing` enabled so the current task can be completed.

jstodola:
The script has been extended and now it can delete old images and volumes.

Do not merge this PR yet, since removing of old volumes is not working yet on Fedora 29 without the updates-testing repo enabled (because of https://bugzilla.redhat.com/show_bug.cgi?id=1663043).
It will be clarified if we will wait for the fixed package to appear in the updates repository or if enabling of the updates-testing repo will be needed.

tests/cleanup/remove_old_objects_openstack.sh
```diff
@@ -0,0 +1,121 @@
...
+            rlFail "OS_USERNAME is empty!"
+        else
+            rlLogInfo "OS_USERNAME is configured"
+        fi
+
+        if [ -z "$OS_PASSWORD" ]; then
+            rlFail "OS_PASSWORD is empty!"
+        else
+            rlLogInfo "OS_PASSWORD is configured"
+        fi
```
atodorov: Ansible will need `OS_TENANT_NAME`, which usually matches `OS_USERNAME` see the test script in `tests/cli`. Isn't this failing for you without this env variable defined ?

tests/cleanup/remove_old_objects_openstack.sh
```diff
@@ -0,0 +1,121 @@
...
+  hosts: localhost
+  gather_facts: False
+  tasks:
+  - name: Get images
+    os_image_facts:
+
+# listing details about all images can produce tens of thousands of lines, so don't do it
+#  - name: List all images
+#    debug:
+#      var: openstack_image
```
atodorov: If this is not going to be used then remove it.

Question: how about the `openstack_image` variable ? If this section in the playbook is commented out how does looping on this variable work on line 89 below ?

tests/cleanup/remove_old_objects_openstack.sh
```diff
@@ -0,0 +1,121 @@
...
+
+    rlPhaseStartTest "Delete old volumes"
+        volume_list=`openstack-3 volume list --format value --column ID`
+        for volume in $volume_list; do
+            creation_date=`openstack-3 volume show $volume --column created_at --format value`
+            if [ $? -ne 0 ]; then
+                rlLogWarning "Failed to get the creation date of volume $volume"
+                continue
+            fi
+            if [ "$creation_date" \< "$TIMESTAMP" ]; then
```
atodorov: What date/time format is this? Are we certain that comparing the dates as strings will compare them in chronological order ?

Also why `\<` ? Isn't `<` enough ?

atodorov:
For the record: I vote for enabling updates-testing repository and merging this PR asap b/c it has been open for over 1 month.

jstodola:
Changes in the latest version:
* Removed the part of ansible playbook listing all images.
* Use updates-testing repo to update the python3-osc-lib package to avoid hitting bug 1663043

tests/cleanup/remove_old_objects_openstack.sh
```diff
@@ -0,0 +1,120 @@
...
+  tasks:
+  - name: Get images
+    os_image_facts:
+
+  - name: Delete old images
+    os_image:
+      name: "{{item.name}}"
+      id: "{{item.id}}"
+      state: absent
+    loop: "{{openstack_image}}"
```
atodorov: Question: how does the `openstack_image` variable work? Where does it come from ?

atodorov:
> The latest version doesn't use updates-testing repo, since the fix for bug 1663043 is available in the updates repository.

this is good news. Can you respond/fix the remaining 3 comments above.

jstodola:
Oops, commends were sent now, I forgot to publish them, I'm sorry.

tests/cleanup/remove_old_objects_openstack.sh
atodorov: that's strange but if it works ok then

tests/cleanup/remove_old_objects_openstack.sh
atodorov: OK, this is important to know. 

Can you add this comment in the source code. Maybe same for the `openstack_image` and `openstack_servers` variables so that when somebody else reads this script later they will know where these things are coming from and why they are written in this way.

Other than the comments this is ready to be merged IMO.

jstodola:
Added comments to the script describing the openstack_servers and openstack_image variables used in the ansible playbooks, and also a comment clarifying the date/time comparison.

tests/cleanup/remove_old_objects_openstack.sh
jstodola: Comments have been added in the latest version of the script.

PR #571, opened 2019-01-21
lorax: Move default tmp dir to /var/tmp/lorax

bcl:
If systemd's tmpfiles.d timer is executed while lorax is running it will
remove any files and directories older than 30 days. This is what has
been causing the occasional error where /proc/ would seem to vanish
during the install.

Upstream has proposed this solution, https://github.com/systemd/systemd/pull/11482
but until that is released we need a work-around to protect the lorax
files.

This commit does several things:

* Move the default tmpdir from /var/tmp/ to /var/tmp/lorax/
* Add a lorax.conf tmpfiles.d file that
  1) removes everything under /var/tmp/lorax/ at boot time.
  2) prevents systemd-tmpfiles from removing anything under /var/tmp/lorax/
     (except at boot time).
* Add an exit handler to lorax so that temporary directories are removed on
  exit or on a python traceback.

Note that this will not remove the tempdir if lorax is killed with a
signal -- those are being caught by dnf and prevent the exit handler
from running.

If you want to keep the working directory around for debugging purposes
use --workdir /var/tmp/lorax/my-workdir and it won't be removed up by
lorax.

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -204,6 +204,7 @@ getent passwd weldr >/dev/null 2>&1 || useradd -r -g weldr -d / -s /sbin/nologin
 %config(noreplace) %{_sysconfdir}/lorax/lorax.conf
 %dir %{_datadir}/lorax
 %{_mandir}/man1/*.1*
+%{_tmpfilesdir}/lorax.conf
 
 %files lmc-virt
 
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,8 @@ data_files = [("/etc/lorax", ["etc/lorax.conf"]),
               ("/etc/lorax", ["etc/composer.conf"]),
               ("/usr/lib/systemd/system", ["systemd/lorax-composer.service",
                                            "systemd/lorax-composer.socket"]),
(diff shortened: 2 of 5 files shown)


bcl:
There's one problem with this PR -- selinux is blocking some parts of the temporary directory from being cleaned up becuause shadow- and opasswd have this set on them:
scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=unconfined_u:object_r:shadow_t:s0

bcl:
https://bugzilla.redhat.com/show_bug.cgi?id=1668054 selinux change should fix this.

bcl:
Pushed a fix that doesn't need SELinux changes, use flock to protect directories in use from being deleted by lorax when it runs.

AdamWill:
Sorry for the delay @bcl (I know you asked me to take a look at this). I did a scratch build and [threw it at openQA](https://openqa.stg.fedoraproject.org/tests/overview?build=Kojitask-32327983-NOREPORT&groupid=2&distri=fedora&version=29). As you can see, it passed. I disabled the workaround for the bug in that run, too, so it seems that it not only works OK, but actually does fix the bug too (one run can't prove that, but it's a strong indication). So :+1:

PR #572, opened 2019-01-22
Switch to xorriso for iso creation

bcl:


docs/fedora-livemedia.ks
```diff
@@ -389,4 +388,13 @@ syslinux
 -gfs2-utils
 -reiserfs-utils
 
+# This package is needed to boot the iso on UEFI
+shim
+shim-ia32
+grub2
+grub2-efi
+grub2-efi-*-cdboot
+grub2-efi-ia32
```
Conan-Kudo: Shouldn't this be `grub2-efi-*`?

docs/fedora-livemedia.ks
```diff
@@ -389,4 +388,13 @@ syslinux
 -gfs2-utils
 -reiserfs-utils
 
+# This package is needed to boot the iso on UEFI
+shim
+shim-ia32
```
Conan-Kudo: I think this is actually `shim-*` since it's `shim-x64` and `shim-ia32`.

docs/fedora-livemedia.ks
```diff
@@ -389,4 +388,13 @@ syslinux
...
+grub2
```
Conan-Kudo: This should be `grub2-pc`?

docs/fedora-livemedia.ks
bcl: No, it's just the basics needed to get it to boot on UEFI.

docs/fedora-livemedia.ks
bcl: I'm not sure, grub2 works fine in my tests.

bofh80: fedora docs suggest "dnf install grub2" which appears to be a magic package that selected grub2-pc on my f29

docs/fedora-livemedia.ks
bofh80: Just for clarity again, grub2-efi is a magic package that selects grub2-efi-x64 on f29.

docs/fedora-livemedia.ks
bofh80: shim is a magic package that installs shim-x64. i would say it's a meta package. but it doesn't come up in the dnf search. hence magic.

AdamWill:
For the record, I'm planning to throw this at openQA and test it as @bcl asked, but I sort of got sucked into a huge yak shave by it, because I figured to test something like this properly we should at least also generate a live ISO and test both BIOS and UEFI cases for both ISOs...so I'm making openQA do that. It's nearly there. :D

lorax.spec
```diff
@@ -27,7 +27,7 @@ Requires:       dosfstools
 Requires:       e2fsprogs
 Requires:       findutils
 Requires:       gawk
-Requires:       genisoimage
+Requires:       xorrisofs
```
AdamWill: This package does not seem to exist in Fedora.

Conan-Kudo: Ah yes, this package should be `xorriso`.

AdamWill: I don't think it's that simple. The package called `xorriso` was last built in 2013 for Fedora 20 and has been retired since Fedora 22, seemingly (f21 is the last release tag it has).

Conan-Kudo: `xorriso` source package is `libisoburn`. It stopped being a separate source package in 2013.

AdamWill: Ah, thanks. I was commenting from my phone so couldn't look up package provides :)

PR #573, opened 2019-01-23
Test related updates for RHEL 8

atodorov:
Now that we finally have working RHEL8 slaves I can start fixing missing bits and pieces around the available test suite.

If there are more changes needed they will land here as additional commits.


tests/pylorax/blueprints/example-http-server.toml
```diff
@@ -20,7 +20,7 @@ version = "7.2.*"
 
 [[packages]]
 name = "tmux"
-version = "2.8"
+version = "*"
```
bcl: You will also need to change all the places where we look for 2.8 in the test results. eg. tests/pylorax/test_server.py,

atodorov: Well, turns out not all of them but only 2. The rest will be fine as long as Fedora has version 2.8.

atodorov:
All 5 test jobs in Jenkins passed. Merging.

PR #578, opened 2019-01-24
Remove duplicate repositories from the sources list

bcl:
In some cases when the host has, for whatever reason, multiple copies of
the same repo listed the build may fail with an error about running out
of space.

So this commit removes duplicate entries after the host's repos have been
loaded. It also adjusts some of the test repos to use different
temporary repo names for the tests.

Resolves: rhbz#1664128

### DIFF
--- a/src/pylorax/api/yumbase.py
+++ b/src/pylorax/api/yumbase.py
@@ -173,6 +173,32 @@ def get_base_object(conf):
         if any(map(lambda pattern: fnmatchcase(name, pattern), enabled_repos)):     # pylint: disable=cell-var-from-loop
             yb.getReposFromConfigFile(repo_file)
 
+    # Remove any duplicate repo entries. These can cause problems with Anaconda, which will fail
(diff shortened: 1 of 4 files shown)


atodorov:
@bcl this seems to resolve the problems I've seen in rhbz#1664128.

Tests for cloud envs are failing due to unrelated issue, see #579. Only the vmware image build failed with similar error as before. Let's see if that is consistent.

bcl:
Merged into rawhide, f29, f28.

atodorov:
@bcl: if you are waiting for the RHEL7 tests to pass don't. They are fixed in #579 which depends on this one.

atodorov:
@bcl - can you merge in `rhel7-extras` as well. This is blocking other PRs.

bcl:
> @bcl - can you merge in `rhel7-extras` as well. This is blocking other PRs.

The bug needs acks first.

PR #581, opened 2019-01-25
Export OS_PROJECT_NAME variable in openstack cleanup script

jstodola:
The OS_PROJECT_NAME (or OS_TENANT_NAME) environment variable needs to be defined.
Use the OS_PROJECT_NAME, since it is recommended in the documentation instead of
the older OS_TENANT_NAME.

### DIFF
--- a/tests/cleanup/remove_old_objects_openstack.sh
+++ b/tests/cleanup/remove_old_objects_openstack.sh
@@ -24,6 +24,9 @@ rlJournalStart
             rlLogInfo "OS_PASSWORD is configured"
         fi
 
+        export OS_PROJECT_NAME="${OS_PROJECT_NAME:-$OS_USERNAME}"
+        rlLogInfo "OS_PROJECT_NAME=$OS_PROJECT_NAME"
+
         # VMs older than HOURS_LIMIT will be deleted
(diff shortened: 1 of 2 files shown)


tests/cleanup/remove_old_objects_openstack.sh
```diff
@@ -24,6 +24,9 @@ rlJournalStart
             rlLogInfo "OS_PASSWORD is configured"
         fi
 
+        export OS_PROJECT_NAME="${OS_PROJECT_NAME:-$OS_USERNAME}"
+        rlLogInfo "OS_PROJECT_NAME=$OS_PROJECT_NAME"
```
atodorov: Since you say it is recommended in the docs that means OS_TENANT_NAME should be changed to OS_PROJECT_NAME everywhere else :-(

jstodola: OS_TENANT_NAME replaced with OS_PROJECT_NAME in tests/cli/test_build_and_deploy_openstack.sh too.
OS_TENANT_NAME is not used anywhere else.

atodorov:
NOTE: I'd like to merge this after #580 to validate the error handling code first!

PR #583, opened 2019-01-25
New CLI test for building compose with SELinux in enforcing mode

jikortus:
New test to check that composer can work with SELinux in enforcing mode on the host system.

### DIFF
--- a/tests/cli/test_compose_ext4-filesystem.sh
+++ b/tests/cli/test_compose_ext4-filesystem.sh
@@ -16,6 +16,7 @@ CLI="./src/bin/composer-cli"
 
 rlJournalStart
     rlPhaseStartTest "compose start"
+        rlAssertEquals "SELinux operates in enforcing mode" "$(getenforce)" "Enforcing"
         UUID=`$CLI compose start example-http-server ext4-filesystem`
         rlAssertEquals "exit code should be zero" $? 0
 
--- a/tests/cli/test_build_and_deploy_aws.sh
+++ b/tests/cli/test_build_and_deploy_aws.sh
@@ -71,6 +71,7 @@ __EOF__
     rlPhaseEnd
 
     rlPhaseStartTest "compose start"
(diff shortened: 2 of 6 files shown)


atodorov:
Note: for rhel7-extras we need #525 to be merged

jikortus:
OK, so I just used a check for selinux mode instead of a separate test to save time and resources.

I wonder if there could be any downsides with putting a selinux mode check in the test execution loop in `tests/test_cli.sh`? At least it wouldn't be necessary to make sure the check is present in every (new) test created, so this might be a more consistent approach.

tests/cli/test_compose_ext4-filesystem.sh
```diff
@@ -15,6 +15,10 @@ CLI="./src/bin/composer-cli"
 
 
 rlJournalStart
+    rlPhaseStart "SELinux operates in enforcing mode"
+        rlAssertEquals "$(getenforce)" "Enforcing"
```
atodorov: `rlAssertEquals "SELinux operates in enforcing mode" "$(getenforce)" "Enforcing"`, see:
https://github.com/beakerlib/beakerlib/wiki/man#rlassertequals

and you can place the assert right after "compose start" opening clause. No need for separate phase only for that.

atodorov:
> I wonder if there could be any downsides with putting a selinux mode check in the test execution loop in tests/test_cli.sh? At least it wouldn't be necessary to make sure the check is present in every (new) test created, so this might be a more consistent approach.

That would be the easiest thing to do but it skrews up the nice results formatting that we get from beakerlib. IMO having easy to understand failure messages will be more important in the future b/c we won't have the time luxury to figure out what went wrong when it does.

tests/cli/test_compose_ext4-filesystem.sh
jikortus: Thank you for pointing this out. Fixed.

atodorov:
Jenkins failures reported in https://bugzilla.redhat.com/show_bug.cgi?id=1672971, not related to current changes.

PR #587, opened 2019-02-01
installer: Run anaconda in a mount and pid namespace

larskarlitski:
Anaconda can leave child processes and mounts around when it crashes or
is canceled before finishing. It also sometimes unmounts unrelated file
systems (https://github.com/rhinstaller/anaconda/issues/1791).

Run it in a mount and pid namespace to clean up after it.

### DIFF
--- a/src/pylorax/installer.py
+++ b/src/pylorax/installer.py
@@ -382,7 +382,8 @@ def novirt_install(opts, disk_img, disk_size, cancel_func=None):
     # Make sure anaconda has the right product and release
     log.info("Running anaconda.")
     try:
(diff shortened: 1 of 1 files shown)


bcl:
I finally got around to testing this with mock, and we need to drop the --mount option. koji uses mock --old-chroot to run livemedia-creator so it needs to be able to work there as well. When run from inside mock with --mount it complains that it cannot change the root propagation.

larskarlitski:
I've added `--propagation unchanged`, which makes `unshare` not mark the `/` mount as private. This alleviates many of the effects of being in a mount namespace, but is still better than nothing.

I've verified that this works in `mock`.

PR #589, opened 2019-02-04
WIP: Add script for removing old artifacts from Azure

jstodola:


jstodola:
This script is not complete yet, please do not merge!
I will have to extend the script not to remove all old resources, since not all of them should be removed.

tests/cleanup/remove_old_objects_azure.sh
```diff
@@ -0,0 +1,133 @@
...
+    rlPhaseEnd
+
+    # A list of Azure resources we want to clean
+    resource_types="
+Microsoft.Compute/virtualMachines
+Microsoft.Network/networkInterfaces
+Microsoft.Network/publicIPAddresses
+Microsoft.Network/networkSecurityGroups
+Microsoft.Compute/disks
+Microsoft.Compute/images
```
atodorov: will this automatically remove what they call blobs ? I guess so but you can never be too sure when things are named differently.

tests/cleanup/remove_old_objects_azure.sh
```diff
@@ -0,0 +1,133 @@
...
+Microsoft.Storage/storageAccounts
```
atodorov: Don't remove storage accounts. This already caused existing Azure tests to fail!

jstodola: A new storage account is created automatically when a new VM is created by the testing script test_build_and_deploy_azure.sh, so here are some possible solutions:
1) modify test_build_and_deploy_azure.sh to use existing storage account (composerredhat) instead of creating new one for each VM. I prefer this option.
2) do not delete storage accounts at all
3) delete storage accounts except of composerredhat

atodorov: +1 for 1. and 2.

tests/cleanup/remove_old_objects_azure.sh
```diff
@@ -0,0 +1,133 @@
...
+# Script removes virtual machines and other artifacts older than HOURS_LIMIT (24 hours by default) from Azure
+
+. /usr/share/beakerlib/beakerlib.sh
+
+
+# Delete old objects based on the $TAG_NAME tag value defined in a previous execution of the script
+delete_old_resources() {
+    local resource_type="$1"
+
+    # list resources older than $TIMESTAMP based on the $TAG_NAME tag created in a previous run
```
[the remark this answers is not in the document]
atodorov: Doesn't that mean the script will tag new resources today and delete them tomorrow, essentially leaving them hanging 2× the defined time ?

Current approach is good (manually created stuff) but adding the tagging when the resources are consumed from the test script will be even better, more optimal.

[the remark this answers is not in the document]
jstodola: Since the cleaning script is executed every hour, the new resource will be removed in ~25 hours.
Extending the test script to add tags for VMs should not be difficult, but tagging other resources (ip addresses, network security groups,...) would require an extra effort to implement.

jstodola:
New version:
1) Do not remove storage accounts
2) Remove old blobs

Related PR: https://github.com/weldr/lorax/pull/590 to record creation date/time of a VM and to use existing storage account for new VMs.

tests/cleanup/remove_old_objects_azure.sh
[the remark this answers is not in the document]
atodorov: VM tagging by the test is fine. The biggest cost on Azure is compute time and then storage. When we had the spike in cost the stats show network badwidth but I can't find more info about that in Azure's reports. Go figure.

PR #593, opened 2019-02-06
Azure: Use existing storage account and record date/time of VM creation

jstodola:


atodorov:
@jstodola next time please leave the PR template text. It serves as a reminder of which tests need to pass before we can merge.

PR #599, opened 2019-02-07
Drop auth from the kickstart examples

bcl:
System defaults should be sufficient these days.

Resolves: rhbz#1672583


bcl:
Pushed to master in commit https://github.com/weldr/lorax/commit/2af0973155425e3e532b5971638a7c175ce89252

PR #625, opened 2019-02-27
New test: Build live-iso and boot with KVM

atodorov:
--- Description of proposed changes ---

explicitly enables sshd for live-iso during testing

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-openstack-runtest` PASS
- [ ] `*-vmware-runtest` PASS
- [ ] For `rhel8-*` and `rhel7-*` branches commit log references an approved
  bug in Bugzilla. Do not merge if the bug doesn't have the 3 ACKs set to `+`!

--- Jenkins commands ---

- `ok to test` to accept this pull request for testing
- `test this please` for a one time test run
- `retest this please` to start a new build

PR #627, opened 2019-02-28
Add checks for disabled root account

jikortus:
The root account checks are applied to generated and deployed images, except
for live ISO ATM (will be addressed subsequently).

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---

### DIFF
--- a/tests/cli/lib/root_account.sh
+++ b/tests/cli/lib/root_account.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+check_root_account() {
+# Try to SSH to a remote machine first using root account using password-based
+# auth (this is expected to fail) and then using key-based auth with the
+# supplied username to check content of /etc/shadow and audit.log.
+#
+# use: check_root_account <user> <machine> [ssh options]
+
+    local ssh_opts="-o StrictHostKeyChecking=no $3"
+    local user="$1"
+    local machine="$2"
+    if [[ "$user" == "" || "$machine" == "" ]]; then
+        rlFail "check_root_account: Missing user or machine parameter."
+        return 1
+    fi
+
+    if [ $ROOT_ACCOUNT_LOCKED == 0 ]; then
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep '^root::' /etc/shadow\"" \
+            0 "Password for root account in /etc/shadow is empty"
+    else
+        # ssh returns 255 in case of any ssh error, so it's better to grep the specific error message
+        rlRun -t -c "ssh $ssh_opts -o PubkeyAuthentication=no root@${machine} 2>&1 | grep -i 'permission denied ('" \
+            0 "Can't ssh to '$machine' as root using password-based auth"
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep -E '^root:(\*LOCK\*|!):' /etc/shadow\"" \
+            0 "root account is disabled in /etc/shadow"
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep 'USER_LOGIN.*acct=\\\"root\\\".*terminal=ssh.*res=failed' /var/log/audit/audit.log\"" \
+            0 "audit.log contains entry about unsuccessful root login"
+        # We modify the default sshd settings on live ISO, so we can only check the default empty password setting
+        # outside of live ISO
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} 'grep -E \"^[[:blank:]]*PermitEmptyPasswords[[:blank:]]*yes\" /etc/ssh/sshd_config'" 1 \
+            "Login with empty passwords is disabled in sshd config file"
+    fi
+    rlRun -t -c "ssh $ssh_opts ${user}@${machine} 'cat /etc/redhat-release'"
+
+}
+
--- a/tests/cli/test_compose_live-iso.sh
+++ b/tests/cli/test_compose_live-iso.sh
@@ -51,7 +51,8 @@ rlJournalStart
 
     rlPhaseStartTest "Verify VM instance"
         # verify we can login into that instance *WITHOUT* a password
-        rlRun -t -c "ssh -oStrictHostKeyChecking=no -p 2222 root@localhost 'cat /etc/redhat-release'"
+        . ./tests/cli/lib/root_account.sh
+        ROOT_ACCOUNT_LOCKED=0 check_root_account root localhost "-p 2222"
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/test_cli.sh
+++ b/tests/test_cli.sh
@@ -8,6 +8,13 @@ function setup_tests {
     # explicitly enable sshd for live-iso b/c it is disabled by default
     # due to security concerns (no root password required)
     sed -i.orig 's/^services.*/services --disabled="network" --enabled="NetworkManager,sshd"/' $1/composer/live-iso.ks
+    # explicitly enable logging in with empty passwords via ssh, because
+    # the default sshd setting for PermitEmptyPasswords is 'no'
+    awk -i inplace "
+        /%post/ && FLAG != 2 {FLAG=1}
+        /%end/ && FLAG == 1 {print \"sed -i 's/.*PermitEmptyPasswords.*/PermitEmptyPasswords yes/' /etc/ssh/sshd_config\"; FLAG=2}
+        {print}" \
+        $1/composer/live-iso.ks
 }
 
 function teardown_tests {
--- a/tests/cli/test_build_and_deploy_aws.sh
+++ b/tests/cli/test_build_and_deploy_aws.sh
@@ -185,8 +185,9 @@ __EOF__
             CLOUD_USER="fedora"
         fi
 
-        # verify we can login into that instance and maybe some other details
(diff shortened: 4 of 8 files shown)


tests/cli/lib/root_account.sh
```diff
@@ -0,0 +1,26 @@
...
+# Try to SSH to a remote machine first using root account using password-based
+# auth (this is expected to fail) and then using key-based auth with the
+# supplied username to check content of /etc/shadow and audit.log.
+#
+# use: check_root_account <user> <machine> [ssh options]
+
+    local ssh_opts="-o StrictHostKeyChecking=no"
+    local user="$1"
+    local machine="$2"
+    ssh_opts="$ssh_opts $3"
```
atodorov: or you can just hard code the options from above here.

tests/cli/lib/root_account.sh
```diff
@@ -0,0 +1,26 @@
...
+    if [[ "$user" == "" || "$machine" == "" ]]; then
+        rlLogError "check_root_account: Missing user or machine parameter."
+        return 1
```
atodorov: not sure how `return 1` plays in the entire scheme but `rlFail "message"` instead of rlLogError should be better. Need to experiment with it though.

tests/cli/lib/root_account.sh
```diff
@@ -0,0 +1,26 @@
...
+    fi
+
+    # ssh returns 255 in case of any ssh error, so it's better to grep the specific error message
+    rlRun -t -c "ssh $ssh_opts -o PubkeyAuthentication=no root@${machine} 2>&1 | grep -i 'permission denied ('" \
+        0 "Can't SSH to '$machine' as root using password-based auth"
+    rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep -E '^root:(\*LOCK\*|!):' /etc/shadow\"" \
+        0 "root account is disabled in /etc/shadow"
+    rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep 'USER_LOGIN.*acct=\\\"root\\\".*terminal=ssh.*res=failed' /var/log/audit/audit.log\"" \
+        0 "sshd log contains entry about unsuccessful root login"
```
atodorov: Add `cat /etc/redhat-release` at the end, just because this is our historical "Check marker". In the future others may follow.

tests/cli/lib/root_account.sh
jikortus: You're right that `rlFail` is a better solution, as `rlLogError` just logs the error, but doesn't affect the test result, which might cause silent failures in this case. The `return 1` by itself doesn't affect the test results in any way, it's just meant to skip the further steps from being executed as they would fail anyway.

jikortus:
> I am not entirely sure what to do with tar images since we use podman/systemd-nspawn there. Maybe try testing without SSH ? I don't have ideas ATM.
I have personally no idea as well as I don't have any experience in this area yet. We might just inspect the archive, or I can investigate further some more advanced solution if you think it would be more appropriate.

> NOTE: #625 also needs this depending on which one gets merged first!
OK, I'm aware of that. I will add a check for the live ISO as well once the PR is merged.

atodorov:
@jikortus please rebase and add changes for live-iso.

jikortus:
@atodorov The checks for live ISO were added, please have a look at the changes.

tests/test_cli.sh
```diff
@@ -8,6 +8,12 @@ function setup_tests {
     # explicitly enable sshd for live-iso b/c it is disabled by default
     # due to security concerns (no root password required)
     sed -i.orig 's/^services.*/services --disabled="network" --enabled="NetworkManager,sshd"/' $1/composer/live-iso.ks
+    # explicitly enable logging in with empty passwords via ssh
+    awk -i inplace "
+        /%post/ && FLAG != 2 {FLAG=1}
+        /%end/ && FLAG == 1 {print \"sed -i 's/.*PermitEmptyPasswords.*/PermitEmptyPasswords yes/' /etc/ssh/sshd_config\"; FLAG=2}
+        {print}" \
+        $1/composer/live-iso.ks
```
atodorov: Why do we need to explicitly enable this? IIRC it is possible to login into the live-iso VM without a password and that is why ssh is disabled by default. And also the existing live-iso tests seem to be working without this change so what causes the need for this ?

jikortus: We need to use the explicit `sshd` setting for `PermitEmptyPasswords` because it is disabled by default for security reasons and this option is not set (is commented out) in `sshd_config`. I also updated the comment before the respective `awk` code to be a bit more descriptive.
The default setting is stated in `man sshd_config` and is consistent with my observations when I tried to login to a machine with empty root password and default `sshd` settings. So I wonder what could be different in my case if your experience is different?

tests/cli/lib/root_account.sh
```diff
@@ -0,0 +1,45 @@
...
+        0 "audit.log contains entry about unsuccessful root login"
+    rlRun -t -c "ssh $ssh_opts ${user}@${machine} 'cat /etc/redhat-release'"
+}
+
+check_root_account_live_iso() {
+# Inspect state of root account in live ISO instance
+#
+# use: check_root_account_live_iso <user> <machine> [ssh options]
+
+    local ssh_opts="-o StrictHostKeyChecking=no -o PubkeyAuthentication=no $3"
```
atodorov: `-o PubkeyAuthentication=no` is going to fail for live-iso b/c it can only support key based authentication

tests/test_cli.sh
atodorov: This snippet needs to be removed, we are not enabling ssh login with empty passwords during testing.

tests/cli/test_compose_live-iso.sh
```diff
@@ -51,7 +51,8 @@ rlJournalStart
 
     rlPhaseStartTest "Verify VM instance"
         # verify we can login into that instance *WITHOUT* a password
-        rlRun -t -c "ssh -oStrictHostKeyChecking=no -p 2222 root@localhost 'cat /etc/redhat-release'"
+        . ./tests/cli/lib/root_account.sh
+        check_root_account_live_iso root localhost "-p 2222 -i $SSH_KEY_DIR/id_rsa"
```
atodorov: I've tried this PR manually with the changes for enabling passwordless logins. You have to remove `i $SSH_KEY_DIR/id_rsa` from here. Very likely `-o PubkeyAuthentication=no` as well and we should be good to go.

jikortus: I updated the code, now when I see your comment again, I'm not sure what exactly you mean by 'changes for enabling passwordless logins'. Are those the changes done in sshd config that we agreed upon should be removed, or you added keys to the live iso when you tried the PR? At this stage the changes don't work for me, I can't think of any other way for the ssh login to work other than changing sshd config or adding the public key in the composed iso.

tests/cli/lib/root_account.sh
```diff
@@ -0,0 +1,35 @@
...
+        # ssh returns 255 in case of any ssh error, so it's better to grep the specific error message
+        rlRun -t -c "ssh $ssh_opts -o PubkeyAuthentication=no root@${machine} 2>&1 | grep -i 'permission denied ('" \
+            0 "Can't ssh to '$machine' as root using password-based auth"
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep -E '^root:(\*LOCK\*|!):' /etc/shadow\"" \
+            0 "root account is disabled in /etc/shadow"
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep 'USER_LOGIN.*acct=\\\"root\\\".*terminal=ssh.*res=failed' /var/log/audit/audit.log\"" \
+            0 "audit.log contains entry about unsuccessful root login"
+    fi
+    rlRun -t -c "ssh $ssh_opts ${user}@${machine} 'grep -E \"^[[:blank:]]*PermitEmptyPasswords[[:blank:]]*yes\" /etc/ssh/sshd_config'" 1 \
+        "Login with empty passwords is disabled in sshd config file"
```
atodorov: shouldn't this line be inside the else block ?

jikortus: No, because this option is not explicitly enabled in the sshd configuration by default in general (I think this is correct due to security reasons, and that default configuration is the reason I think the ssh login hasn't worked for me).

tests/cli/lib/root_account.sh
```diff
@@ -0,0 +1,35 @@
...
+    local user="$1"
+    local machine="$2"
+    if [[ "$user" == "" || "$machine" == "" ]]; then
+        rlFail "check_root_account: Missing user or machine parameter."
+        return 1
+    fi
+
+    if [ $ROOT_ACCOUNT_LOCKED == 0 ]; then
+        rlRun -t -c "ssh $ssh_opts ${user}@${machine} \"sudo grep '^root::' /etc/shadow\"" \
+            0 "Password for root account in /etc/shadow is empty"
```
atodorov: I was wrong before. This will not work if passwordless login isn't explicitly enabled with the awk code snippet which you had before. ATM I get stuck at a password prompt:
```
STDERR: Warning: Permanently added '[localhost]:2222' (ECDSA) to the list of known hosts.
root@localhost's password: 
```

jikortus: Code updated, should be fully functional now.

PR #666, opened 2019-03-28
Pass ssl certificate options to anaconda

larskarlitski:


bcl:
Thanks, adjusted locally for the pykickstart version and bz#

PR #668, opened 2019-03-29
Google cloud

dashea:
Add support for google cloud images.

The "google" compose type will create a raw partitioned disk image named "disk.raw", aligned to a whole number of GiB, with the configuration described in https://cloud.google.com/compute/docs/images/configuring-imported-images and the bootloader options from 
https://cloud.google.com/compute/docs/images/import-existing-image#prepare_boot_disk. The resulting disk image is compressed in a .tar.gz file.

Some points to think about:
  * This image will still have `rhgb quiet` in the bootloader config, even though google says not to include those. There is not currently a way to remove these options except through a %post script.
  * I also don't know if `rhgb quiet` matters or not since I haven't gotten anything to boot yet due to issues (probably) unrelated to the actual image.
  * The ssh configuration might be overkill
  * Not real sure how we want to test this

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---


dashea:
Fixed the list for the `compose types` test and add a test for making a tarball of a single file to test_imgutils.py.

I also changed the options I had added for make_tar_disk. I had originally changed opts.image_name to be the disk image to be wrapped in a tarball (i.e., the file that anaconda writes out), but I found a couple other places where I missed adding a conditional and decided that it was bad idea. opts.image_name is now the final output file for all compose types, and the new option, tar_disk_name, is the filename for the intermediate image.

Also added a script to test compose google cloud images. It just makes sure that the result is a tarball that contains "disk.raw"

dashea:
> Also added a script to test compose google cloud images. It just makes sure that the result is a tarball that contains "disk.raw"

One more commit for this one, I forgot to chmod +x.

PR #671, opened 2019-04-02
Add a new compose type for Hyper-V

dashea:
This is based on the VHD compose type, with the following differences:

  * Use the vhdx format instead of vhd
  * No WALinuxAgent
  * Install hyperv-daemons

The hyperv-daemons are activated through udev rules, so there is no need
to add them to the services line.

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---


src/pylorax/api/compose.py
```diff
@@ -848,6 +848,28 @@ def compose_args(compose_type):
                                  "app_template":            None,
                                  "app_file":                None,
                                 },
+            "hyper-v":          {"make_iso":                False,
```
bcl: This section needs to have all the new fields we've added, like tar_disk_name, etc.

dashea: oops. Fixed.

src/pylorax/api/compose.py
```diff
@@ -848,6 +848,31 @@ def compose_args(compose_type):
...
+                                 "make_disk":               True,
+                                 "make_fsimage":            False,
+                                 "make_appliance":          False,
+                                 "make_ami":                False,
+                                 "make_tar":                False,
+                                 "make_tar_disk":           True,
```
bcl: Should be false :)

dashea: wow, was I braindead yesterday or what. Fixed.

PR #675, opened 2019-04-04
Changes needed for running tests against installed package

bcl:
--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---


bcl:
We can't install with dnf because the packages aren't (yet) available. So we have to use pip (you'll see that part when I do the osci PR to do a full-up test of this). We already duplicate .spec packages in the Dockerfile, and this is the least horrible solution (other possibility is to encode it into tests.yml).

The goal here is to have one codebase that will pass the tests when run with the F28 container, and will also pass when running on a RHEL8 system with the packages installed instead of the source.

atodorov:
@bcl in this case I guess we are safe to merge.

PR #684, opened 2019-04-09
Add test for passing custom option on kernel command line

jikortus:
Related: rhbz#1687595

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---


tests/cli/test_build_and_deploy_vmware.sh
```diff
@@ -81,6 +81,9 @@ version = "*"
 [[customizations.user]]
 name = "root"
 key = "$PUB_KEY"
+
+[customizations.kernel]
```
atodorov: Is this `[customizations.kernel]` or `[[customizations.kernel]]`, see above ?

tests/cli/test_build_and_deploy_aws.sh
```diff
@@ -188,6 +188,10 @@ __EOF__
         # verify we can login into that instance and root account is disabled
         . ./tests/cli/lib/root_account.sh
         check_root_account $CLOUD_USER $IP_ADDRESS "-i $SSH_KEY_DIR/id_rsa"
+
+        # verify that a kernel command line argument was passed from the blueprint (this is added to the blueprint in ../test_cli.sh)
+        rlRun -t -c "ssh -o StrictHostKeyChecking=no -i $SSH_KEY_DIR/id_rsa $CLOUD_USER@$IP_ADDRESS \"grep custom_cmdline_arg /proc/cmdline\"" 0 \
+            "System booted from the image contains specified parameter on kernel command line"
```
atodorov: I am wondering if you can put this into `check_root_account` and not modify every single file. And probably rename `check_root_account` so it is less confusing ?

tests/cli/test_build_and_deploy_vmware.sh
bcl: It is not a list, so it is [customizations.kernel]

The [[foo]] form is for things that can be listed multiple times, like users.

tests/cli/test_build_and_deploy_aws.sh
jikortus: I can see your point, but I'm not sure if moving this to `check_root_account` and renaming it would be the right approach as IMHO this check isn't anyhow related to what `check_root_account` does ATM and this way it would/could become just some 'generic checking function'.

I also considered creating a separate function for the new check, but in the end didn't think it's worth going this direction, but perhaps it could be a feasible compromise to what you suggested as it would at least save some repetitions (part of ssh parameters + the rlRun message)?

atodorov: >  'generic checking function'.

this is my idea. The benefits I see is minimizing the code which needs to be shared between the bash scripts and having most(all) checks executed on all image types. This is at the expense of making this function grow over time, possibly making it sort of complicated.

OTOH the new generic function can serve as an entry point and it will (over time) just call other functions which will perform the actual checks. What could go wrong (tm) ?

jikortus: I incorporated the changes we talked about on IRC - created a generic check function that in turn calls the already existing root account checking function, plus a new kernel command line checking function. Hopefully this will make more sense now and will save unnecessary typing for the future 'generic' tests.

jikortus:
Sorry for that - it should be OK now.

PR #686, opened 2019-04-10
Backport additional tests for blueprints/compose

atodorov:

Referenced bug still not approved.

--- Merge policy ---

- [x] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---

### DIFF
--- a/tests/cli/lib/toml-compare
+++ b/tests/cli/lib/toml-compare
@@ -1,0 +1,20 @@
+#!/usr/bin/env python
+"""
+    Will compare 2 blueprints for equality!
+    On RHEL7/Python 2 toml uses dictionaries and
+    blueprint show doesn't keep the order of the elements.
+
+    For master/Python3 this is not the case.
+"""
+
+import sys
+import pytoml
+
+if len(sys.argv) != 3:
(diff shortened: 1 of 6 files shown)


atodorov:
@bcl referenced bug has been approved, PR has been updated to what is in master. Please review again.

bcl:
I don't think there's anything else to do for this. But the order of groups/modules looks like it makes the test fail. It should be:
    groups = []
    modules = []

atodorov:
> ... the order of groups/modules looks like it makes the test fail. It should be:
> groups = []
> modules = []

@bcl checkout my latest commit here. Apparently `blueprints show` in RHEL7 scrambles the contents, which is not what we see on master and rhel8. If you think the latest commit is acceptable workaround I will squash it into the existing ones. Otherwise we probably need to treat this as a bug.

atodorov:
@bcl I worked around the fact that TOML dicts on RHEL7 don't keep their ordering by adding a small script to compare 2 blueprints. Please review and merge if you are happy with the changes.

PR #689, opened 2019-04-12
Use Cockpit's test images and infrastructure

larskarlitski:
Allows to run the tests on multiple operating systems and on the
infrastructure that the Cockpit team maintains.

`make vm` downloads one of Cockpit's test images (override which one
with TEST_OS) and installs a local build of lorax rpms. The resulting
image is placed in `test/images/$TEST_OS`.

TEST_OS can be set to any of Cockpit's supported images (default:
fedora-30).

Run `make check-vm` to run the CLI checks in the VM. The bulk of the
work is done in `test/check-cli`, which uses Cockpit's `bots` library to
start the VM and run the script in it.

Also included is a `test/run` script, which is the entrypoint for
Cockpit's test infrastructure.

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,12 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 
 IMAGE_RELEASE = $(shell awk -F: '/FROM/ { print $$2}' Dockerfile.test)
 
+ifeq ($(TEST_OS),)
+TEST_OS = fedora-30
+endif
+export TEST_OS
+VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
+
 default: all
 
 src/composer/version.py: lorax.spec
@@ -95,11 +101,17 @@ set-docs-owner:
 
 archive:
 	@git archive --format=tar --prefix=$(PKGNAME)-$(VERSION)/ $(TAG) > $(PKGNAME)-$(VERSION).tar
-	@gzip $(PKGNAME)-$(VERSION).tar
+	@gzip -f $(PKGNAME)-$(VERSION).tar
 	@echo "The archive is in $(PKGNAME)-$(VERSION).tar.gz"
 
 dist: tag archive
 	scp $(PKGNAME)-$(VERSION).tar.gz fedorahosted.org:lorax
+
+srpm: archive $(PKGNAME).spec
+	rpmbuild -bs \
+	  --define "_sourcedir $(CURDIR)" \
+	  --define "_srcrpmdir $(CURDIR)" \
+	  lorax.spec
 
 local:
 	@rm -rf $(PKGNAME)-$(VERSION).tar.gz
@@ -129,8 +141,34 @@ docs-in-docker:
 
 ci: check test
 
+$(VM_IMAGE): TAG=HEAD
+$(VM_IMAGE): srpm bots
+	srpm=$(shell rpm --qf '%{Name}-%{Version}-%{Release}.src.rpm\n' -q --specfile lorax.spec | head -n1) ; \
+	bots/image-customize -v \
+		--resize 20G \
+		--upload $$srpm:/var/tmp \
+		--upload $(CURDIR)/test/vm.install:/var/tmp/vm.install \
+		--upload $(realpath tests):/ \
+		--run-command "chmod +x /var/tmp/vm.install" \
+		--run-command "cd /var/tmp; /var/tmp/vm.install $$srpm" \
+		$(TEST_OS)
+
+# convenience target for the above
+vm: $(VM_IMAGE)
+	echo $(VM_IMAGE)
+
+vm-reset:
+	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
+
+# checkout Cockpit's bots/ directory for standard test VM images and API to launch them
+# must be from cockpit's master, as only that has current and existing images; but testvm.py API is stable
+bots:
+	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git
+	git checkout --force FETCH_HEAD -- bots/
+	git reset bots
+
 .PHONY: ci_after_success
 ci_after_success:
 # nothing to do here, but Jenkins expects this to be present, otherwise fails
 
-.PHONY: docs
+.PHONY: docs check test srpm vm vm-reset
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,57 @@
+# Integration Tests
+
+lorax uses Cockpit's integration test framework and infrastructure. To do this,
+we're checking out Cockpit's `bots/` subdirectory. It contains links to test
+images and tools to manipulate and start virtual machines from them.
+
+Each test is run on a new instance of a virtual machine. 
+
+## Dependencies
+
+These dependencies are needed on Fedora to run tests locally:
+
+    $ sudo dnf install curl expect \
+        libvirt libvirt-client libvirt-daemon libvirt-python \
+        python python-libguestfs python-lxml libguestfs-xfs \
+        python3 libvirt-python3 \
+        libguestfs-tools qemu qemu-kvm rpm-build rsync xz
+
+## Building a test VM
+
+To build a test VM, run
+
+    $ make vm
+
+This downloads a base image from Cockpit's infrastructure. You can control
+which image is downloaded with the `TEST_OS` environment variable. Cockpit's
+[documentation](https://github.com/cockpit-project/cockpit/blob/master/test/README.md#test-configuration)
+lists accepted values. It then creates a new image based on that (a qemu
+snapshot) in `tests/images`, which contain the current `tests/` directory and
+have newly built rpms from the current checkout installed.
+
+To delete the generated image, run
+
+    $ make vm-reset
+
+Base images are stored in `bots/images`. Set `TEST_DATA` to override this
+directory.
+
+## Running tests
+
+After building a test image, run
+
+    $ ./test/check-cli [TESTNAME]
+
+or any of the other `check-*` scripts. Right after the VM is started, these
+scripts print an `ssh` line to connect to it.
+
+Run `make vm` after changing tests or lorax source to recreate the test
+machine. It is usually not necessary to reset the VM.
+
+## Updating images
+
+The `bots/` directory is checked out from Cockpit when `make vm` is first run.
+To get the latest images you need to update it manually (in order not to poll
+GitHub every time):
+
+    $ make -B bots
--- a/test/check-cli
+++ b/test/check-cli
@@ -1,0 +1,32 @@
+#!/usr/bin/python3
+
+import composertest
+
+
+class TestSanity(composertest.ComposerTestCase):
+    def test_blueprint_sanity(self):
+        self.runCliTest("/tests/cli/test_blueprints_sanity.sh")
+
+    def test_compose_sanity(self):
+        self.runCliTest("/tests/cli/test_compose_sanity.sh")
+
+
+class TestImages(composertest.ComposerTestCase):
+    def test_ext4_filesystem(self):
+        self.runCliTest("/tests/cli/test_compose_ext4-filesystem.sh")
+
+    def test_partitioned_disk(self):
+        self.runCliTest("/tests/cli/test_compose_partitioned-disk.sh")
+
+    def test_tar(self):
+        self.runCliTest("/tests/cli/test_compose_tar.sh")
+
+    def test_qcow2(self):
+        self.runCliTest("/tests/cli/test_compose_qcow2.sh")
+
+    def test_live_iso(self):
+        self.runCliTest("/tests/cli/test_compose_live-iso.sh")
+
+
+if __name__ == '__main__':
+    composertest.main()
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -1,0 +1,59 @@
+#!/usr/bin/python3
+
+import os
+import subprocess
+import sys
+import unittest
+
+# import Cockpit's machinery for test VMs and its browser test API
+sys.path.append(os.path.join(os.path.dirname(__file__), "../bots/machine"))
+import testvm # pylint: disable=import-error
+
+
+class ComposerTestCase(unittest.TestCase):
+    image = testvm.DEFAULT_IMAGE
+
+    def setUp(self):
+        network = testvm.VirtNetwork(0)
+        self.machine = testvm.VirtMachine(self.image, networking=network.host(), memory_mb=2048)
+
+        print(f"Starting virtual machine '{self.image}'")
+        self.machine.start()
+        self.machine.wait_boot()
+
+        # run a command to force starting the SSH master
+        self.machine.execute("uptime")
+
+        self.ssh_command = ["ssh", "-o", "ControlPath=" + self.machine.ssh_master,
+                                   "-p", self.machine.ssh_port,
+                                   self.machine.ssh_user + "@" + self.machine.ssh_address]
+
+        print("Machine is up. Connect to it via:")
+        print(" ".join(self.ssh_command))
+        print()
+
+        print("Waiting for lorax-composer to become ready...")
+        curl_command = ["curl", "--max-time", "360",
+                                "--silent",
+                                "--unix-socket", "/run/weldr/api.socket",
+                                "http://localhost/api/status"]
+        r = subprocess.run(self.ssh_command + curl_command, stdout=subprocess.DEVNULL)
+        self.assertEqual(r.returncode, 0)
+
+    def tearDown(self):
+        self.machine.stop()
+
+    def execute(self, command, **args):
+        """Execute a command on the test machine.
+
+        **args and return value are the same as those for subprocess.run().
+        """
+        return subprocess.run(self.ssh_command + command, **args)
+
+    def runCliTest(self, script):
+        r = self.execute(["CLI=/usr/bin/composer-cli", "TEST=" + self.id(), "/tests/test_cli.sh", script])
+        self.assertEqual(r.returncode, 0)
+
+
+def main():
+    unittest.main(verbosity=2)
--- a/test/run
+++ b/test/run
@@ -1,0 +1,11 @@
+#!/bin/sh -e
+# This is the expected entry point for Cockpit CI; will be called without
+# arguments but with an appropriate $TEST_OS, and optionally $TEST_SCENARIO
+
+make vm
+
+if [ -n "$TEST_SCENARIO" ]; then
+  test/check-cloud TestCloud.test_$TEST_SCENARIO
+else
+  test/check-cli
+fi
--- a/test/vm.install
+++ b/test/vm.install
@@ -1,0 +1,18 @@
+#!/bin/sh -eux
+
+SRPM="$1"
+
+# Grow root partition to make room for images. This only works on Fedora right now.
+echo ", +" | sfdisk -N 2 -f /dev/vda
+partprobe
+pvresize /dev/vda2
+lvresize fedora/root -l+100%FREE -r
+
+su builder -c "/usr/bin/mock --no-clean --resultdir build-results --rebuild $SRPM"
+
+packages=$(find build-results -name '*.rpm' -not -name '*.src.rpm')
+rpm -e $(basename -a ${packages[@]} | sed 's/-[0-9].*.rpm$//') || true
+yum install -y beakerlib $packages
+
+systemctl enable lorax-composer.socket
+systemctl enable docker.service
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
+*.rpm
 src/pylorax/version.py*
 _build/
 tests/pylint/.pylint.d/
@@ -7,3 +8,7 @@ __pycache__/
 pylint-log
 .pytest_cache/
 .test-results/
+/lorax-*.tar.gz
+/bots
+/test/images
+/tmp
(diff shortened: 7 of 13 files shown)


test/check-cli
```diff
@@ -0,0 +1,44 @@
...
+
+    print("ssh -o ControlPath=%s -p %s %s@%s" %
+          (machine.ssh_master, machine.ssh_port, machine.ssh_user, machine.ssh_address))
+
+    # run a command to force starting the SSH master
+    machine.execute('uptime')
+
+    r = subprocess.run(["ssh", "-o", "ControlPath=" + machine.ssh_master,
+                               "-p", machine.ssh_port,
+                               machine.ssh_user + "@" + machine.ssh_address,
```
martinpitt: I suggest to assemble the ssh argv prefix above,  and re-use it both here and in the print (but I suppose the print is also just an intermediate debugging thing)

larskarlitski: Indeed, I had used it for debugging. It might be helpful in the future though. I've moved the common prefix out and printed a message with it. Thanks!

Makefile
```diff
@@ -12,6 +12,12 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 
 IMAGE_RELEASE = $(shell awk -F: '/FROM/ { print $$2}' Dockerfile.test)
 
+ifeq ($(TEST_OS),)
+TEST_OS = fedora-30
```
atodorov: Does the CI infrastructure defined TEST_OS for the various RHEL builds? 

Is it possible we have access to the latest RHEL snapshots/content repositories ?

test/vm.install
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -eu
+
+rpm -e lorax lorax-templates-generic lorax-lmc-virt lorax-lmc-novirt lorax-composer composer-cli || true
```
atodorov: For composer we'll need the ability to install an arbitrary list of RPMs (URLs to RPMs more precisely). This will facilitate testing pre-built packages from Brew/RHEL snapshots and also allow upgrading/installing other packages as necessary - for example anaconda RPMs to test changes in anaconda before they are merged. 

At the moment Jenkins supports the `EXTRA_RPMS` variable which contains all of these URLs and the system provisioning takes care to download them and install them onto the system before testing.

test/check-cli
```diff
@@ -0,0 +1,45 @@
...
+
+    ssh_command = ["ssh", "-o", "ControlPath=" + machine.ssh_master,
+                          "-p", machine.ssh_port,
+                          machine.ssh_user + "@" + machine.ssh_address]
+
+    print(f"Virtual machine running {image} is up. Connect to it via:")
+    print(" ".join(ssh_command))
+    print()
+
+    r = subprocess.run(ssh_command + ["CLI=/usr/bin/composer-cli", "/tests/test_cli.sh"])
```
atodorov: or maybe `make test_images test_aws test_azure`, etc.

All of these require credentials for the various cloud providers. Is it possible to securely store these credentials in the new CI infrastructure? 

More importantly quickly updating the credentials (also via ENV/manually) when they change or when we want to execute tests with a different account ?

Makefile
larskarlitski: Yes. We're installing and subscribing those VMs in [rhel.setup]( https://github.com/cockpit-project/cockpit/blob/master/bots/images/scripts/rhel.setup). Cockpit only subscribes them during installation. For Composer, we probably want to resubscribe them to test pulling content from those repos. 

You can see all available images in Cockpit's [image-refresh](https://github.com/cockpit-project/cockpit/blob/master/bots/image-refresh#L28) script.

test/vm.install
larskarlitski: `vm.install` installs the rpms that were just built in the source directory. We could add something like `EXTRA_RPMS` here, but I'm afraid that conflating these two use cases in one script might make things complicated (for example, we must ensure that the right versions of the packages are installed). It might be simpler to have this in two scripts.

test/check-cli
larskarlitski: Let's call the test scripts directly from here, as this script is already called from the Makefile. Too many circles :)

Yes, we can store credentials in the CI infrastructure. We already do this for GitHub tokens and the RHEL subscription. Are you ok to get the basic machinery working in this PR with only the tests that run without extra credentials. Then do a follow up with one that adds support for the cloud tests?

larskarlitski:
What do you think, should we merge `test` and `tests`? The entry point for Cockpit's infrastructure is `test/run`, so we're at least stuck with that.

test/vm.install
```diff
@@ -0,0 +1,10 @@
...
+
+# librepo 1.9.6-1 crashes when fetching metadata, which breaks lorax-composer
+sudo dnf update -y librepo
+
+sudo dnf install -y beakerlib /var/tmp/*.rpm
+
+systemctl enable lorax-composer.socket
```
bcl: Since this is a test we may as well just start up lorax-composer.service instead, may save a small amount of time waiting for the metadata to be fetched.

larskarlitski: This script is only run once to install the virtual machine. Starting the service at this stage won't make a difference. (The VM is then booted for every test.)

bcl: If we use lorax-composer.service instead it will start at boot time when it reboots.
Enabling the  .socket is really most useful for systems where you may or many not be using it right away.

larskarlitski: Ah sorry, I misunderstood (you probably didn't mean that, but wrote "_start_ up..." instead of "enable ..."- systemd nomenclature is weird).

As you say, it will probably only make a small difference, but it's cleaner indeed. Thanks!

martinpitt: Please do whatever happens on a production system. It doesn't hurt to test that the socket activation (including the delay on the first call) works as expected.

test/check-cli
atodorov: Add 3rd parameter the path to the test you want to execute. Probably remove the target from Makefile ?

Will require `./$TEST` to change to `$TEST` inside test_cli.sh to account for the fact that you are passing absolute path. I've tried it locally and it worked for me.

test/vm.install
larskarlitski: Good point. Left it as is.

test/check-cli
larskarlitski: I converted `test/check-cli` to use `unittest` and included all scripts under `tests/cli`. I've also changed `./$TEST` to `$TEST`. I left the `make` targets in there, because you said you sometimes manually run the tests for different rpms?

We probably want to run some of the tests in `/tests/composer` in virtual machines as well.

test/check-cli
```diff
@@ -0,0 +1,50 @@
...
+        self.assertEqual(r.returncode, 0)
+
+    def testBlueprintsSanity(self):
+        self.runCliTest("/tests/cli/test_blueprints_sanity.sh")
+
+    def testComposeSanity(self):
+        self.runCliTest("/tests/cli/test_compose_sanity.sh")
+
+
+class TestImages(TestCli):
```
atodorov: This will also inherit the `test*Sanity` methods and execute them when this class is executed, which will add more time to the execution duration. Probably not what you want.

IMO move `runCliTest()` into `ComposerTestCase` class.

test/check-cli
```diff
@@ -0,0 +1,50 @@
...
+        self.runCliTest("/tests/cli/test_compose_tar.sh")
+
+    def testQcow2(self):
+        self.runCliTest("/tests/cli/test_compose_qcow2.sh")
+
+    def testLiveIso(self):
+        self.runCliTest("/tests/cli/test_compose_live-iso.sh")
+
+
+class TestCloud(TestCli):
```
atodorov: Just a note at this point:
here all of the test methods will be executed in sequence and uploads (especially for Azure) tend to take a lot of time. We want to execute all of the cloud methods in parallel, and have them report separate statuses via the GitHub API. This we can deal with later.

Execution wise all of these need a VM just to build the image and then everything else happens on the cloud vendor side.

test/composertest.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def execute(self, command, **args):
+        """Execute a command on the test machine.
+
+        **args and return value are the same as those for subprocess.run().
+        """
+        return subprocess.run(self.ssh_command + command, **args)
+
+
+def main():
```
atodorov: That should be `if __name__ == "__main__":`

atodorov:
@larskarlitski - IMO you can move everything from `test` to `tests` but that will make the PR very big

test/check-cli
larskarlitski: Oops indeed. Changed it back. Thanks!

test/check-cli
larskarlitski: Yeah... I've moved them into a separate file and made the test being run dependent on `TEST_SCENARIO`, which is set by Cockpit's test framework to get separate GitHub statusses.

test/composertest.py
larskarlitski: No, this script is not meant to be run by itself. The `main` method is used by the `check-*` scripts. It will grow command line options in the future, like `--sit`.

larskarlitski:
Update:
* Resize the image to make room for a composer-generated one
* Add `test/README.md`
* incorporate Alex' suggestions (thanks!)

> @larskarlitski - IMO you can move everything from `tests` to `test` but that will make the PR very big

Let's do this later to keep this easy to review.

martinpitt:
@larskarlitski : Seems we need a few more packages in our cockpit/tasks container for building?

larskarlitski:
> @larskarlitski : Seems we need a few more packages in our cockpit/tasks container for building?

Yeah. How do I go about doing that?

martinpitt:
@larskarlitski : They need to be added to https://github.com/cockpit-project/cockpituous/blob/master/tasks/Dockerfile , the container rebuilt and pushed (`sudo make tasks-container tasks-push`). Then we restart the bots.

martinpitt:
@larskarlitski : New containers are rolled out, so this is the time to try again.

martinpitt:
The travis errors also seem relevant.

larskarlitski:
> The travis errors also seem relevant.

Indeed! One error wasn't. I've put that in a separate commit.

larskarlitski:
Another update: travis was still failing because `test` does exist now and the `test` make target was not in `.PHONY`.

Also remove the hack around resizing the image manually now. https://github.com/cockpit-project/cockpit/pull/11681 is merged, so we can use `image-customize --resize` now. (The hack didn't work on Cockpit's infrastructure, because it didn't take `TEST_DATA` into account.)

larskarlitski:
```
- nothing provides python(abi) = 3.6 needed by composer-cli-31.3-1.fc28.x86_64
```
@pitti Looks like we want to build the rpms inside of the VM? (Fedora 30 has python 3.7, but the build container only 3.6.)

The remaining travis errors seem unrelated. @atodorov do they work on master?

test/README.md
```diff
@@ -0,0 +1,56 @@
...
+        libguestfs-tools qemu qemu-kvm rpm-build rsync xz
+
+## Building a test VM
+
+To build a test VM, run
+
+    $ make vm
+
+This downloads a base image from Cockpit's infrastructure. You can control
+which image is downloaded with the `TEST_OS` environment variable. It then
```
atodorov: Link or examples of acceptable values would be nice

test/README.md
```diff
@@ -0,0 +1,56 @@
...
+creates a new image based one that (a qemu snapshot) in `tests/images`, which
```
atodorov: typo: one -> on

test/run
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+# This is the expected entry point for Cockpit CI; will be called without
+# arguments but with an appropriate $TEST_OS, and optionally $TEST_SCENARIO
+
+make vm
+
+if [ -n "$TEST_SCENARIO" ]; then
+  test/check-cloud TestCloud.test_$TEST_SCENARIO
```
atodorov: Can $TEST_SCENARIO actually be defined as Class.test_Method? Then we can move all of the runCliTest methods in one file and be able to start whichever scenario we want. Sort of treating the new Python unittest files like a substitute for Makefile.

test/README.md
martinpitt: https://github.com/cockpit-project/cockpit/blob/master/test/README.md#test-configuration

atodorov:
> The remaining travis errors seem unrelated. @atodorov do they work on master?

No, master is broken in Travis CI. I think this has to do with recent rpm changes which broke rpmfluff, which Brian mentioned some time ago.

larskarlitski:
Sorry, I didn't have time to work on this recently (vacation and Red Hat Summit).

Update: As I mentioned in a previous comment, we need to build the packages on the test VM itself to get the packages to depend on the right python version. I've changed `vm.install` to build them in a mock. Alas, this makes `make vm` a bit slower, because mock takes some time to unpack/pack the build root.

On first run, I see this error from mock: @martinpitt, do you know what could be going on here?
```
Warning: Module or Group 'buildsys-build' does not exist.
```

test/README.md
larskarlitski: Good point, thanks. I've added this link.

test/README.md
larskarlitski: :+1:

test/run
larskarlitski: I think this is going a bit far. The test runners won't run all those jobs in parallel anyway. Are you ok with going forward like this and changing it if we notice problems?

martinpitt:
> Alas, this makes make vm a bit slower, because mock takes some time to unpack/pack the build root.

Our test images have a pre-unpacked mock chroot in /var/lib/mock. You should be able to run

    su builder -c "/usr/bin/mock --offline --no-clean --shell"

without much delay (< 1 s). Same with building the actual package, of course. For building lorax you probably need to drop  `--offline` so that it can install new build deps; we use that for Cockpit as we know/want to ensure everything is pre-installed. Once that works, we can of course do that for lorax too, to speed up the tests.

Makefile
```diff
@@ -95,12 +101,27 @@ set-docs-owner:
...
 archive:
 	@git archive --format=tar --prefix=$(PKGNAME)-$(VERSION)/ $(TAG) > $(PKGNAME)-$(VERSION).tar
-	@gzip $(PKGNAME)-$(VERSION).tar
+	@gzip -f $(PKGNAME)-$(VERSION).tar
 	@echo "The archive is in $(PKGNAME)-$(VERSION).tar.gz"
 
 dist: tag archive
 	scp $(PKGNAME)-$(VERSION).tar.gz fedorahosted.org:lorax
 
+rpm: archive $(PKGNAME).spec
```
martinpitt: Is that still an actually useful target? Or just for developer convenience? If the former, please add it to `.PHONY`.

test/vm.install
```diff
@@ -0,0 +1,34 @@
...
+
+# Grow root partition to make room for images. This only works on Fedora right now.
+echo ", +" | sfdisk -N 2 -f /dev/vda
+partprobe
+pvresize /dev/vda2
+lvresize fedora/root -l+100%FREE -r
+
+tmpdir=$(mktemp -d $PWD/srpm-build.XXXXXX)
+tar -xaf $TEST_SOURCE -O $(basename --suffix .tar.gz $TEST_SOURCE)/lorax.spec > $tmpdir/lorax.spec
+rpmbuild -bs \
```
martinpitt: FTR, the *source* rpm can easily be built in the container, outside of the VM, like we do in starter-kit and friends. This works as well, but the .srpm is an useful developer artifact to have outside, for copr, mock builds, inspection, etc.

test/vm.install
```diff
@@ -0,0 +1,34 @@
...
+         --define "_srcrpmdir `pwd`" \
+         --define "_rpmdir $tmpdir" \
+         --define "_buildrootdir $tmpdir/.build" \
+         $tmpdir/lorax.spec
+
+srpm=$(rpm --qf '%{Name}-%{Version}-%{Release}.src.rpm\n' -q --specfile $tmpdir/lorax.spec | head -n1)
+rm -rf $tmpdir
+
+mock --installdeps $srpm
+mock --offline --resultdir build-results --rebuild $srpm
```
martinpitt: Please run these two as user "builder". This may explain the permission errors you saw.

test/run
atodorov: +1 for merging as-is and I can tweak the rest later.

larskarlitski:
Thanks Martin! I've changed it to build the srpm in the container / on the host (and added `srpm` to `.PHONY`) and am running mock as `builder` now. This makes `vm.install` much cleaner.

Makefile
```diff
@@ -129,8 +141,34 @@ docs-in-docker:
...
+	git fetch --depth=1 https://github.com/cockpit-project/cockpit.git
+	git checkout --force FETCH_HEAD -- bots/
+	git reset bots
+
 .PHONY: ci_after_success
 ci_after_success:
 # nothing to do here, but Jenkins expects this to be present, otherwise fails
 
-.PHONY: docs
+.PHONY: docs check test srpm
```
martinpitt: plus `vm vm-reset`

Makefile
```diff
@@ -129,8 +141,34 @@ docs-in-docker:
 
 ci: check test
 
+$(VM_IMAGE): TAG=HEAD
+$(VM_IMAGE): archive bots
```
martinpitt: I think this needs to depend on `srpm`, not `archive`?

larskarlitski: Yeah, indeed. Worked for me because I didn't delete it between runs.

Makefile
larskarlitski: Thanks!

martinpitt:
[last run](https://209.132.184.41:8493/logs/pull-689-20190517-100707-cc6a6561-cockpit-fedora-30/log) failed with an ENOSPC, investigating whether it's a problem inside the VM (needs to grow larger?) or outside in the infra.

martinpitt:
This PR is alright for the test integration changes now. But it still fails due to ENOSPC, the test should be fixed to clean up built images after it's done verifying them. Otherwise the demands on the test beds are just unreasonably high (also for developers running the test locally).

martinpitt:
@larskarlitski : FYI, these are the two commands to run in the 
[webhook container](https://console.apps.ci.centos.org:8443/console/project/cockpit/browse/pods/webhook-mwqw2?tab=terminal) to trigger a test:
```
bots/tests-trigger --repo weldr/lorax 689 cockpit/fedora-30
bots/tests-scan --repo weldr/lorax -p 689 -c cockpit/fedora-30 --amqp amqp.cockpit.svc.cluster.local:5671
```

larskarlitski:
Added a few commits to fix tar and qcow2 tests and to make the log more readable (no `tar -v`, set `TEST` so that the test name appears in the beakerlib summary, and incrrease `unittest`'s verbosity.

test/vm.install
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh -eux
+
+SRPM="$1"
+
+# Grow root partition to make room for images. This only works on Fedora right now.
+echo ", +" | sfdisk -N 2 -f /dev/vda
+partprobe
+pvresize /dev/vda2
```
martinpitt: This works on RHEL images, too, as "vda2" is defined by the outside image build process and QEMU/libvirt invocation. If you want to keep it a bit more flexible, use 

    pvs --noheadings -opv_name | xargs pvresize

test/vm.install
```diff
@@ -0,0 +1,18 @@
...
+lvresize fedora/root -l+100%FREE -r
```
martinpitt: Generalization that also works on RHEL:
```
rootlv=$(findmnt --noheadings -oSOURCE /)
lvresize $rootlv -l+100%FREE -r
```

test/composertest.py
```diff
@@ -0,0 +1,59 @@
...
+
+# import Cockpit's machinery for test VMs and its browser test API
+sys.path.append(os.path.join(os.path.dirname(__file__), "../bots/machine"))
+import testvm # pylint: disable=import-error
+
+
+class ComposerTestCase(unittest.TestCase):
+    image = testvm.DEFAULT_IMAGE
+
+    def setUp(self):
```
jikortus: Some of the current tests need some sort of specific setup (changing the existing files used by composer) - I think this could be the right place to do this setup? So far I can think of two cases - changing sshd configuration to accept login with empty passwords and modifying a blueprint to contain custom kernel commandline parameters - please see https://github.com/weldr/lorax/blob/master/tests/test_cli.sh#L7.

larskarlitski: Yes and no :)

This `setUp` is called for all tests. Specific setup that should only happen for some of the tests should go into subclasses of this class (see test/check-cli for example, you could add a `setUp` there).

For now, the tests simply call out the `tests/cli/*.sh` scripts. We don't need to move anything over, unless it's easier to do in python.

larskarlitski:
Update: increase the timeout for starting lorax to 6 minutes. 90 seconds was too little.

martinpitt:
Whee! Thanks @larskarlitski !

test/composertest.py
atodorov: FTR @jikortus means we do modify some of the blueprints/kickstart files to make it easier for testing. For example live-iso by default forbids SSH connections because root account doesn't have a password there. 

Since this PR has already been merged we can work on fixing the remaining issues one by one. However be mindful that QE uses the existing bash scripts as they are to validate RPM builds so they must continue to work in stand-alone mode. Which likely means duplicating setup in some places (yikes).

PR #706, opened 2019-05-06
Add test for passing custom option on kernel command line

jikortus:

--- Merge policy ---

- [x] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---


atodorov:
@jikortus 
`:: [ 10:37:07 ] :: [   FAIL   ] :: root account is disabled in /etc/shadow (Expected 0, got 1)`

for OpenStack. Looks like the new test discovered a bug, please investigate.

The other failures are all known issues which are being worked on.

jikortus:
@atodorov Thanks for pointing this out. It probably isn't a bug, it rather seems I have to change the regex checking if the root account is disabled - we've been using `^root:(\*LOCK\*|!):` so far. This is what I found in the built compose in `/etc/shadow`:

`root:!!$6$MXuTViMuVGURPO8V$gAfZftDhnuKWkrLwta45MJP5RuXYqiZ4rwWEJK6j1ibC17bsppMUmuEhc0ZOLri432CyMmG3DvO1O.ueu94wO0::0:99999:7:::`

Based on that, it's clear that the account is disabled. Other question is, however, if it's actually OK this way and thus the regex needs to be updated, or if we want to insist on a particular form for the root account record. @bcl, what do you think about this please?

bcl:
> Other question is, however, if it's actually OK this way and thus the regex needs to be updated, or if we want to insist on a particular form for the root account record.

I don't think we want to make this too strict. root:!! is a valid way to lock an account that already has a password set on it. Unlocking removes the !! and restores the login capability. Just drop the trailing ':' from the regex and it should be fine since it is matching the start of the string you don't really need that part.

larskarlitski:
`passwd -S <account>` returns status of an account, including whether it's locked. Cockpit uses that to find out locked status: https://github.com/cockpit-project/cockpit/blob/master/pkg/users/local.js#L810.

PR #711, opened 2019-05-09
Create a lorax-docs package with the html docs

bcl:
Also include the other top level .rst docs in the main lorax package.

Resolves: rhbz#1695274

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -87,6 +87,14 @@ It also includes livemedia-creator which is used to create bootable livemedia,
 including live isos and disk images. It can use libvirtd for the install, or
 Anaconda's image install feature.
 
+%package docs
+Summary: Lorax html documentation
+Requires: lorax = %{version}-%{release}
+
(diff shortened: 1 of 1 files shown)


lorax.spec
```diff
@@ -196,7 +204,9 @@ getent passwd weldr >/dev/null 2>&1 || useradd -r -g weldr -d / -s /sbin/nologin
 %files
 %defattr(-,root,root,-)
 %license COPYING
-%doc AUTHORS docs/livemedia-creator.rst docs/product-images.rst
+%doc AUTHORS
+%doc docs/composer-cli.rst docs/lorax.rst docs/livemedia-creator.rst docs/product-images.rst
+%doc docs/lorax-composer.rst
```
atodorov: Why not `docs/*.rst` ?

bcl: Because some of the .rst files are useless without sphinx, like index.rst for example.

PR #712, opened 2019-05-10
Use ansible instead of awscli

jstodola:

The following parts of the script haven't been rewritten to ansible yet:
* Import of the image from S3 as a snapshot
* Registerring AMI image from the imported snapshot.

Ansible modules do not support importing an s3 object as a snapshot, reported as:
https://github.com/ansible/ansible/issues/53453

A possible workaround using the image_location parameter of the ec2_ami ansible module is being investigated.

--- Merge policy ---

- [ ] Travis CI PASS
- [x] `*-aws-runtest` PASS
- [x] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [x] `*-openstack-runtest` PASS
- [x] `*-vmware-runtest` PASS

--- Jenkins commands ---

### DIFF
--- a/tests/cli/test_build_and_deploy_aws.sh
+++ b/tests/cli/test_build_and_deploy_aws.sh
@@ -33,12 +33,14 @@ rlJournalStart
         rlLogInfo "AWS_BUCKET=$AWS_BUCKET"
         rlLogInfo "AWS_REGION=$AWS_REGION"
 
-        if ! rlCheckRpm "python3-pip"; then
-            rlRun -t -c "dnf -y install python3-pip"
-            rlAssertRpm python3-pip
-        fi
+        for package in python3-pip python3-boto3; do
+            if ! rlCheckRpm "$package"; then
+                rlRun -t -c "dnf -y install $package"
+                rlAssertRpm "$package"
+            fi
+        done
 
-        rlRun -t -c "pip3 install awscli"
+        rlRun -t -c "pip3 install awscli ansible[aws]"
 
         # aws configure
         [ -d ~/.aws/ ] || mkdir ~/.aws/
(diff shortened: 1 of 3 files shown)


larskarlitski:
Nice! Looks like this is much less complex than using `awscli` directly. Is there another benefit of switching to ansible?

Is that role using `awscli` under the hood? Is it ok that we're not testing with `awscli` anymore?

Is the idea to switch tests for other cloud providers to ansible as well?

I think it'd be more readable if the playbook were in separate files.

tests/cli/test_build_and_deploy_aws.sh
```diff
@@ -38,7 +38,7 @@ rlJournalStart
             rlAssertRpm python3-pip
         fi
 
-        rlRun -t -c "pip3 install awscli"
+        rlRun -t -c "pip3 install awscli ansible[aws]"
```
atodorov:
```
STDOUT: TASK [Make sure bucket exists] ***********************************************************************************************************
STDOUT: fatal: [localhost]: FAILED! => {"changed": false, "msg": "Python modules \"botocore\" or \"boto3\" are missing, please install both"}
```

atodorov: Installing python3-boto3 from RPM fixed it for me. (Or if you prefer from pypi).

atodorov:
> I think it'd be more readable if the playbook were in separate files.

@larskarlitski how do we pass specific values then (which we have as variables in the script) ?

jstodola:
I agree that separating the ansible playbooks makes sense to improve readability. Passing the values can be done via --extra-vars: 
ansible-playbook release.yml --extra-vars "version=1.23.45 other_variable=foo"
Doc: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#passing-variables-on-the-command-line

tests/cli/test_build_and_deploy_aws.sh
jstodola: Fixed, python3-boto3 rpm is now installed

jstodola:
Latest version installs python3-boto3 rpm and ansible playbooks are in separate files.

jstodola:
Removed .instance.yml.swp, thanks for catching it.
The cleanup script will be tracked in a separate PR.

PR #720, opened 2019-05-21
tests: Source lib.sh from the right directory

larskarlitski:
When testing in one of the VMs, tests are in /tests. Source the lib from
a relative directory instead.

### DIFF
--- a/test/check-cli
+++ b/test/check-cli
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 
 import composertest
+import unittest
 
 
 class TestSanity(composertest.ComposerTestCase):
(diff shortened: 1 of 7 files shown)


larskarlitski:
This means we weren't running the full tests yet. Unfortunately, beakerlib did not fail when a command was not found:

```
/tests/cli/test_compose_qcow2.sh: line 77: verify_image: command not found
```

I filed https://github.com/weldr/lorax/issues/727 to make the tests more robust.

larskarlitski:
Let's wait for https://github.com/weldr/lorax/pull/730 to land and rebase. `beakerlib` is missing on fedora-30 image and we didn't notice because of https://github.com/weldr/lorax/issues/727.

martinpitt:
This test wrongly passed due to the lack of beakerlib. This got fixed in #730, I restarted the test.

larskarlitski:
The live-iso test seems broken. I've added a commit that skips it and filed https://github.com/weldr/lorax/issues/731

PR #723, opened 2019-05-21
test: Generalize fs resizing in vm.install

larskarlitski:
This should work on RHEL as well.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -2,6 +2,17 @@
 
 SRPM="$1"
 
+if ! rpm -q beakerlib && [ $(. /etc/os-release && echo $ID) = "rhel" ]; then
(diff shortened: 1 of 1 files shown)


larskarlitski:
Rebased to fix conflict and typo in commit message.

martinpitt:
The [rhel-8-1](https://209.132.184.41:8493/logs/pull-723-20190521-204545-4519e22b-cockpit-rhel-8-1/log) test got past the beakerlib installation and the fs resizing (which provably worked as the log shows the new disk size, 19 GB). It fails later on as master's .spec dependencies don't work on RHEL 8.1, but this isn't related to this PR. So that's fine. I'll wait until the fedora-30 test is green, to prove that this didn't break anything there, and then land this. Thanks @larskarlitski !

PR #726, opened 2019-05-21
Update live iso kickstarts

bcl:
Some things in Fedora have changed. Update the live iso templates so that a live iso actually works.

### DIFF
--- a/docs/fedora-livemedia.ks
+++ b/docs/fedora-livemedia.ks
@@ -175,6 +175,10 @@ systemctl --no-reload disable atd.service 2> /dev/null || :
 systemctl stop crond.service 2> /dev/null || :
 systemctl stop atd.service 2> /dev/null || :
 
+# turn off abrtd on a live image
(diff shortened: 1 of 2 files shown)


larskarlitski:
Accidentally manually triggered the rhel-8-1 test. Please ignore it. Sorry!

larskarlitski:
These two changes seem to be synced from the [official fedor live cd kickstart](https://pagure.io/fedora-kickstarts/blob/master/f/fedora-live-base.ks). I wonder if it makes sense to sync the whole thing?

bcl:
> These two changes seem to be synced from the [official fedor live cd kickstart](https://pagure.io/fedora-kickstarts/blob/master/f/fedora-live-base.ks). I wonder if it makes sense to sync the whole thing?

No, I want to (try) to keep the example a bit smaller than the official one. And for the composer template we certainly don't want to pull in all of its packages.

PR #730, opened 2019-05-22
test: Install beakerlib on non-RHEL images

martinpitt:
Commit 4783f6562ffe introduced the assumption that beakerlib is already
installed in non-RHEL images. As the test script runs without `set -e`,
this hasn't been noticed as the test silently succeeds.
    
Go back to installing beakerlib everywhere.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -2,15 +2,19 @@
 
 SRPM="$1"
 
-if ! rpm -q beakerlib && [ $(. /etc/os-release && echo $ID) = "rhel" ]; then
-  (cd /etc/yum.repos.d; curl -O -L http://download.devel.redhat.com/beakerrepos/beaker-client-RedHatEnterpriseLinux.repo)
+if ! rpm -q beakerlib; then
+    if [ $(. /etc/os-release && echo $ID) = "rhel" ]; then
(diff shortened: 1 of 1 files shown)


martinpitt:
Correction: In PR #723, the [fedora-30](https://209.132.184.41:8493/logs/pull-730-20190522-144545-081dd99c-cockpit-fedora-30/log) test silently succeeded, but beakerlib also wasn't installed. Seriously, tests really can't run without `set -e`. beakerlib needs to support some kind of `trap` cleanup actions that are done on failure.

martinpitt:
Force-pushed to update the commit description, to avoid confusing people.

atodorov:
LGTM however cockpit/fedora-30 reported failure, not sure why. Looks like a timeout to me.

I think this is safe to merge though.

larskarlitski:
> LGTM however cockpit/fedora-30 reported failure, not sure why. Looks like a timeout to me.

Thanks! A return code of `124` is indeed a timeout. It happened because the test container ran out of space. @martinpitt is working on a fix.

martinpitt:
Fixed container config rolled out, test restarted.

martinpitt:
test passed now! landing, this is urgent and blocks all other PRs.

PR #735, opened 2019-05-24
Enable cloud tests for Cockpit CI

atodorov:

As a started revert a few `rlAssert` statements which got deleted during the migration.

I will kick-off the cloud tests manually at first and add more commits here if necessary.

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -4,7 +4,9 @@ lorax uses Cockpit's integration test framework and infrastructure. To do this,
 we're checking out Cockpit's `bots/` subdirectory. It contains links to test
 images and tools to manipulate and start virtual machines from them.
 
-Each test is run on a new instance of a virtual machine. 
+Each test is run on a new instance of a virtual machine.
+Branch/test matrix is configured in `bots/tests-scan` inside the
+[cockpit repository](https://github.com/cockpit-project/cockpit).
 
 ## Dependencies
 
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -77,7 +77,10 @@ class ComposerTestCase(unittest.TestCase):
         return subprocess.run(self.ssh_command + command, **args)
 
     def runCliTest(self, script):
-        r = self.execute(["CLI=/usr/bin/composer-cli", "TEST=" + self.id(), "/tests/test_cli.sh", script])
+        r = self.execute(["CLI=/usr/bin/composer-cli",
+                          "TEST=" + self.id(),
+                          "PACKAGE=composer-cli",
+                          "/tests/test_cli.sh", script])
         self.assertEqual(r.returncode, 0)
 
 
--- a/test/run
+++ b/test/run
@@ -5,7 +5,13 @@
 make vm
 
 if [ -n "$TEST_SCENARIO" ]; then
-  test/check-cloud TestCloud.test_$TEST_SCENARIO
+  if [ "$TEST_SCENARIO" == "live-iso" ]; then
+    test/check-cli TestLiveIso
+  elif [ "$TEST_SCENARIO" == "qcow2" ]; then
+    test/check-cli TestQcow2
+  else
+    test/check-cloud TestCloud.test_$TEST_SCENARIO
+  fi
 else
-  test/check-cli
+  test/check-cli TestImages
 fi
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,10 @@ $(VM_IMAGE): srpm bots
 		--run-command "chmod +x /var/tmp/vm.install" \
 		--run-command "cd /var/tmp; /var/tmp/vm.install $$srpm" \
 		$(TEST_OS)
+	[ -f "~/.config/lorax-test-env" ] && bots/image-customize \
+		--upload ~/.config/lorax-test-env:/var/tmp/lorax-test-env \
+		$(TEST_OS) || echo
+
 
 # convenience target for the above
 vm: $(VM_IMAGE)
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,6 @@ test:
 
 	coverage3 report -m
 	[ -f "/usr/bin/coveralls" ] && [ -n "$(COVERALLS_REPO_TOKEN)" ] && coveralls || echo
-	
-	./tests/test_cli.sh
 
 # need `losetup`, which needs Docker to be in privileged mode (--privileged)
 # but even so fails in Travis CI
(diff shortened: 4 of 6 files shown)


atodorov:
> What are the asserts for QEMU for?

Because we execute these scripts outside CI as well and these asserts are going to tell us when something is missing. I had them wrong initially so I've updated them now.

atodorov:
Also see https://github.com/cockpit-project/cockpit/pull/11933

atodorov:
@larskarlitski note/question:

The sanity tests are already included in Travis CI but I think we can move them out of there for consistency. (and I will document that all integration type tests are executed outside Travis) What do you think?

test/run
```diff
@@ -5,7 +5,7 @@
 make vm
 
 if [ -n "$TEST_SCENARIO" ]; then
-  test/check-cloud TestCloud.test_$TEST_SCENARIO
+  test/check $TEST_SCENARIO
 else
-  test/check-cli
+  test/check TestImages
```
martinpitt: Wouldn't it make more sense to specify "TestImages" as a scenario as well? Then this would avoid the entire if condition, and if it's empty, just calling `test/check` should still DTRT and run all tests?

larskarlitski:
> Because we execute these scripts outside CI as well and these asserts are going to tell us when something is missing. I had them wrong initially so I've updated them now.

But the assert doesn't fail. If you run the remaining stages anyway, they will fail.

I agree that it makes sense to fail early, but this doesn't do that.

larskarlitski:
> The sanity tests are already included in Travis CI but I think we can move them out of there for consistency. (and I will document that all integration type tests are executed outside Travis) What do you think?

Good idea! In addition, they'll get tested on different operating systems as well.

test/run
atodorov: See this comment:
https://github.com/cockpit-project/cockpit/pull/11933#discussion_r287764189

We already have the image tests specified as a scenario. I will update the scenarios for better grouping. However if nothing is selected then all tests will be executed which will be quite long. Each of the cloud tests takes around 30-40 mins depending on network speed. IMO it is better to default to something smaller if no scenarios are specified.

martinpitt: Agreed about the default if nothing is specified, that's a good point.

martinpitt:
The cloud failures  are expected, due to missing credentials. But [TestImages](https://209.132.184.41:8493/logs/pull-735-20190528-100909-8b5c5425-cockpit-fedora-30-TestImages/log) fails in the same way as the previous [test without a scenario](https://209.132.184.41:8493/logs/pull-735-20190527-133131-8b5c5425-cockpit-fedora-30/log). This doesn't need to block *this* PR of course, but it needs to be fixed ASAP -- somehow this regression slipped in in some PR.

.github/pull_request_template.md
```diff
@@ -6,16 +6,11 @@
...
-- [ ] `*-azure-runtest` PASS
-- [ ] `*-images-runtest` PASS
-- [ ] `*-openstack-runtest` PASS
-- [ ] `*-vmware-runtest` PASS
+- [ ] `cockpit/*/TestSanity` PASS
+- [ ] `cockpit/*/TestImages` PASS
+- [ ] `cockpit/*/TestAWS` PASS
+- [ ] `cockpit/*/TestAzure` PASS
+- [ ] `cockpit/*/TestOpenStack` PASS
+- [ ] `cockpit/*/TestVmware` PASS
```
larskarlitski: I think we can leave those out now: GitHub show the status and even allows marking some tests as `required`.

test/README.md
```diff
@@ -4,7 +4,8 @@ lorax uses Cockpit's integration test framework and infrastructure. To do this,
 we're checking out Cockpit's `bots/` subdirectory. It contains links to test
 images and tools to manipulate and start virtual machines from them.
 
-Each test is run on a new instance of a virtual machine. 
+Each test is run on a new instance of a virtual machine.
+Branch/test matrix is configured in `bots/tests-scan`.
```
larskarlitski: Please mention that it's configured on the [cockpit-project/cockpit](https://github.com/cockpit-project/cockpit) repository.

test/composertest.py
```diff
@@ -77,7 +77,10 @@ def execute(self, command, **args):
         return subprocess.run(self.ssh_command + command, **args)
 
     def runCliTest(self, script):
-        r = self.execute(["CLI=/usr/bin/composer-cli", "TEST=" + self.id(), "/tests/test_cli.sh", script])
+        r = self.execute(["CLI=/usr/bin/composer-cli",
+                          "TEST=" + self.id(),
+                          "PACKAGE_NAME=composer-cli",
```
larskarlitski: `PACKAGE_NAME` is `composer-cli`?

atodorov:
> Why do you think this is preferable to having the cloud tests in a separate file

On the cockpit PR you mentioned that you don't like the idea of passing Python class name/test method as TEST_SCENARIO. If we want to have a short and easy to read scenarios then all of the test classes and methods need to be in the same file.

As it is right now we've got sanity tests + image tests separated b/c they take time. Cloud tests should also be separated b/c they take even more time. ATM this simply won't work when there are more than one `check-*` Python scripts.

larskarlitski:
> If we want to have a short and easy to read scenarios then all of the test classes and methods need to be in the same file.

On current master, they are `cockpit/fedora-30/aws`. That's short. I have no opinion on the actual names (other than not to use python's test names), but there's no need to move everything into one script.

atodorov:
> On current master, they are cockpit/fedora-30/aws

this currently works only for splitting cloud tests (assuming all of them are in check-cloud)!

With the current master there's no way I can split sanity tests from image tests or split live-iso test from qcow, tar, etc if we decide they take too long and want to execute them in parallel.

I'm happy with whatever scheme which will allow us to arbitrarily group test scenarios together. Having the Python classes defined in multiple files just doesn't allow this.

larskarlitski:
I think https://github.com/weldr/lorax/pull/735/commits/012dfb485a14e62bbadf724730ea972e6c9cc6f1 is too inflexible. I'd like to be able to add tests without changing the test matrix in cockpit's repository.

The main point is to cut down test time. To get towards that quickly, I've opened https://github.com/weldr/lorax/pull/742, which only splits the existing image tests in two.

PR #736, opened 2019-05-24
Update qemu arguments to work correctly with nographic

bcl:
Add -monitor none to turn off the qemu monitor multiplexing.
Pass -boot d for -cdrom booting instead of 'c'.

Add 'console=ttyS0,115200n8' to the boot arguments so that kernel output
will show up on the serial port.

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---

### DIFF
--- a/tests/cli/test_compose_live-iso.sh
+++ b/tests/cli/test_compose_live-iso.sh
@@ -46,7 +46,7 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Start VM instance"
-        rlRun -t -c "$QEMU -m 2048 -boot c -cdrom $IMAGE -nographic \
+        rlRun -t -c "$QEMU -m 2048 -boot d -cdrom $IMAGE -nographic -monitor none \
                            -net user,id=nic0,hostfwd=tcp::2222-:22 -net nic &"
         # 60 seconds timeout at boot menu screen
         # then media check + boot ~ 30 seconds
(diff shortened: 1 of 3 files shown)


larskarlitski:
> LGTM however tests are still failing. I've restarted them and would like to see them pass before approving this PR.

The fedora-30 test fails for me locally as well. I think this PR doesn't fix the bug.

atodorov:
> The fedora-30 test fails for me locally as well. I think this PR doesn't fix the bug.

The live-iso test fails both in Jenkins and in Cockpit CI. I will try debugging it locally and see what is going on. It may be related to something else.

atodorov:
@larskarlitski, @bcl I've run the live-iso test manually and it passed. The failure reported in the logs above looks like this:
https://bugzilla.redhat.com/show_bug.cgi?id=1710877

which is some kind of networking problem where the VM doesn't get an IP address and ssh fails.

bcl:
> which is some kind of networking problem where the VM doesn't get an IP address and ssh fails.

If that were the case we should still see the kernel output getting dumped to stdio. I think it's not booting, or the timer is too short (which doesn't seem likely since the logs seem to say it waited for 266s).

PR #741, opened 2019-05-30
composer: Don't reply with HTML on HTTP errors

larskarlitski:
Override flask's default error handler, because that returns html. Return JSON instead with the usual `{ "status": false, "errors": [ ... ] }` pattern.

Also add a new integration test `check-api` which tests this change.

### DIFF
--- a/src/pylorax/api/errors.py
+++ b/src/pylorax/api/errors.py
@@ -16,6 +16,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+# HTTP errors
+HTTP_ERROR = "HTTPError"
+
 # Returned from the API when either an invalid compose type is given, or not
 # compose type is given.
 BAD_COMPOSE_TYPE = "BadComposeType"
--- a/src/pylorax/api/server.py
+++ b/src/pylorax/api/server.py
@@ -21,9 +21,11 @@ from collections import namedtuple
 from flask import Flask, jsonify, redirect, send_from_directory
 from glob import glob
 import os
+import werkzeug
 
 from pylorax import vernum
 from pylorax.api.crossdomain import crossdomain
+from pylorax.api.errors import HTTP_ERROR
 from pylorax.api.v0 import v0_api
 from pylorax.sysutils import joinpaths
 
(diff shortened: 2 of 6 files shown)


larskarlitski:
> Looks good, but we should also have a test in test_server.py

Added a corresponding test there.

> IMO this type of test is better suited inside the existing unit test suite, for example see `test_server.py`

Why?

atodorov:
> Why?

- because unit test will be faster compared to integration test
- because test_server.py already has other similar tests (for /api/status there is test_01_status)
- because the change will be smaller, e.g. something like

```
response = self.server.get("/marmalade")
data = json.loads(response.data)

self.assertEqual(response.status_code, 404)
self.assertDictEqual(data, {
        "status": False,
        "errors": [{ "id": "HTTPError", "code": 404, "msg": "Not Found" }]
    })
```

larskarlitski:
Thanks for giving your reasons :)

I've added the test in addition to the other one.

> * because unit test will be faster compared to integration test 

They're not really unit tests. They're hard to run individually, are not separate from each other, need to be run as root, and leave things in `/tmp` when an error happens.

They also take quite long, because they're fetching dnf metadata before running a test.

> * because test_server.py already has other similar tests (for /api/status there is test_01_status)

I like to have integration tests for the API as well. `test_server.py` is not that, because it doesn't test the installed version of lorax-composer and doesn't talk to it via the HTTP API.

> * because the change will be smaller, e.g. something like

I think the benefit of having an integration test outweigh this.

atodorov:
@larskarlitski FTR while you are correct about the existing API test suite (yes, technically they are not unit tests) your change is at odds with everything else that exists.  I am not 100% opposed to that (b/c more test coverage is always better than no test coverage) but it is weird to me, doesn't feel natural.

Everything that exists up to now in terms of integration tests doesn't talk to the API directly. Until recently it did run against the local git checkout, with the Cockpit CI changes it does build and install the RPM first. All the bash scripts available till now were necessary so we could build a cloud image and test it somehow.

Maybe it is time to document some guidelines or rethink how we do things.

larskarlitski:
Pushed to fix the conflict in `test/run`.

> Everything that exists up to now in terms of integration tests doesn't talk to the API directly.

I think having a full integration test of the HTTP API of the fully installed lorax-composer is necessary. I'd like to expand it in the future to test for headers, for example.

PR #742, opened 2019-05-31
test: Split image tests

larskarlitski:
The image tests take quite long. Split them off the default test run and
into two scenarios: images-1 and images-2. They're named generically so
that we can extend them without adding scenarios.

Also fail when the scenario is unknown instead of running the default
tests again.

### DIFF
--- a/test/run
+++ b/test/run
@@ -4,8 +4,28 @@
 
 make vm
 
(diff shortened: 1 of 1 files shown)


larskarlitski:
Closing in favor of #735.

PR #744, opened 2019-05-31
Remove PR template in favor of branch protection rules

atodorov:

at some point in time these checks will be made Required but
for now we rely on human guidance to not merge without having
tests PASS.

Note: I'm also adding a 'no-test' label here to see if cockpituous will respect that.


.github/pull_request_template.md
```diff
@@ -5,17 +5,6 @@
...
-- [ ] Travis CI PASS
-- [ ] `*-aws-runtest` PASS
-- [ ] `*-azure-runtest` PASS
-- [ ] `*-images-runtest` PASS
-- [ ] `*-openstack-runtest` PASS
-- [ ] `*-vmware-runtest` PASS
+- [ ] `Required` CI checks PASS
 - [ ] For `rhel8-*` and `rhel7-*` branches commit log references an approved
   bug in Bugzilla. Do not merge if the bug doesn't have the 3 ACKs set to `+`!
-
```
bcl: With the new CI system are there any commands we can use in the PR? Or do we have to re-trigger them using cmdline tools?

atodorov: @martinpitt, @larskarlitski ^^^

@bcl I know only about cli tools for now.

martinpitt: Tests can be retriggered only with CLI tools right now.

martinpitt:
To be honest, I'd drop this PR form completely. If you want to enforce that a set of tests is green (which makes a lot of sense!), configure that directly in the GitHub branch protection rules.

larskarlitski:
> To be honest, I'd drop this PR form completely. If you want to enforce that a set of tests is green (which makes a lot of sense!), configure that directly in the GitHub branch protection rules.

I agree. We do have the branch protection rules set up (see the "Required" labels next to some of the tests).

atodorov:
Updated PR to remove the template then.

larskarlitski:
Tests failure is unrelated: this patch removes a file that's only relevant to GitHub.

PR #745, opened 2019-05-31
Changes related to cloud testing in Cockpit CI

atodorov:
otherwise make thinks credentials file doesn't exist and
doesn't include it in the VM image causing all cloud tests to
fail.

### DIFF
--- a/tests/test_cli.sh
+++ b/tests/test_cli.sh
@@ -20,6 +20,12 @@ function setup_tests {
         {print}" \
         $share_dir/composer/live-iso.ks
 
+    # do a backup of the original blueprints directory and get rid of the git
+    # directory (otherwise all of the initial changes in blueprints would have
+    # to be done using blueprints push)
+    cp -r $blueprints_dir ${blueprints_dir}.orig
+    rm -rf $blueprints_dir/git
+
     # append a section with additional option on kernel command line to example-http-server blueprint
     # which is used for building of most of the images
     cat >> $blueprints_dir/example-http-server.toml << __EOF__
@@ -30,7 +36,12 @@ __EOF__
 }
 
 function teardown_tests {
-    mv $1/composer/live-iso.ks.orig $1/composer/live-iso.ks
+    local share_dir=$1
+    local blueprints_dir=$2
+
+    mv $share_dir/composer/live-iso.ks.orig $share_dir/composer/live-iso.ks
+    rm -rf $blueprints_dir
+    mv ${blueprints_dir}.orig $blueprints_dir
 }
 
 # cloud credentials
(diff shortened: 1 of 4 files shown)


atodorov:
> The same in tests/test_cli.sh, where we have the check twice for some reason.

In `test_cli.sh` quotes can stay, bash resolves them just fine. It's make that doesn't grok them.

As for the 2 locations my intention was local testing (directly on a suitable machine running the script) vs. running them inside the VM.

atodorov:
And a question:

the AWS test reports:
> :: [ 10:45:42 ] :: [   FAIL   ] :: AWS_ACCESS_KEY_ID is empty! 
>:: [ 10:45:42 ] :: [   FAIL   ] :: AWS_SECRET_ACCESS_KEY is empty! 

which is what I was seeing before this PR. When running `make vm` locally  and then testing inside the VM I see credentials being configured and test continues further. Question is in CI do we apply the latest changes before running the tests ?

larskarlitski:
> As for the 2 locations my intention was local testing (directly on a suitable machine running the script) vs. running them inside the VM.

Ah sorry. I didn't see that those were two different locations.

atodorov:
This one is ready for review. We can see cloud tests passing. For vmware and openstack we've got a related PR in Cockpit which makes them use internal resources.

martinpitt:
Indeed the [previous vmware test](https://209.132.184.41:8493/logs/pull-745-20190605-105151-a3762178-cockpit-fedora-30-vmware/log) was running on CentOS CI. The cockpit bots PR to route these only to the RH-internal infra landed now, so retries should work.

tests/test_cli.sh
```diff
@@ -23,8 +23,8 @@ function setup_tests {
     # do a backup of the original blueprints directory and get rid of the git
     # directory (otherwise all of the initial changes in blueprints would have
     # to be done using blueprints push)
-    cp -r $BLUEPRINTS_DIR ${BLUEPRINTS_DIR}.orig
-    rm -rf $BLUEPRINTS_DIR/git
+    cp -r $blueprints_dir ${blueprints_dir}.orig
+    rm -rf $blueprints_dir/git
```
martinpitt: Would be nice to squash this into the previous commit to avoid an untidy git history.

tests/test_cli.sh
```diff
@@ -66,8 +69,9 @@ if [ -z "$CLI" ]; then
     ./src/sbin/lorax-composer --sharedir $SHARE_DIR $BLUEPRINTS_DIR &
 else
     export PACKAGE="composer-cli"
+    systemctl stop lorax-composer
     setup_tests /usr/share/lorax /var/lib/lorax/composer/blueprints
-    systemctl restart lorax-composer
+    systemctl start lorax-composer
```
martinpitt: This is an unrelated change, should better be in a separate commit. How does this change behaviour?

atodorov: It is related to the previous commit. I've squashed the two and now it is more evident why we need this change.

atodorov:
> Thanks! The aws failure is "known", i. e. not a regression from this PR?

well we started seeing this only inside Cockpit CI. For some reason aws cli reports credentials as invalid (but only after ansible is able to upload the image to S3) which doesn't happen in Jenkins.

My vote goes for merging and enabling all cloud tests and we'll resolve the AWS issue separately.

PR #747, opened 2019-05-31
Don't send CORS headers

larskarlitski:
These are meant for web applications that are accessed by browsers, not REST APIs.

This includes https://github.com/weldr/lorax/pull/741, because I'm using `check-api` to test reply headers. I'll rebase this pull request when the other one is merged.


larskarlitski:
> I'm ok with it as long as this doesn't break something :) Does cockpit add these headers when it proxy's access to the API?

Hehe it doesn't break anything :). cockpit-composer talks to the composer API via cockpit-bridge on the host itself. The browser never sees it. So there's no need to relax the same origin policy. 

Update to rebase after #741 landed.

PR #749, opened 2019-06-02
Allow failing tests on first assert 

larskarlitski:
Beakerlib upstream can't do this yet, but might at some point:

    https://github.com/beakerlib/beakerlib/issues/42

This is only enabled in combination with the `--sit` option of the
`test/check-*` scripts. It leaves the system in exacly the state it was
in when an assertion failed. Finishing the test run would run cleanup as
well (such as deleting created images). It also takes longer.

### DIFF
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -1,5 +1,21 @@
 #!/usr/bin/env bash
 
+# Monkey-patch beakerlib to exit on first failure if COMPOSER_TEST_FAIL_FAST is
+# set. https://github.com/beakerlib/beakerlib/issues/42
+if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
+  eval "original$(declare -f __INTERNAL_LogAndJournalFail)"
+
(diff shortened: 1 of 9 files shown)


tests/cli/lib/lib.sh
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env bash
 
+# Monkey-patch beakerlib to exit on first failure if COMPOSER_TEST_FAIL_FAST is
+# set. https://github.com/beakerlib/beakerlib/issues/42
+if [ -n "$COMPOSER_TEST_FAIL_FAST" ]; then
```
atodorov: Not a big deal but can we compare if this variable is == 1 or != 0. This will make it more stable in scenarios where you can be doing some debugging or wanting to disable this feature and set the value to anything but 1.

larskarlitski: Yeah, makes sense. I've pushed an update. Thank you.

larskarlitski:
Update to resolve conflict.

PR #715, opened 2019-05-14
Initialize BLUEPRINTS_DIR when testing RPMs

atodorov:
--- Description of proposed changes ---

works around a problem where setup_tests() tries to modify existing
blueprints and can't find them.

Related: rhbz#1698366

CC @jikortus 

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-openstack-runtest` PASS
- [ ] `*-vmware-runtest` PASS
- [ ] For `rhel8-*` and `rhel7-*` branches commit log references an approved
  bug in Bugzilla. Do not merge if the bug doesn't have the 3 ACKs set to `+`!

--- Jenkins commands ---

- `ok to test` to accept this pull request for testing
- `test this please` for a one time test run
- `retest this please` to start a new build

PR #714, opened 2019-05-14
Initialize BLUEPRINTS_DIR when testing RPMs

atodorov:
--- Description of proposed changes ---

works around a problem where setup_tests() tries to modify existing
blueprints and can't find them.

@jikortus FYI. 

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-openstack-runtest` PASS
- [ ] `*-vmware-runtest` PASS
- [ ] For `rhel8-*` and `rhel7-*` branches commit log references an approved
  bug in Bugzilla. Do not merge if the bug doesn't have the 3 ACKs set to `+`!

--- Jenkins commands ---

- `ok to test` to accept this pull request for testing
- `test this please` for a one time test run
- `retest this please` to start a new build

PR #750, opened 2019-06-02
tests: Fail on script errors

### LINKED ISSUE #727: tests: We're not catching errors in test scripts
beakerlib doesn't seem to fail when there's an error in the test script. I'm not sure if we can just `set -e`, because that would kill the test process and might not produce the status report.

To reproduce, call an unknown command in one of the test scripts. I noticed it in https://github.com/weldr/lorax/pull/720

larskarlitski:
Fixes #727

### DIFF
--- a/tests/cli/test_blueprints_sanity.sh
+++ b/tests/cli/test_blueprints_sanity.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 # Note: execute this file from the project root directory
 
+set -e
+
 . /usr/share/beakerlib/beakerlib.sh
 
 CLI="${CLI:-./src/bin/composer-cli}"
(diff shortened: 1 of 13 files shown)


larskarlitski:
First bug found: `$BLUEPRINTS_DIR` was unbound in `test_cli.sh`.

I wonder how the `custom_cmdline_arg` tests were passing?

larskarlitski:
Update: can't use `set -u` when sourcing beakerlib. It has unbound variables. Leaving `-e`.

atodorov:
Obsoletes #714, #715, #716 for the $BLUEPRINTS_DIR change. The test looks OK.

larskarlitski:
> Obsoletes #714, #715, #716 for the $BLUEPRINTS_DIR change. The test looks OK.

Oops sorry I didn't see those.

It looks like the change is breaking something on the Jenkins test runners. Is that why you haven't approved this yet?

By the way, I've opened https://github.com/beakerlib/beakerlib/issues/43.

PR #751, opened 2019-06-02
test: Don't wait on --sit when test succeeds

larskarlitski:


larskarlitski:
Update to resolve conflict.

Issue #696, opened 2019-04-18
blueprints should have a sanity check when pushed


bcl:
We should do a bit more to check the blueprint before accepting a push.
Right now it is just checked to make sure the TOML parser can handle it.
Should check for known problems like using [[foo]] for things that aren't lists.
But we don't want to be so restrictive that it is hard to extend in the future.

And maybe do a depsolve and still accept the blueprint, but return a warning? This will add a delay which may not be acceptable.

bcl:
Fixed by #776

PR #752, opened 2019-06-06
Strip newlines from SSH keys before passing to ks

evan-goode:
OpenSSH puts a newline at the end of public key files, so currently, copying and pasting the entire file into the SSH key field via `xsel -ib < ~/.ssh/id_rsa.pub` screws up the kickstart parse and throws a strange error.

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-vmware-runtest` PASS

--- Jenkins commands ---

### DIFF
--- a/src/pylorax/api/compose.py
+++ b/src/pylorax/api/compose.py
@@ -494,7 +494,7 @@ def write_ks_root(f, user):
 
     # ssh key uses the sshkey kickstart command
     if "key" in user:
-        f.write('sshkey --user %s "%s"\n' % (user["name"], user["key"]))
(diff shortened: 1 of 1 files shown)


atodorov:

bcl:
I don't think this is the right place to fix this. The same could be true of any of the fields in a blueprint, and while we do need better sanity checking on the toml (see issue #696), this specific case seems like it should be handled by the the UI not the backend.

evan-goode:
Makes sense :+1:. I do feel like we should go ahead and address this bug somewhere; I think it's likely to trip people up. I'll take a look at the front end.

PR #754, opened 2019-06-11
WIP: Ignore this - dry-run for Cockpit CI

atodorov:
--- Description of proposed changes ---

I want to see all Cockpit CI test scenarios scheduled automatically and all Jenkins jobs disabled for master branch!

PR #748, opened 2019-05-31
[don't merge] Fix VMware not consistent biosdevname issue

chris1984:
--- Description of proposed changes ---

PR to attempt to fix VMware tests and have the image pick up an IP

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-openstack-runtest` PASS
- [ ] `*-vmware-runtest` PASS
- [ ] For `rhel8-*` and `rhel7-*` branches commit log references an approved
  bug in Bugzilla. Do not merge if the bug doesn't have the 3 ACKs set to `+`!

--- Jenkins commands ---

- `ok to test` to accept this pull request for testing
- `test this please` for a one time test run
- `retest this please` to start a new build

PR #756, opened 2019-06-11
[don't merge] Fix VMware not consistent biosdevname issue

atodorov:

Cherry-pick #748 onto master b/c in (from #754)
https://209.132.184.41:8493/logs/pull-754-20190611-132929-55616c2c-cockpit-fedora-30-vmware/log

we see vmware VM failing to get an IP address.

Opening this PR so we can see the changes tested in the new CI system.

CC @chris1984

### DIFF
--- a/share/composer/vmdk.ks
+++ b/share/composer/vmdk.ks
@@ -5,7 +5,7 @@ firewall --enabled
 
 # NOTE: The root account is locked by default
 # Network information
-network  --bootproto=dhcp --onboot=on --activate
+network  --bootproto=dhcp --onboot=on --activate --device=eth0
 # NOTE: keyboard and lang can be replaced by blueprint customizations.locale settings
 # System keyboard
 keyboard --xlayouts=us --vckeymap=us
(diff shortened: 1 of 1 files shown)


atodorov:
@larskarlitski in the test statuses for https://github.com/weldr/lorax/pull/744 we can see that vmware test passes even without these changes so I'm not sure if we want to pick them up. Surely having eth0 as the NIC sounds more stable but I'm reluctant to introduce changes to something which isn't necessarily broken.

@chris1984 can you comment on this one please. (note: we are yet to enable the new CI for rhel7 and see if the ip issue with vmware vms reproduces there).

larskarlitski:
> @larskarlitski in the test statuses for #744 we can see that vmware test passes even without these changes so I'm not sure if we want to pick them up. Surely having eth0 as the NIC sounds more stable but I'm reluctant to introduce changes to something which isn't necessarily broken.

Ah I didn't realise. Good point. I agree that we shouldn't do this then. We should try to test the images that are as close to the official ones as possible.

atodorov:
closing b/c the vmware issues seems to not be related. If we start seeing consistent failures then we can re-open.

PR #757, opened 2019-06-11
Backport blueprint customizations to rhel7-extras

bcl:


atodorov:
RHBZ 1718473 has finally been approved so I will merge this. The failures in rhel-7-8/tar scenario are unrelated to these changes and further changes to the test suite are made in #897 which also makes use of the customizations features.

PR #759, opened 2019-06-13
[don't merge] test_compose_tar: Fix docker test

jikortus:
The docker phase always failed because `-ti` was passed even though the
the output was not a terminal. Moreover docker service isn't running by
default on RHEL-7, so it's necessary to start it first explicitly.

Also remove the check for /usr/bin/docker in the setup phase. It didn't
test that the daemon was running. More importantly, it didn't abort the
test anwyay (and there doesn't seem to be a good way to do this in
beakerlib).

Related: rhbz#1720224

### DIFF
--- a/tests/cli/test_compose_tar.sh
+++ b/tests/cli/test_compose_tar.sh
@@ -15,6 +15,7 @@ CLI="${CLI:-./src/bin/composer-cli}"
 rlJournalStart
     rlPhaseStartSetup
         rlAssertExists /usr/bin/docker
+        rlRun -t -c "systemctl restart docker"
     rlPhaseEnd
 
     rlPhaseStartTest "compose start"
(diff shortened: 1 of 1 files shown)


tests/cli/test_compose_tar.sh
```diff
@@ -13,10 +13,6 @@ CLI="${CLI:-./src/bin/composer-cli}"
 
 
 rlJournalStart
-    rlPhaseStartSetup
-        rlAssertExists /usr/bin/docker
-    rlPhaseEnd
```
atodorov: Don't remove this

tests/cli/test_compose_tar.sh
```diff
@@ -35,6 +31,8 @@ rlJournalStart
             rlFail "Compose UUID is empty!"
         fi
 
+        rlRun -t -c "systemctl restart docker"
+
```
atodorov: This should be in the setup section if you want to have it.

atodorov:
merged as part of #769

PR #761, opened 2019-06-14
Switch to new toml library

bcl:
The previous library, pytoml from https://github.com/avakar/pytoml is no
longer supported. So this adds a compatibility layer on top of the
suggested replacement, toml from https://github.com/uiri/toml


bcl:
Merged to master with 1 change, Dockerfile.test needed to change python-pytoml to python-toml
rawhide module for libgit is still screwed up so travis test is failing.

PR #765, opened 2019-06-17
Backport changes for Cockpit CI to RHEL 7

atodorov:
RelatedL rhbz#1698366

I did a diff against master and dropped unrelated patches.

@larskarlitski I'm not sure if `check-api` will work on RHEL 7.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,12 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 
 IMAGE_RELEASE = $(shell awk -F: '/FROM/ { print $$2}' Dockerfile.test)
 
+ifeq ($(TEST_OS),)
+TEST_OS = rhel-7-7
+endif
+export TEST_OS
+VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
+
 default: all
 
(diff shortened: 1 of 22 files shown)


atodorov:
@larskarlitski, @martinpitt - from what I can see we fail at sfdisk trying to extend the root partition size. Any advice ?

bcl:
composertest.py needs to be backported to py2, it is using python 3.6 features like f"string", and list expansion with *extra_env

atodorov:
> composertest.py needs to be backported to py2, it is using python 3.6 features like f"string", and list expansion with *extra_env

@bcl thanks for spotting this. updated.

larskarlitski:
> composertest.py needs to be backported to py2, it is using python 3.6 features like f"string", and list expansion with *extra_env

`composertest.py` runs in the test container, not inside the VM. It's the script that spins up the VM itself. There's no need for it to be ported.

atodorov:
> composertest.py runs in the test container, not inside the VM. It's the script that spins up the VM itself. There's no need for it to be ported.

I kept the shebang but didn't revert the rest of the changes b/c pylint will find this file and complain about it!

PR #766, opened 2019-06-17
Enable networking in lorax-composer templates

bcl:
Because anaconda --dirinstall is used the kickstart's network like isn't
processed at all. So we need to remove the NetworkManager-server-config
package which disables networking.

Resolves: rhbz#1710877

### DIFF
--- a/share/composer/ami.ks
+++ b/share/composer/ami.ks
@@ -44,6 +44,8 @@ sed -i 's/cloud-user/ec2-user/' /etc/cloud/cloud.cfg
 %packages
 kernel
 -dracut-config-rescue
+# Enable networking by removing the config file that disables it
+-NetworkManager-config-server
 
 grub2
 
--- a/share/composer/ext4-filesystem.ks
+++ b/share/composer/ext4-filesystem.ks
@@ -33,5 +33,7 @@ touch /etc/machine-id
 %packages --nobase
 # Packages requires to support this output format go here
 selinux-policy-targeted
+# Enable networking by removing the config file that disables it
+-NetworkManager-config-server
 
 # NOTE lorax-composer will add the blueprint packages below here, including the final %end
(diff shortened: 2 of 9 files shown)


atodorov:
@bcl this patch seems to help with the vmware VMs not getting their IP addresses (CC @chris1984) but so far I wasn't able to make qcow2 and live-iso tests working. I am in the process of manually running VMs from a qcow2/live-iso image including your changes to check if we still have the same networking problem or something else.

larskarlitski:
Wow.

> Because anaconda --dirinstall is used the kickstart's network like isn't
processed at all.

Which other commands aren't processed because of this?

The RPM's description says this:

> This package is intended to be installed by default for server
deployments.

Is it safe to just remove it? The official Fedora images include it.

bcl:
> Which other commands aren't processed because of this?

Take a look at pyanaconda/install.py and search for flags.flags.imageInstall and flags.flags.dirInstall
> 
> The RPM's description says this:
> 
> > This package is intended to be installed by default for server
> > deployments.
> 
> Is it safe to just remove it? The official Fedora images include it.

In RHEL7 it is safe, we asked Terry as well. Default network behavior in Fedora and RHEL are different, so I'm not sure why Fedora would be setting `no-auto-default=*` which is what causes the problem.

bcl:
> Scratch my previous comment. I must have executed the tests wrongly. Retried again and I can see that the VM is up and running and we can ssh into it (there are still some errors from ssh asking me for a password but I believe these are unrelated to the issue at hand). Please merge ASAP.

I noticed that when running locally, it's because there is a terminal present so it prompts. When run from the Jenkins system there is no terminal so it spits out the error message w/o prompting. At some point we should fix it so it can be run either way.

larskarlitski:
> In RHEL7 it is safe, we asked Terry as well. Default network behavior in Fedora and RHEL are different, so I'm not sure why Fedora would be setting `no-auto-default=*` which is what causes the problem.

How did this work before? This file hasn't changed in Fedora in [a while](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/commits/master/data/server.conf.in).

PR #768, opened 2019-06-18
Do not generate journal.xml from beakerlib

atodorov:
bacause this requires additional Python modules and we don't
really use it! Fixes
[ WARNING  ] :: cannot create journal.xml due to missing python interpreter


atodorov:
> The xml file only contains the same output that we already see on the console?

yes

larskarlitski:
aws and openstack are expected failures / flaky. They're not marked as required now. I've verified that the patch is unrelated.

PR #772, opened 2019-06-20
Backport repos.git to RHEL8

bcl:


bcl:
@atodorov looks like the builders need to have python3-rpmfluff installed (and probably git as well).

PR #775, opened 2019-06-21
More descriptive error for a bad ref in repos.git

### LINKED ISSUE #771: Misleading error when using repos.git
If you use a non-existent reference in the repos.git the error returned is that it could not clone the repo. The real error is that it could not archive that reference/tag, so the error needs to be improved to make it easier to debug and fix the `ref` entry in the blueprint.

evan-goode:
Fixes #771

I also redirected stderr to stdout in a couple subprocess calls so that git's error messages are captured in the log.

### DIFF
--- a/tests/pylorax/test_gitrpm.py
+++ b/tests/pylorax/test_gitrpm.py
@@ -111,9 +111,10 @@ class GitArchiveTest(unittest.TestCase):
             ref="v1.1.0"
             destination="/srv/testing-rpm/"
         """ % ("/tmp/no-repo-here/"))
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(RuntimeError) as e:
             archive = GitArchiveTarball(git_repo["repos"]["git"][0])
             self._check_tar(archive, "git-rpm-test/", None)
+        self.assertIn("Failed to clone", str(e.exception))
 
     def git_fail_ref_test(self):
         """Test creating an archive from a bad ref"""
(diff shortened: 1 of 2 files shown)


atodorov:

bcl:
Looks good so far, could you also add a test for the archive failure case?

evan-goode:
I may be misunderstanding, but it looks like there already is one:
https://github.com/weldr/lorax/blob/c1aa57022bdc99b8609ae3fc9296302d6420c626/tests/pylorax/test_gitrpm.py#L118

bcl:
> I may be misunderstanding, but it looks like there already is one:

Oh, right, we're using RuntimeError for both. I think it would be good to test for the two failure strings 'Failed to clone' and 'Failed to archive' just to be sure we're telling the users something useful.

tests/pylorax/test_gitrpm.py
```diff
@@ -111,9 +111,10 @@ def git_fail_repo_test(self):
             ref="v1.1.0"
             destination="/srv/testing-rpm/"
         """ % ("/tmp/no-repo-here/"))
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(RuntimeError) as cm:
```
bcl: Small nitpick, sorry, we use 'as e' for error results, not 'cm'. Looks good other than that.

evan-goode: Sure, `e` works.

chris1984:
ok to test

chris1984:
[test]

chris1984:
test

bcl:
Merging this. Tried to force new tests with #778 but infrastructure failed. Version pulled into rhel8-branch ran fine, also passes locally.

PR #777, opened 2019-06-24
Mention python3-magic in HACKING.md

evan-goode:
Some tests fail for me when python3-magic is not installed. Since python3-magic isn't listed as a dependency of lorax in `lorax.spec`, I think it's worth mentioning here.


atodorov:

PR #778, opened 2019-06-24
Fix a couple test related things on master

bcl:
Looks like rawhide has sorted out libgit2 issues, so drop installing the rpm from koji.
rpm now returns str() so drop decoding the header when using it in tests.


larskarlitski:
`cockpit/fedora-30` seems stuck, possibly related to the fedorainfracloud server not responding (@martinpitt is working on getting a more robust solution there). The log shows that all tests ran and the only failure was the one that #779 fixes. Since that one depends on this one, let's merge this.

aws/openstack failures are known issues and unrelated to the changes in this pull request.

PR #779, opened 2019-06-24
test: Fix test_blueprint_sanity

larskarlitski:
The new toml library, introduced with abe7df34f, outputs different
whitespace from the old one. Fix the test expectation and strip()
results from toml.dumps(), because it contains superfluous newlines at
the end.


larskarlitski:
Rebased to include #778, which fixes the travis test.

larskarlitski:
> Looks good, aws test failure is the only one.

Thanks!

PR #781, opened 2019-06-24
WIP: API access to composer, "program", and "dnf" logs

evan-goode:
Let me know whether it would be more appropriate to have different routes to fetch the different logs.

### DIFF
--- a/src/pylorax/api/queue.py
+++ b/src/pylorax/api/queue.py
@@ -16,6 +16,8 @@
 """ Functions to monitor compose queue and run anaconda"""
 import logging
 log = logging.getLogger("pylorax")
+program_log = logging.getLogger("program")
+dnf_log = logging.getLogger("dnf")
 
 import os
 import grp
@@ -34,7 +36,7 @@ from pylorax.api.timestamp import TS_CREATED, TS_STARTED, TS_FINISHED, write_tim
 import pylorax.api.toml as toml
 from pylorax.base import DataHolder
 from pylorax.creator import run_creator
-from pylorax.sysutils import joinpaths
+from pylorax.sysutils import joinpaths, read_tail
 
 def check_queues(cfg):
     """Check to make sure the new and run queue symlinks are correct
@@ -136,6 +138,23 @@ def monitor(cfg):
                 # The symlink may vanish if uuid_cancel() has been called
                 continue
 
+            # The anaconda logs are also copied into ./anaconda/ in this directory
+            os.makedirs(joinpaths(dst, "logs"), exist_ok=True)
+
+            def open_handler(loggers, file_name):
+                handler = logging.FileHandler(joinpaths(dst, "logs", file_name))
+                handler.setLevel(logging.DEBUG)
+                handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s: %(message)s"))
+                for logger in loggers:
+                    logger.addHandler(handler)
+                return (handler, loggers)
+
+            loggers = (((log, program_log, dnf_log), "combined.log"),
+                       ((log,), "composer.log"),
+                       ((program_log,), "program.log"),
+                       ((dnf_log,), "dnf.log"))
+            handlers = [open_handler(loggers, file_name) for loggers, file_name in loggers]
+
             log.info("Starting new compose: %s", dst)
             open(joinpaths(dst, "STATUS"), "w").write("RUNNING\n")
 
@@ -152,6 +171,11 @@ def monitor(cfg):
 #                log.error("Error running compose: %s", e)
                 open(joinpaths(dst, "STATUS"), "w").write("FAILED\n")
                 write_timestamp(dst, TS_FINISHED)
+            finally:
+                for handler, loggers in handlers:
+                    for logger in loggers:
+                        logger.removeHandler(handler)
+                    handler.close()
 
             os.unlink(dst)
 
@@ -178,11 +202,6 @@ def make_compose(cfg, results_dir):
     ks_path = joinpaths(results_dir, "final-kickstart.ks")
     if not os.path.exists(ks_path):
         raise RuntimeError("Missing kickstart file at %s" % ks_path)
-
-    # The anaconda logs are copied into ./anaconda/ in this directory
-    log_dir = joinpaths(results_dir, "logs/")
-    if not os.path.exists(log_dir):
-        os.makedirs(log_dir)
 
     # Load the compose configuration
     cfg_path = joinpaths(results_dir, "config.toml")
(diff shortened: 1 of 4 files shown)


bcl:
I don't really want to expose the log filenames in the API. I know it's a bit of a weak distinction at this point, but the user of the API shouldn't have to know there is more than one logfile.

I think the most common use of it will be to see if the install is making progress, not to examine specific logs (they can always download the metadata to do an analysis of things).

Right now it only returns anaconda.log, which goes away when anaconda finishes, but there is still activity happening. Now that we can log build specific composer.log it would be good to tail that when the anaconda.log output isn't available. What do you think?

evan-goode:
Yeah, I figured we might want to abstract that away. Another common use case will be cockpit-composer users wanting to know why their compose failed. In that situation, and for simply tracking a compose's progress, returning the tail of the most recently updated log could work.

Honestly, I think merging the logs into one could also make sense in these scenarios, it would provide an experience similar to `journalctl -fu lorax-composer`. Like you said, if a user then wanted to examine the logs individually, they could download the metadata.

bcl:
> Honestly, I think merging the logs into one could also make sense in these scenarios

Maybe, or it may make them harder to read. Although it's hard to know for sure because it'll depend on what the failure is. Do you think we could merge them by adding another logger to catch them all before separating them? Or would it require something more complex?

evan-goode:
The composer, program, and dnf logs would be easy, I'd just make another file handler and have them share it. The anaconda log might be trickier since it doesn't seem to go through python's logging.

If instead we wanted to return the "currently active" log, we could read the last line of each log file and return the one with the most recent message.

bcl:
Anaconda is run via subprocess, so yes we don't have direct access to it's log object. And when anaconda is done the logs are moved from /tmp/ into the results directory so it's pretty easy to detect. So maybe this would work: Make a combined logger for composer, output the end of it when there is no /tmp/anaconda.log and output the end of anaconda.log when there is one.

src/pylorax/api/queue.py
```diff
@@ -141,18 +141,22 @@ def queue_sort(uuid):
...
-            handlers = []
-            for logger, log_file_name in zip(loggers, log_file_names):
+            def make_handler(log_file_name):
                 handler = logging.FileHandler(joinpaths(dst, "logs", log_file_name))
                 handler.setLevel(logging.DEBUG)
                 handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s: %(message)s"))
-                logger.addHandler(handler)
+                return handler
+
+            loggers = (log, program_log, dnf_log)
```
bcl: You could just make this a list of tuples and then unpack them instead of using zip() lower down. That would make the relationship between the elements clearer.

src/pylorax/api/v0.py
```diff
@@ -2006,20 +2006,15 @@ def v0_compose_image(uuid):
...
-      is not included. The returned data is raw text from the end of the
-      logfile, starting on a line boundary.
-
-      Possible values for log_file_name are:
-        anaconda.log (default)
-        composer.log
-        program.log
-        dnf.log
+      Returns the end of the anaconda log, or the composer logs if anaconda is
+      not running anaconda log. The size parameter is optional and defaults to
```
bcl: Typo, dangling 'anaconda log'

evan-goode: Oops thanks.

src/pylorax/api/queue.py
evan-goode: Yeah, 083f60b should make this a bit cleaner.

bcl:
Looks pretty good so far, one idea for switching between anaconda.log and packaging.log is to maybe set a timeout? If packaging log is > 15s newer than anaconda use it, otherwise switch back. running stat on both during an install seems to show it writing to both before the actual package installation.

src/pylorax/api/queue.py
```diff
@@ -676,10 +678,24 @@ def uuid_log(cfg, uuid, size=1024):
...
-    else:
-        log_path = joinpaths(uuid_dir, "logs", "combined.log")
+    def get_log_path():
+        # Try to return the most relevant log at any given time during the
+        # compose. If the compose is not running, return the composer log.
+        if status["queue_status"] != "RUNNING":
+            return joinpaths(uuid_dir, "logs", "combined.log")
+        anaconda_log = "/tmp/anaconda.log"
+        packaging_log = "/tmp/packaging.log"
+        if not (os.path.isfile(anaconda_log) and os.path.isfile(packaging_log)):
```
bcl: I think this would be easier to read as `if os.path.isfile(anaconda_log) and not os.path.isfile(packaging_log)`

src/pylorax/api/queue.py
```diff
@@ -676,10 +678,24 @@ def uuid_log(cfg, uuid, size=1024):
...
+            return anaconda_log
+        anaconda_mtime = os.stat(anaconda_log).st_mtime
+        packaging_mtime = os.stat(packaging_log).st_mtime
+        # If the packaging log exists and its last message is at least 15
+        # seconds newer than the anaconda log, return the packaging log.
+        if packaging_mtime > anaconda_mtime + 15:
+            return packaging_log
+        return anaconda_log
+    log_path = get_log_path()
+
```
bcl: Also, this block should be wrapped in a try/except that returns the combined log -- the anaconda logs could vanish at any time.

evan-goode:
Should be more resilient now. Thanks for the suggestions!

src/pylorax/sysutils.py
```diff
@@ -130,3 +130,16 @@ def flatconfig(filename):
...
+
+def read_tail(path, size):
+    """Read up to `size` kibibytes from the end of a file"""
+    with open(path, "r") as f:
+        f.seek(0, 2)
+        end = f.tell()
+        if end < 1024 * size:
+            f.seek(0, 0)
+        else:
+            f.seek(end - (1024 * size))
```
larskarlitski: Calling `f.seek()` with anything except the return value from `f.tell()` is undefined for text files (https://docs.python.org/3/tutorial/inputoutput.html). This will probably work for your case, because you're seeking to the next line below (which should be mentioned in the docstring). It seems smelly to me, though.

To be honest, I'd scrap this whole function and just load the full file into memory as a string above. It's 2019, we can load a MB of text easily.

evan-goode:
As a user, I'd rather read a combined log, but I don't think we can do it without something hacky. We could parse the logs and sort them together, but it wouldn't be very clean since the anaconda log messages don't have full dates, only times. We could merge the logs as they're written with a background subprocess like `tail -F /tmp/anaconda.log /tmp/packaging.log ... > combined.log`, or somehow capture the lorax-composer systemd journal, but I'm not sure it'd be worth the added complexity.

src/pylorax/sysutils.py
evan-goode: Sounds fine to me. We can keep the existing get-last-n-bytes behavior by just assuming 1 byte per character.

bcl: > Calling `f.seek()` with anything except the return value from `f.tell()` is undefined for text files (https://docs.python.org/3/tutorial/inputoutput.html). This will probably work for your case, because you're seeking to the next line below (which should be mentioned in the docstring). It seems smelly to me, though.

Oh geez. Apparently python3 changed that and no longer matches normal fseek behavior. Maybe we should open it as binary then?

> To be honest, I'd scrap this whole function and just load the full file into memory as a string above. It's 2019, we can load a MB of text easily.

The problem with that is python's garbage collector is sometime terrible about getting rid of things. We may be re-reading this same file repeatedly, so I'm worried about excess memory usage. We really ought to be able to properly tail a file in python in 2019 :)

bcl:
I have a patch to read the end of the file in binary mode, I just need to add a bunch of tests to it tomorrow.

src/pylorax/sysutils.py
larskarlitski: > The problem with that is python's garbage collector is sometime terrible about getting rid of things. We may be re-reading this same file repeatedly, so I'm worried about excess memory usage. We really ought to be able to properly tail a file in python in 2019 :)

Oh my. Good points. Let's open it in binary mode then. We just need to make sure that we're on the boundary of a code point, which seeking to the next `\n` will do.

Thanks.

bcl:
Patch is in #811

PR #782, opened 2019-06-25
Backport Cockpit CI changes for RHEL 8

atodorov:


atodorov:
@bcl there's no approved bug for this PR. The closest one I could find is rhbz#1698368. I can use this one if you want or open a new one?

bcl:
> @bcl there's no approved bug for this PR. The closest one I could find is rhbz#1698368. I can use this one if you want or open a new one?

I think we should probably use a new one, there's a lot of change here.

atodorov:
@bcl - live-iso and qcow2 tests fail. We had the same on RHEL7 and this was resolved with the update of lorax templates. Not sure if we need the same thing here.

bcl:
qemu is in a different location, /usr/libexec/qemu-kvm

It failed to start qemu:
/usr/bin/qemu-system-x86_64: No such file or directory

atodorov:
> qemu is in a different location, /usr/libexec/qemu-kvm
> 
> It failed to start qemu:
> /usr/bin/qemu-system-x86_64: No such file or directory

@larskarlitski can you tell us why did we switch to `qemu-system` ? IIRC it was something along the lines that kvm isn't available on all systems in Cockpit CI. 

The qemu-system package on RHEL 7 is part of EPEL, which is still not available for RHEL 8 so we're kind of blocked here.

larskarlitski:
> @larskarlitski can you tell us why did we switch to `qemu-system` ? IIRC it was something along the lines that kvm isn't available on all systems in Cockpit CI.
> 
> The qemu-system package on RHEL 7 is part of EPEL, which is still not available for RHEL 8 so we're kind of blocked here.

Yeah, I switched because `qemu-kvm` and `qemu-system-$(uname -a) -machine accel=kvm:tcg` are equivalent, except that the latter falls back to non-hardware virtualization.
 
qemu is not supported in RHEL, other than using it through libvirt. That's why it is installed to `/usr/libexec/qemu-kvm` instead of `/usr/bin`. There are no guarantees on it, because it's a build made for libvirt. It seems to work differently from `qemu-kvm` in Fedora, in that it doesn't error out when kvm is not installed, falling back to `tcg` instead.

We could use `virt-install` instead. Something like this might work, but hasn't on first try (thanks @martinpitt):
```
virt-install -n c1 --memory 1024 --transient --import --disk <disk>
```

I would say using the qemu from `/usr/libexec` is fine for tests, though.

This is the plan for a RHEL 8 EPEL: https://fedoraproject.org/wiki/Infrastructure_2020/EPEL-8

bcl:
I experimented with ssh settings over here - https://github.com/weldr/lorax/pull/807/commits/167c103614890d968ccccba82fbbb093ad75e367 and seem to have fixed things. I haven't figured out how to add commits to someone else's PR yet, and don't want to scramble things so @atodorov can cherry-pick that commit when he has a chance.

atodorov:
> I haven't figured out how to add commits to someone else's PR yet

the only way you can do this is either send a PR to the forked repository against the appropriate branch or we start opening PRs not from forked repos but from topic branches inside this repo. 

I cherry-picked the 2 commits from #807 and restarted all tests. Let's see what will happen.

larskarlitski:
> the only way you can do this is either send a PR to the forked repository against the appropriate branch or we start opening PRs not from forked repos but from topic branches inside this repo.

You totally can, if the author of the pull request checks "allow maintainers to push to this branch". You're then able to push to the branch in the author's fork directly. Use with caution though :)

bcl:
> push to the branch in the author's fork directly

That's probably what github means by 'Add more commits by pushing to the cockpit_ci_8 branch on atodorov/lorax.' which seems like a good way to mess up his branch, not a good way to have a conversation about patches. I'll stick with making new PRs for now.

And we're still getting `kex_exchange_identification: read: Connection reset by peer` on this branch. What the heck.

bcl:
> I cherry-picked the 2 commits from #807 and restarted all tests. Let's see what will happen.

@atodorov Don't forget the fix the commit message in one of them, still says `WIP:`

atodorov:
@bcl - cherry-picked the commit from #812 and squashed with the previous "debugging commit" and restarted tests

atodorov:
@larskarlitski looks live qcow2 is also being slow due to the software emulation. Should I disable it like live-iso ?

atodorov:
kernel panic seen in https://209.132.184.41:8493/logs/pull-782-20190730-094545-87cf125c-weldr-lorax-rhel-8-1-qcow2/log reported as https://bugzilla.redhat.com/show_bug.cgi?id=1734382.

I will restart this test job to see if we can get a consistent failure or something else.

atodorov:
Latest update: I've skipped both live-iso and qcow2 tests for the moment so we can review and merge the rest of this PR. Will open another one to continue looking into the issues that we're seeing for qcow and live-iso. This should be ready for review now.

tests/cli/test_compose_live-iso.sh
```diff
@@ -13,11 +13,12 @@ set -e
...
 
 CLI="${CLI:-./src/bin/composer-cli}"
-QEMU="/usr/libexec/qemu-kvm"
+QEMU_BIN="/usr/libexec/qemu-kvm"
+QEMU="$QEMU_BIN -machine accel=tcg"
 
 rlJournalStart
     rlPhaseStartSetup
-        rlAssertExists $QEMU
+        rlAssertExists $QEMU_BIN
```
larskarlitski: I don't think this is necessary now that we're skipping those tests.

tests/cli/test_compose_qcow2.sh
```diff
@@ -13,11 +13,12 @@ set -e
 . $(dirname $0)/lib/lib.sh
 
 CLI="${CLI:-./src/bin/composer-cli}"
-QEMU="/usr/libexec/qemu-kvm"
+QEMU_BIN="/usr/libexec/qemu-kvm"
+QEMU="$QEMU_BIN -machine accel=tcg"
```
larskarlitski: I don't think this is necessary now that we're skipping those tests.

test/check-cli
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/python3
 
+from unittest import skip
```
larskarlitski: Please use `import unittest` and the fully qualified name (`unittest.skip`) further down. It's hard for me to find the definition of things when the namespace is not attached to names.

atodorov:
> The huge "backport" commit is a 1:1 backport from master, or are there any changes?

I don't think there are any changes but honestly I've forgotten by now.

atodorov:
All checks passing so merging. I will enable these checks in testmap.py and continue to work on left-overs (skipped tests, etc) that I have in another branch.

PR #783, opened 2019-06-25
test: Output results for cockpit's log.html

larskarlitski:
Cockpit's log.html has a peculiar format, but parsing logs in html is
much easier, because it shows expandable sections for each test.

I'll switch cockpit over to generating links to the html after this lands. Otherwise, we'll put the current logs through log.html, which might garble them.

Example: https://209.132.184.41:8493/logs/pull-12141-20190624-211212-54fc92b5-verify-fedora-30/log.html

### DIFF
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -4,6 +4,7 @@ import argparse
 import os
 import subprocess
 import sys
+import traceback
 import unittest
 
 # import Cockpit's machinery for test VMs and its browser test API
@@ -12,8 +13,6 @@ import testvm # pylint: disable=import-error
 
 
 def print_exception(etype, value, tb):
-    import traceback
-
     # only include relevant lines
     limit = 0
     while tb and '__unittest' in tb.tb_frame.f_globals:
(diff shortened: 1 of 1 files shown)


larskarlitski:
> Looks okay apart from the f-strings. But IMHO it would make sense to also add the log.html bits so that we can actually see the result and verify that the format is good.

Oh I thought passing `--html-output` to `tests-invoke` was all that's needed for that?

martinpitt:
Travis failure:
```
************* Module /lorax/test/composertest.py
W0621(redefined-outer-name): 16,4: print_exception: Redefining name 'traceback' from outer scope (line 7)
W0404(reimported): 16,4: print_exception: Reimport 'traceback' (imported line 7)
```

larskarlitski:
> Travis failure:

Yep, saw it. Fixed in the last push. I also removed all the f-strings :cry:

martinpitt:
The [aws](https://209.132.184.41:8493/logs/pull-783-20190625-180909-70f318f6-cockpit-fedora-30-aws/log) and [openstack](https://209.132.184.41:8493/logs/pull-783-20190625-180909-70f318f6-cockpit-fedora-30-openstack/log) tests failed. These may be network flakes, retrying.

Issue #786, opened 2019-06-27
Also search for pxeboot kernel and initrd


hadess:
I had to apply the following patch so that lorax could find a kernel and initrd in the [aarch64 server DVD ISOs](https://dl.fedoraproject.org/pub/fedora/linux/releases/30/Server/aarch64/iso/).

```patch
diff --git a/src/pylorax/mount.py b/src/pylorax/mount.py
index d8e2b1da..19e399b3 100644
--- a/src/pylorax/mount.py
+++ b/src/pylorax/mount.py
@@ -59,7 +59,8 @@ class IsoMountpoint(object):
             self.mount_dir = self.initrd_path
 
         kernel_list = [("/isolinux/vmlinuz", "/isolinux/initrd.img"),
-                       ("/ppc/ppc64/vmlinuz", "/ppc/ppc64/initrd.img")]
+                       ("/ppc/ppc64/vmlinuz", "/ppc/ppc64/initrd.img"),
+                       ("/images/pxeboot/vmlinuz", "/images/pxeboot/initrd.img")]
 
         if os.path.isdir(self.mount_dir+"/repodata"):
             self.repo = self.mount_dir
```

Let me know if you want an MR for that.

PR #787, opened 2019-06-27
Also search for pxeboot kernel and initrd pairs

hadess:
As used in the Fedora 30 aarch64 Server DVDs.

Closes: #786


atodorov:

PR #790, opened 2019-07-02
Several patches to RHEL7 test suite

atodorov:
on interactive terminal

Related: rhbz#1704209

@bcl I am using a non-related BZ number which has been previously approved.

### DIFF
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -1,11 +1,10 @@
 #!/usr/bin/python3
 
-from __future__ import print_function
-
 import argparse
 import os
 import subprocess
 import sys
+import traceback
 import unittest
 
 # import Cockpit's machinery for test VMs and its browser test API
@@ -14,8 +13,6 @@ import testvm # pylint: disable=import-error
 
 
 def print_exception(etype, value, tb):
-    import traceback
-
     # only include relevant lines
     limit = 0
     while tb and '__unittest' in tb.tb_frame.f_globals:
@@ -29,22 +26,11 @@ class ComposerTestCase(unittest.TestCase):
     image = testvm.DEFAULT_IMAGE
     sit = False
 
-    def __init__(self, methodName='runTest'):
-        super(ComposerTestCase, self).__init__(methodName=methodName)
-        # by default run() does this and defaultTestResult()
-        # always creates new object which is local for the .run() method
-        self.ci_result = self.defaultTestResult()
-
-    def run(self, result=None):
-        # so we override run() and use an object attribute which we can
-        # reference later in tearDown() and extract the errors from
-        super(ComposerTestCase, self).run(result=self.ci_result)
-
     def setUp(self):
         self.network = testvm.VirtNetwork(0)
         self.machine = testvm.VirtMachine(self.image, networking=self.network.host(), memory_mb=2048)
 
-        print("Starting virtual machine '%s'" % self.image)
+        print("Starting virtual machine '{}'".format(self.image))
         self.machine.start()
         self.machine.wait_boot()
 
@@ -64,12 +50,13 @@ class ComposerTestCase(unittest.TestCase):
                                 "--silent",
                                 "--unix-socket", "/run/weldr/api.socket",
                                 "http://localhost/api/status"]
-        r = subprocess.call(self.ssh_command + curl_command, stdout=open(os.devnull, 'w'))
+        r = subprocess.run(self.ssh_command + curl_command, stdout=subprocess.DEVNULL)
         self.assertEqual(r.returncode, 0)
 
     def tearDown(self):
-        # `errors` is a list of tuples (method, error)
-        errors = list(e[1] for e in self.ci_result.errors if e[1])
+        # Peek into internal data structure, because there's no way to get the
+        # TestResult at this point. `errors` is a list of tuples (method, error)
+        errors = list(e[1] for e in self._outcome.errors if e[1])
 
         if errors and self.sit:
             for e in errors:
@@ -81,31 +68,98 @@ class ComposerTestCase(unittest.TestCase):
 
         self.machine.stop()
 
-    def execute(self, command):
-        """Execute a command on the test machine."""
-        return subprocess.call(self.ssh_command + command)
+    def execute(self, command, **args):
+        """Execute a command on the test machine.
+
+        **args and return value are the same as those for subprocess.run().
+        """
+        return subprocess.run(self.ssh_command + command, **args)
 
     def runCliTest(self, script):
-        execute_params = ["CLI=/usr/bin/composer-cli",
+        extra_env = []
+        if self.sit:
+            extra_env.append("COMPOSER_TEST_FAIL_FAST=1")
+
+        r = self.execute(["CLI=/usr/bin/composer-cli",
                           "TEST=" + self.id(),
                           "PACKAGE=composer-cli",
-                          "/tests/test_cli.sh", script]
-        if self.sit:
-            execute_params.insert(0, "COMPOSER_TEST_FAIL_FAST=1")
-
-        r = self.execute(execute_params)
+                          *extra_env,
+                          "/tests/test_cli.sh", script])
         self.assertEqual(r.returncode, 0)
+
+
+class ComposerTestResult(unittest.TestResult):
+    def name(self, test):
+        name = test.id().replace("__main__.", "")
+        if test.shortDescription():
+            name += ": " + test.shortDescription()
+        return name
+
+    def startTest(self, test):
+        super().startTest(test)
+
+        print("# ----------------------------------------------------------------------")
+        print("# ", self.name(test))
+        print("", flush=True)
+
+    def stopTest(self, test):
+        print(flush=True)
+
+    def addSuccess(self, test):
+        super().addSuccess(test)
+        print("ok {} {}".format(self.testsRun, self.name(test)))
+
+    def addError(self, test, err):
+        super().addError(test, err)
+        traceback.print_exception(*err, file=sys.stdout)
+        print("not ok {} {}".format(self.testsRun, self.name(test)))
+
+    def addFailure(self, test, err):
+        super().addError(test, err)
+        traceback.print_exception(*err, file=sys.stdout)
+        print("not ok {} {}".format(self.testsRun, self.name(test)))
+
+    def addSkip(self, test, reason):
+        super().addSkip(test, reason)
+        print("ok {} {} # SKIP {}".format(self.testsRun, self.name(test), reason))
+
+    def addExpectedFailure(self, test, err):
+        super().addExpectedFailure(test, err)
+        print("ok {} {}".format(self.testsRun, self.name(test)))
+
+    def addUnexpectedSuccess(self, test):
+        super().addUnexpectedSuccess(test)
+        print("not ok {} {}".format(self.testsRun, self.name(test)))
+
+
+class ComposerTestRunner(object):
+    """A test runner that (in combination with ComposerTestResult) outputs
+    results in a way that cockpit's log.html can read and format them.
+    """
+
+    def __init__(self, failfast=False):
+        self.failfast = failfast
+
+    def run(self, testable):
+        result = ComposerTestResult()
+        result.failfast = self.failfast
+        result.startTestRun()
+        count = testable.countTestCases()
+        print("1.." + str(count))
+        try:
+            testable(result)
+        finally:
+            result.stopTestRun()
+        return result
 
 
 def print_tests(tests):
     for test in tests:
         if isinstance(test, unittest.TestSuite):
             print_tests(test)
-        # I don't know how this is used when running the tests
-        # (maybe not used from what it looks like) so not sure how to refactor it
-        # elif isinstance(test, unittest.loader._FailedTest):
-        #    name = test.id().replace("unittest.loader._FailedTest.", "")
-        #    print("Error: '%s' does not match a test" % name, file=sys.stderr)
+        elif isinstance(test, unittest.loader._FailedTest):
+            name = test.id().replace("unittest.loader._FailedTest.", "")
+            print("Error: '{}' does not match a test".format(name), file=sys.stderr)
         else:
             print(test.id().replace("__main__.", ""))
 
@@ -129,7 +183,7 @@ def main():
         print_tests(tests)
         return 0
 
-    runner = unittest.TextTestRunner(verbosity=2, failfast=args.sit)
+    runner = ComposerTestRunner(failfast=args.sit)
     result = runner.run(tests)
 
     if tests.countTestCases() != result.testsRun:
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -132,4 +132,8 @@ def main():
     runner = unittest.TextTestRunner(verbosity=2, failfast=args.sit)
     result = runner.run(tests)
 
+    if tests.countTestCases() != result.testsRun:
+        print("Error: unexpected number of tests were run", file=sys.stderr)
+        sys.exit(1)
+
     sys.exit(not result.wasSuccessful())
--- a/test/vm.install
+++ b/test/vm.install
@@ -29,6 +29,22 @@ if ! rpm -q beakerlib; then
     fi
 fi
 
+if rpm -q python2-cryptography; then
+    yum remove -y python2-cryptography
+fi
+
+if rpm -q pyparsing; then
+    yum remove -y pyparsing
+fi
+
+if ! rpm -q git; then
+    yum install -y git
+fi
+
+if ! rpm -q qemu-system-x86; then
+     yum install -y qemu-system-x86
+fi
+
 # Grow root partition to make room for images. This only works on Fedora right now.
 parted --script /dev/vda resizepart 2 100%
 partprobe
--- a/test/vm.install
+++ b/test/vm.install
@@ -2,6 +2,18 @@
 
 SRPM="$1"
 
+LATEST_REPO="/etc/yum.repos.d/rhel7-rel-eng-latest.repo"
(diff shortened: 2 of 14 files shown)


atodorov:
When I execute this by hand I still get:
```
:: [ 11:26:21 ] :: [   FAIL   ] :: root account is disabled in /etc/shadow (Expected 0, got 1)
```

not sure why. But at least ssh doesn't ask for a password.

jikortus:
> When I execute this by hand I still get:
> 
> ```
> :: [ 11:26:21 ] :: [   FAIL   ] :: root account is disabled in /etc/shadow (Expected 0, got 1)
> ```
> 
> not sure why. But at least ssh doesn't ask for a password.

Might this be a real issue? I'd almost say so, at the first sight. I also wonder why the images tests failed (`ssh: connect to host localhost port 2222: Connection refused`) as IIRC this worked for me the last time I ran the tests manually.

Lars's suggestion makes sense - I knew about the `BatchMode` option from the ssh man page, however, for some reason, I thought it would just make the password authentication fail if other auth methods don't succeed. It comes out this assumption was wrong, I just did a quick check and it seems to work just fine for our case.

atodorov:
>>     :: [ 11:26:21 ] :: [   FAIL   ] :: root account is disabled in /etc/shadow (Expected 0, got 1)
> Might this be a real issue? 

Well it doesn't happen in CI, only when I start the script manually. That is odd!

@bcl ^^^ do you have any comment/idea about root account being disabled and why the test script  may be failing sometimes ?

atodorov:
@jikortus good catch:

```
Ran 0 tests in 52.222s

OK
```

@larskarlitski ^^^

atodorov:
@larskarlitski I think I've found the issue with not running tests. I get this locally:

```
$ ssh -o ControlPath=/tmp/ssh-%h-%p-%r-3421 -p 2202 root@127.0.0.2  curl --max-time 360 --silent --unix-socket /run/weldr/api.socket http://localhost/api/status
{
  "api": "0", 
  "backend": "lorax-composer", 
  "build": "19.7.35-1", 
  "db_supported": true, 
  "db_version": "0", 
  "msgs": [
    "Error depsolving live-iso: The following package(s) had problems: grub2-efi-*-cdboot (No package name matching grub2-efi-*-cdboot)"
  ], 
  "schema_version": "0"
}
```

larskarlitski:
The problem is the overriden [`run()`](https://github.com/weldr/lorax/blob/rhel7-extras/test/composertest.py#L38) method, which doesn't return the result. Why did you do this here and not on master? I thought this is for backporting only.

> and exit code is 0 of course .

The above error is unrelated. I don't know if that command should return an error if one of the output types can't be depsolved. I'm tending towards "no", because the other output types are still usable.

test/composertest.py
```diff
@@ -132,4 +132,8 @@ def main():
     runner = unittest.TextTestRunner(verbosity=2, failfast=args.sit)
     result = runner.run(tests)
 
+    if not result.testsRun:
+        print("No tests executed")
+        sys.exit(1)
```
larskarlitski: Running 0 tests is not an error.

atodorov:
> Why did you do this here and not on master? I thought this is for backporting only.

FTR I did this b/c at the time I didn't know that we execute these files with Python 3. On Python 2 `self._outcome` is not available so I tried to expose an attribute which will hold the test results.

I have reverted these changes and it seems that test suite is able to start the shell scripts.

test/composertest.py
```diff
@@ -93,8 +82,8 @@ def runCliTest(self, script):
         if self.sit:
             execute_params.insert(0, "COMPOSER_TEST_FAIL_FAST=1")
 
-        r = self.execute(execute_params)
-        self.assertEqual(r.returncode, 0)
+        returncode = self.execute(execute_params)
+        self.assertEqual(returncode, 0)
```
larskarlitski: `execute` returns a [`CompletedProcess`](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess) instance. It contains the return code in its `returncode` attribute.

test/composertest.py
```diff
@@ -64,12 +53,12 @@ def setUp(self):
                                 "--silent",
                                 "--unix-socket", "/run/weldr/api.socket",
                                 "http://localhost/api/status"]
-        r = subprocess.call(self.ssh_command + curl_command, stdout=open(os.devnull, 'w'))
-        self.assertEqual(r.returncode, 0)
+        returncode = subprocess.call(self.ssh_command + curl_command, stdout=open(os.devnull, 'w'))
+        self.assertEqual(returncode, 0)
```
larskarlitski: Same here (see comment below).

test/composertest.py
atodorov: Nope it doesn't. As part of the original backport the code on the RHEL7 branch now uses `subprocess.call()` which returns an `int`. On master the code uses `subprocess.run()` which returns the `CompletedProcess` instance.

See the notes in the extended comment on `Don't override TestCase.run()` commit. 

I can revert to the same code on master but I will need to add pylint disable at the very least. It will also be a bigger change.

test/composertest.py
larskarlitski: I agree that this is useful, but we really want to check that the number of tests we expected to run equals the number of tests that were run. Something like this:

```python
if tests.countTestCases() != result.testsRun:
     print("Error: unexpected number of tests were run", file=sys.stderr)
     sys.exit(1)
```

test/composertest.py
larskarlitski: That's super confusing. Why change it at all? `subprocess.run()` exists in RHEL 7's python3.

> - I have kept the other backport changes partly due to pylint and
  partly b/c they have been introduced with a single commit and
  require manually to check them out and revert them (these are mostly
  around how we call subprocess-es).

I'm sorry for the extra work, but this is not an excuse to keep it in the code.

atodorov:
current failures:
- cockpit/rhel-7-7 - looks like a time out (restarted)
- AWS - RequestTimeTooSkewed - this is a known issue since several months
- OpenStack & vmware - looks like a proper bug
```
:: [ 18:34:28 ] :: [   FAIL   ] :: root account is disabled in /etc/shadow (Expected 0, got 1)
```
- qcow2 - can't ssh into the machine, however I see the following error from qemu:
```
STDERR: (process:8990): GLib-WARNING **: 18:18:45.890: gmem.c:489: custom memory allocation vtable not supported
STDERR: Cannot set up guest memory 'pc.ram': Cannot allocate memory
```

@bcl, @larskarlitski - tips or suggestions are welcome

test/composertest.py
```diff
@@ -64,12 +53,13 @@ def setUp(self):
                                 "--silent",
                                 "--unix-socket", "/run/weldr/api.socket",
                                 "http://localhost/api/status"]
-        r = subprocess.call(self.ssh_command + curl_command, stdout=open(os.devnull, 'w'))
+        r = subprocess.run(self.ssh_command + curl_command, stdout=subprocess.DEVNULL)  # pylint: disable=no-member
```
larskarlitski: We're running this under python3. There's no need for the `no-member` pylint rule.

test/composertest.py
```diff
@@ -64,12 +53,13 @@ def setUp(self):
...
         self.assertEqual(r.returncode, 0)
 
     def tearDown(self):
-        # `errors` is a list of tuples (method, error)
-        errors = list(e[1] for e in self.ci_result.errors if e[1])
+        # Peek into internal data structure, because there's no way to get the
+        # TestResult at this point. `errors` is a list of tuples (method, error)
+        errors = list(e[1] for e in self._outcome.errors if e[1]) # pylint: disable=no-member
```
larskarlitski: We're running this under python3. There's no need for the `no-member` pylint rule.

larskarlitski:
>     * cockpit/rhel-7-7 - looks like a time out (restarted)

This is indeed a timeout. Timeouts can be recognized by exit code 124 on the GitHub status. After reverting ed3c2fcbedcf84ec283df05af4ec0994543be54e (see my review), the partitioned_disk test runs in ~25 minutes on my laptop. This is similar to what we had before.

>     * AWS - RequestTimeTooSkewed - this is a known issue since several months

@martinpitt can we include something into our tests scripts to set the clock to something that AWS expects?

>     * OpenStack & vmware - looks like a proper bug

Indeed. I'll dig deeper on this in the coming days. Does this happen on master as well?

>     * qcow2 - can't ssh into the machine, however I see the following error from qemu:
> ```
> STDERR: (process:8990): GLib-WARNING **: 18:18:45.890: gmem.c:489: custom memory allocation vtable not supported
> STDERR: Cannot set up guest memory 'pc.ram': Cannot allocate memory
> ```

This happens when qemu cannot allocate enough memory when starting up. Is this only happening here? Do we need 2GB of memory to boot those VMs?

martinpitt:
> AWS - RequestTimeTooSkewed - this is a known issue since several months

Argh, about half of our boxes have skewed clocks again indeed. I now installed a cron.hourly script to sync them, which should help. I'll also ask the admins about un-firewalling NTP, so that we can drop this hilarious hack again.

(FYI:)
```
#!/bin/sh -e
date --set @$(curl -s https://www.currenttimestamp.com/ | sed -n '/current_time =/ { s/^.* = //; s/;//; p }')
```

jikortus:
> > * OpenStack & vmware - looks like a proper bug
> 
> Indeed. I'll dig deeper on this in the coming days. Does this happen on master as well?

This was a bug in the test - too strict regex for checking that the root account is disabled in `/etc/shadow`. Should be fixed for good by #799.

larskarlitski:
> This was a bug in the test - too strict regex for checking that the root account is disabled in `/etc/shadow`. Should be fixed for good by #799.

Thanks! Let's land that and then rebase this one.

test/composertest.py
atodorov: There's pylint running under Python2 (to inspect the source code) which also discovers these files and complains about them. Hence the no-member disable.

test/vm.install
```diff
@@ -17,6 +19,22 @@ if ! rpm -q beakerlib; then
...
+fi
+
+if ! rpm -q git; then
+    yum install -y git
+fi
+
+if ! rpm -q qemu-system-x86; then
+     yum install -y qemu-system-x86
+fi
+
```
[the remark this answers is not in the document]
atodorov: Initially all tests were using `qemu-kvm`, then for master branch you've changed them to use qemu-system. The agreement you are referring to was for the rhel8-branch b/c we don't have qemu-system there. For the rhel7-extras branch I'm just installing the necessary RPM into the VM to minimize the amount of changes in this PR.

atodorov:
The latest test status that I see is:

- cockpit/rhel-7-7 - timeout, probably just takes too long
- cockpit/rhel-7-7/aws - RrequestTimeTooSkewed - known issue since a long time ago, not realted to changes here
- cockpit/rhel-7-7/qcow2 - still failing, even after I downsized the VMs to 1024 MB. It looks to me that either the VMs are very slow to boot (I've bumped sleep to 200 sec) or they don't boot at all (but there are some boot messages in the log) or we still have a networking problem (which was fixed and we've verified it fixed on several occasions)
- cockpit/rhel-7-7/openstack - running ATM, we had some issues with the OpenStack environment today. 

@larskarlitski, @martinpitt - please comment what else I can do here so we can land this PR. As it seems there are still lots of items that need attention.

atodorov:
I've disabled live-iso, aws and qcow2 tests in order to get this ready for review and be able to merge it. Will deal with the rest in another PR. Please review.

test/vm.install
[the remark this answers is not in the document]
larskarlitski: But the same argument as for RHEL 8 applies: `qemu-system-x86` is only available in EPEL.

Now that we're disabling some of the tests, is there even one left that needs qemu?

Please also answer the other two questions above.

tests/cli/test_compose_live-iso.sh
```diff
@@ -14,7 +14,7 @@ set -e
 
 CLI="${CLI:-./src/bin/composer-cli}"
 QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-QEMU="$QEMU_BIN -machine accel=kvm:tcg"
+QEMU="$QEMU_BIN -machine accel=tcg"
```
larskarlitski: Now that we're disabling the tests, we don't need to switch to software emulation.

tests/cli/test_compose_qcow2.sh
```diff
@@ -14,7 +14,7 @@ set -e
 
 CLI="${CLI:-./src/bin/composer-cli}"
 QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-QEMU="$QEMU_BIN -machine accel=kvm:tcg"
+QEMU="$QEMU_BIN -machine accel=tcg"
```
larskarlitski: Now that we're disabling the tests, we don't need to switch to software emulation.

Makefile
```diff
@@ -123,6 +123,7 @@ $(VM_IMAGE): srpm bots
 		--resize 20G \
 		--upload $$srpm:/var/tmp \
 		--upload $(CURDIR)/test/vm.install:/var/tmp/vm.install \
+		--upload $(CURDIR)/test/yum.repos.d:/var/tmp \
```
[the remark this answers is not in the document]
atodorov: My understanding was that all of them need them. @bcl can you tell us which blueprints need the optional and extras repositories on RHEL 7?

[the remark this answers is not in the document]
atodorov: For EPEL - `qemu-system-x86` comes from there. I need to disable the repo and adjust the test scripts to see what else will fail

For `-latest` - we use it to get the latest NVRs when depsolving and also latest of lorax-composer dependencies.

@larskarlitski which repos in particular you don't like and why ?

tests/cli/test_compose_live-iso.sh
atodorov: Not in this pull request. I will drop the commit for now.

atodorov:
> But the same argument as for RHEL 8 applies: qemu-system-x86 is only available in EPEL.

Can't answer directly under than comment so answering in the main thread.

Option 1) Keep using `qemu-system-x86` == we need EPEL, makes tests closer to what we have on master
Option 2) Revert back to `/usr/libexec/qemu-kvm` == scripts slightly different from master, similar on RHEL branches, remove EPEL repo if nothing else needs it.

If we go with 2 then should we apply the same changes on master ?

test/vm.install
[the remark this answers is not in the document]
atodorov: Option 1) Keep using qemu-system-x86 == we need EPEL, makes tests closer to what we have on master

Option 2) Revert back to /usr/libexec/qemu-kvm == scripts slightly different from master, similar on RHEL branches, remove EPEL repo if nothing else needs it.

If we go with 2 then should we apply the same changes on master ?

test/composertest.py
atodorov: There's pylint running under Python2 (to inspect the source code) which also discovers these files and complains about them. Hence the no-member disable.

Makefile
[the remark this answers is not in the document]
bcl: > My understanding was that all of them need them. @bcl can you tell us which blueprints need the optional and extras repositories on RHEL 7?

It's not the blueprints, it's the image types. live-iso needs grub2-efi-*-cdboot from optional.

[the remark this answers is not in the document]
atodorov: @larskarlitski how do you want this to be updated: 

1) Keep uploading the repo files in make vm or
2) Use their base URLs instead, moving optional to the live-iso test script ?

test/composertest.py
larskarlitski: Please don't resolve conversations as the one who is supplying information. The person who asked a question is the one to do that, when the answer does in fact resolve the conversation. (Also, GitHub hides the answer in the UI, which makes it hard to see.)

I don't believe I have to say this, but you cannot check python 3 code with a python 2 linter.

Makefile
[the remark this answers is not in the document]
larskarlitski: As we discussed on irc. Only include the minimal amount of extra repositories, if possible only for single packages. So, (2).

test/composertest.py
atodorov: Actually you can because the linter doesn't know what is the language version you intended these source files to be run with. When resolving the attributes it tries to load the respective module/class and when it doesn't find the desired attributes it complains.

I see a few options here: 

1) Configure the linter to ignore these files or
2) Don't copy them inside the docker image in the first place or
3) Run a linter with Python 3 to also examine these scripts ?

Also about the already changed test modules:

1) Revert everything to be exactly the same with master or
2) Remove the pylint: disable comments and leave the code as-is, which includes some differences that I've made initially

Tell me which one you want to have and I will update the code.

bcl: We should just make it skip the test scripts.

atodorov: I've reverted the changes made to satisfy the linter essentially making these files looks the same as on master branch. I've added an `ignoreNames` attribute to the linter configuration. Will adjust the format if Travis CI reports failure.

Makefile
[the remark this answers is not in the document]
atodorov: I've made addjustments in all places where we use repositories, incl. EPEL and optional. I don't think we need extras so I've removed it for now.

atodorov:
@larskarlitski I did revert `check-api` but I think this also changed it a little bit so I'm not sure what the current failure means. Can you take a look ?

atodorov:
@larskarlitski,
can you review again when you have the time.

Also please check the `check-api` test. I'm still getting connection error. Here's a link to the last execution log before I squashed few unrelated commits:
https://209.132.184.41:8493/logs/pull-790-20190910-121313-c91c034d-weldr-lorax--rhel-7-7/log

PR #792, opened 2019-07-04
Add kickstart tar installation test

jikortus:
Marked as [WIP] as it hasn't been fully tested yet due to [RHBZ1724181](https://bugzilla.redhat.com/show_bug.cgi?id=1724181).

### DIFF
--- a/test/check-cli
+++ b/test/check-cli
@@ -25,9 +25,6 @@ class TestImages(composertest.ComposerTestCase):
     def test_partitioned_disk(self):
         self.runCliTest("/tests/cli/test_compose_partitioned-disk.sh")
 
-    def test_tar(self):
-        self.runCliTest("/tests/cli/test_compose_tar.sh")
-
 
 class TestQcow2(composertest.ComposerTestCase):
     def test_qcow2(self):
@@ -37,6 +34,14 @@ class TestQcow2(composertest.ComposerTestCase):
 class TestLiveIso(composertest.ComposerTestCase):
     def test_live_iso(self):
         self.runCliTest("/tests/cli/test_compose_live-iso.sh")
+
+
+class TestTar(composertest.ComposerTestCase):
+    def test_tar(self):
+        self.runCliTest("/tests/cli/test_compose_tar.sh")
+
+    def test_tar_kickstart(self):
+        self.runCliTest("/tests/cli/test_compose_tar_kickstart.sh")
 
 
 if __name__ == '__main__':
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -16,6 +16,28 @@ if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
   }
 fi
 
+export QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
+export QEMU="$QEMU_BIN -machine accel=kvm:tcg"
+export SSH_PORT=2222
+
+boot_image() {
+    QEMU_BOOT=$1
+    TIMEOUT=$2
+    rlRun -t -c "$QEMU -m 2048 $QEMU_BOOT -nographic -monitor none \
+                 -net user,id=nic0,hostfwd=tcp::$SSH_PORT-:22 -net nic &"
+    # wait for ssh to become ready (yes, http is the wrong protocol, but it returns the header)
+    tries=0
+    until curl -sS -m 15 "http://localhost:$SSH_PORT/" | grep 'OpenSSH'; do
+        tries=$((tries + 1))
+        if [ $tries -gt $TIMEOUT ]; then
+            exit 1
+        fi
+        sleep 1
+        echo "DEBUG: Waiting for ssh become ready before testing ..."
+    done;
+}
+
+
 # a generic helper function unifying the specific checks executed on a running
 # image instance
 verify_image() {
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -46,7 +46,9 @@ verify_image() {
     SSH_OPTS="-o StrictHostKeyChecking=no $3"
     rlLogInfo "verify_image: SSH_OPTS:'$SSH_OPTS' SSH_USER:'$SSH_USER' SSH_MACHINE: '$SSH_MACHINE'"
     check_root_account "$@"
-    check_kernel_cmdline "$@"
+    if [ "$CHECK_CMDLINE" != 0 ]; then
+        check_kernel_cmdline "$@"
+    fi
 }
 
 check_root_account() {
--- a/tests/cli/test_compose_live-iso.sh
+++ b/tests/cli/test_compose_live-iso.sh
@@ -13,8 +13,6 @@ set -e
 . $(dirname $0)/lib/lib.sh
 
 CLI="${CLI:-./src/bin/composer-cli}"
-QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-QEMU="$QEMU_BIN -machine accel=kvm:tcg"
 
 rlJournalStart
     rlPhaseStartSetup
@@ -47,16 +45,12 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "Start VM instance"
-        rlRun -t -c "$QEMU -m 2048 -boot d -cdrom $IMAGE -nographic -monitor none \
-                           -net user,id=nic0,hostfwd=tcp::2222-:22 -net nic &"
-        # 60 seconds timeout at boot menu screen
-        # then media check + boot ~ 30 seconds
-        sleep 120
+        boot_image "-boot d -cdrom $IMAGE" 120
     rlPhaseEnd
 
     rlPhaseStartTest "Verify VM instance"
-        # run generic tests to verify the instance, log in a liveuser with no password
-        ROOT_ACCOUNT_LOCKED=0 verify_image liveuser localhost "-p 2222"
+        # run generic tests to verify the instance
+        ROOT_ACCOUNT_LOCKED=0 verify_image liveuser localhost "-p $SSH_PORT"
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/cli/test_compose_qcow2.sh
+++ b/tests/cli/test_compose_qcow2.sh
@@ -13,8 +13,6 @@ set -e
 . $(dirname $0)/lib/lib.sh
 
 CLI="${CLI:-./src/bin/composer-cli}"
-QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-QEMU="$QEMU_BIN -machine accel=kvm:tcg"
 
 rlJournalStart
     rlPhaseStartSetup
@@ -75,14 +73,12 @@ __EOF__
     rlPhaseEnd
 
     rlPhaseStartTest "Start VM instance"
-        rlRun -t -c "$QEMU -m 2048 -boot c -hda $IMAGE -nographic -monitor none \
-                           -net user,id=nic0,hostfwd=tcp::2222-:22 -net nic &"
-        sleep 60
+        boot_image "-boot c -hda $IMAGE" 60
     rlPhaseEnd
 
     rlPhaseStartTest "Verify VM instance"
         # run generic tests to verify the instance
-        verify_image root localhost "-i $SSH_KEY_DIR/id_rsa -p 2222"
+        verify_image root localhost "-i $SSH_KEY_DIR/id_rsa -p $SSH_PORT"
     rlPhaseEnd
 
     rlPhaseStartCleanup
--- a/tests/cli/test_compose_tar_kickstart.sh
+++ b/tests/cli/test_compose_tar_kickstart.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+#####
+#
+# Build tar image and install it using liveimg kickstart command
+#
+#####
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+CLI="${CLI:-./src/bin/composer-cli}"
+
+rlJournalStart
+    rlPhaseStartSetup
+        rlAssertExists $QEMU_BIN
+        if ! rlCheckRpm httpd; then
+            dnf -y install httpd
+        fi
+        systemctl start httpd
+
+        ks_path="/var/www/html/ks-tar.cfg"
+        tmp_dir=$(mktemp -d /tmp/composer.XXXXX)
+        ssh_key_dir=$(mktemp -d /tmp/composer-ssh-keys.XXXXXX)
+
+        rlRun -t -c "ssh-keygen -t rsa -N '' -f $ssh_key_dir/id_rsa"
+        pub_key=$(cat $ssh_key_dir/id_rsa.pub)
+
+        bp_name="test-tar"
+        blueprint="$bp_name.toml"
+        cat > $blueprint << __EOF__
+name = "$bp_name"
+description = "tar image test"
+version = "0.0.1"
+modules = []
+
+[[packages]]
+name = "openssh-server"
+version = "*"
+
+[[customizations.user]]
+name = "root"
+key = "$pub_key"
+
+__EOF__
+        rlRun -t -c "$CLI blueprints push $blueprint"
+        image_path="/var/www/html/root.tar.xz"
+
+        version=$(awk -F = '$1 == "VERSION_ID" { print $2 }' /etc/os-release | tr -d \")
+        arch=$(uname -m)
+        baseurl=$(curl "https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-${version}&arch=${arch}" | \
+            grep -v "^#" | head -n 1)
+        rlRun -t -c "curl --remote-name-all $baseurl/images/pxeboot/{vmlinuz,initrd.img}"
+
+        rlRun -t -c "fallocate -l 5G disk.img"
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose start"
+        rlAssertEquals "SELinux operates in enforcing mode" "$(getenforce)" "Enforcing"
+        uuid=$($CLI compose start $bp_name liveimg-tar)
+        rlAssertEquals "exit code should be zero" $? 0
+
+        uuid=$(echo $uuid | cut -f 2 -d' ')
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose finished"
+        if [ -n "$uuid" ]; then
+            until $CLI compose info $uuid | grep 'FINISHED\|FAILED'; do
+                sleep 60
+                rlLogInfo "Waiting for compose to finish ..."
+            done;
+            check_compose_status "$UUID"
+        else
+            rlFail "Compose uuid is empty!"
+        fi
+
+        rlRun -t -c "$CLI compose image $uuid"
+        image="$uuid-root.tar.xz"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Install tar image using kickstart liveimg command"
+        cat > $ks_path << __EOF__
+cmdline
+lang en_US.UTF-8
+timezone America/New_York
+keyboard us
+rootpw --lock
+sshkey --username root "$pub_key"
+bootloader --location=mbr
+zerombr
+clearpart --initlabel --all
+autopart
+# reboot is used together with --no-reboot qemu-kvm parameter, which makes the qemu-kvm
+# process exit after the installation is complete and anaconda reboots the system
+# (using 'poweroff' ks command just halted the machine without powering it off)
+reboot
+
+liveimg --url http://10.0.2.2/root.tar.xz
+
+__EOF__
+        mv $image $image_path
+        restorecon $image_path
+        rlLogInfo "Starting installation from tar image in a VM"
+        $QEMU -m 2048 -drive file=disk.img,format=raw -nographic -kernel vmlinuz -initrd initrd.img \
+            -append "inst.ks=http://10.0.2.2/ks-tar.cfg inst.stage2=$baseurl console=ttyS0" --no-reboot
+
+        rlLogInfo "Installation of the image finished."
+    rlPhaseEnd
+
+    rlPhaseStartTest "Boot and check the installed system"
+        boot_image "-drive file=disk.img,format=raw" 600
+        # run generic tests to verify the instance
+        CHECK_CMDLINE=0 verify_image root localhost "-i $ssh_key_dir/id_rsa -p 2222"
+    rlPhaseEnd
+
+    rlPhaseStartCleanup
+        rlRun -t -c "killall -9 $QEMU_BIN"
+        rlRun -t -c "rm -rf $image $blueprint $image_path vmlinuz initrd.img disk.img $ks_path"
+        rlRun -t -c "$CLI blueprints delete $bp_name"
+        rlRun -t -c "$CLI compose delete $uuid"
+        rlRun -t -c "systemctl stop httpd"
+    rlPhaseEnd
+
+rlJournalEnd
+rlJournalPrintText
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,7 @@ test_images:
 	sudo -E ./tests/test_cli.sh tests/cli/test_compose_ext4-filesystem.sh  \
 				    tests/cli/test_compose_partitioned-disk.sh \
 				    tests/cli/test_compose_tar.sh              \
+				    tests/cli/test_compose_tar_kickstart.sh    \
 				    tests/cli/test_compose_qcow2.sh            \
 				    tests/cli/test_compose_live-iso.sh
 


tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
...
+name = "grubby-deprecated"
+version = "*"
+
+[[packages]]
+name = "kernel"
+version = "*"
+
+[[packages]]
+name = "openssh-server"
+version = "7.*"
```
atodorov: `*` is fine for a version. This already differs in downstream branches I think.

tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
...
+
+[[groups]]
+name = "anaconda-tools"
+
+[[customizations.user]]
+name = "root"
+key = "$pub_key"
+__EOF__
+        composer-cli blueprints push $blueprint
```
atodorov: Use "$CLI"

tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
...
+        image="$uuid-root.tar.xz"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Install tar image using kickstart liveimg command"
+        cat >> $ks_path << __EOF__
+%post
+echo -e "curl http://10.0.2.2/image-$uuid-booted\n" >> /etc/rc.d/rc.local
+chmod +x /etc/rc.d/rc.local
+%end
+__EOF__
```
atodorov: You can create the entire kickstart file here, instead of splitting it in 2 parts.

tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
...
+        mv $image $image_path
+        restorecon $image_path
+        $QEMU_BIN -m 2048 -drive file=disk.img,format=raw -nographic -monitor none -net user,id=nic0,hostfwd=tcp::2222-:22 \
+            -net nic -kernel vmlinuz -initrd initrd.img \
+            -append "inst.ks=http://10.0.2.2/ks-tar.cfg inst.stage2=$baseurl console=ttyS0" &
+        timeout=$((60*60)) # wait up to 60 minutes
+        t=0
+        while ((t < timeout)) && ! grep -q "image-$uuid-booted" /var/log/httpd/access_log; do
+            rlLogInfo "Waiting for the tar image to be installed and system to boot up."
+            sleep 10
```
atodorov: IMO you can increase sleeps to 60 everywhere b/c these operations take a long time, though not strictly necessary.

tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
...
+lang en_US.UTF-8
+timezone America/New_York
+keyboard us
+rootpw redhat
+bootloader --location=mbr
+zerombr
+clearpart --initlabel --all
+autopart
+
+liveimg --url http://10.0.2.2/root.tar.xz
```
atodorov: Is this IP address a stable interface ? With libvirtd on RHEL the default is 192.168.122.1, never seen this one. Maybe you can also grab the actual address of the system and use it (if it is not overly complicated). Whichever one is good as long as we know it is stable.

tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
...
+key = "$pub_key"
+__EOF__
+        composer-cli blueprints push $blueprint
+        image_path="/var/www/html/root.tar.xz"
+
+        version=$(awk -F = '$1 == "VERSION_ID" { print $2 }' /etc/os-release)
+        arch=$(uname -m)
+        baseurl=$(curl "https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-${version}&arch=${arch}" | \
+            grep -v "^#" | head -n 1)
+        curl --remote-name-all $baseurl/images/pxeboot/{vmlinuz,initrd.img}
```
atodorov: @bcl I am wondering if lorax-composer keeps a reference (or has an easy access to that reference) to the kernel/initrd used during building the image (i.e. from anaconda)?

If we could grab these from a local directory referenced by the compose UUID it would be much more easier instead of trying to figure out what distro we are running on and where to get these files from.

tests/cli/test_install_tar.sh
larskarlitski: `file:///` urls work as well.

tests/cli/test_install_tar.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+#####
+#
+# Builds tar image and install it using liveimg kickstart command
```
larskarlitski: What exactly is this testing? Is this something people do?

atodorov: We don't have accurate information to answer this question but this scenario was identified when we did the original assessment for Composer and has always been part of our integration test suite. So far we've executed it manually.

atodorov: FTR my understanding is that using tarball images to deploy a system is somewhat common, Brian or Terry may know more.

tests/cli/test_install_tar.sh
atodorov: What do you mean by that ? If I understand you correctly the tarball image will need to be made available via partition on the hard-drive before we do the installation. Not really easier than firing up httpd and getting the IP address right.

tests/cli/test_install_tar.sh
larskarlitski: Thanks, that helps. I don't see the use-case myself, but I'll investigate.

tests/cli/test_install_tar.sh
larskarlitski: Heh, sorry for the brevity. I mistook the for a kickstart we're running from the host, in which case absolute `file://` urls would have worked. This one is passed to a booted system though, so your approach is totally valid. Thanks!

tests/cli/test_install_tar.sh
bcl: If the blueprint includes the kernel the final-kickstart.ks will have the version in it.

bcl:
I saw a comment about IP addresses, and can't tell if it was answered. You can set the subnet to use by passing in `net=192.168.103.0/24` as one of the netdev options.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+        $CLI blueprints push $blueprint
+        image_path="/var/www/html/root.tar.xz"
+
+        version=$(awk -F = '$1 == "VERSION_ID" { print $2 }' /etc/os-release | tr -d \")
+        arch=$(uname -m)
+        baseurl=$(curl "https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-${version}&arch=${arch}" | \
+            grep -v "^#" | head -n 1)
+        curl --remote-name-all $baseurl/images/pxeboot/{vmlinuz,initrd.img}
+
+        fallocate -l 5G disk.img
```
atodorov: surround with `rlRun` so we get a failure status if the system doesn't have enough free disk space

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+
+__EOF__
+        $CLI blueprints push $blueprint
+        image_path="/var/www/html/root.tar.xz"
+
+        version=$(awk -F = '$1 == "VERSION_ID" { print $2 }' /etc/os-release | tr -d \")
+        arch=$(uname -m)
+        baseurl=$(curl "https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-${version}&arch=${arch}" | \
+            grep -v "^#" | head -n 1)
+        curl --remote-name-all $baseurl/images/pxeboot/{vmlinuz,initrd.img}
```
atodorov: IMO use rlRun so we can get errors reported in case of networking issues

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+        uuid=$($CLI compose start $bp_name tar)
+        rlAssertEquals "exit code should be zero" $? 0
+
+        uuid=$(echo $uuid | cut -f 2 -d' ')
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose finished"
+        if [ -n "$uuid" ]; then
+            until $CLI compose info $uuid | grep 'FINISHED\|FAILED'; do
+                sleep 10
```
atodorov: You can bump this sleep to 60 seconds. We don't need to be filling the logs with lots of info messages.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+chmod +x /etc/rc.d/rc.local
+curl http://10.0.2.2/image-$uuid-installed
+%end
+__EOF__
+        mv $image $image_path
+        restorecon $image_path
+        $QEMU_BIN -m 2048 -drive file=disk.img,format=raw -nographic -kernel vmlinuz -initrd initrd.img \
+            -append "inst.ks=http://10.0.2.2/ks-tar.cfg inst.stage2=$baseurl console=ttyS0" --no-reboot &
+
+        timeout=60 # wait up to 60 minutes
```
atodorov: IMO 60 minutes is too much to wait, maybe something like 30 mins to wait for installation. Eitherway it takes around 20-25 min to build the image and the test will timeout when run in CI before the current timeout is exhausted.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+        t=0
+        while ((t < timeout)) && ! grep -q "image-$uuid-installed" /var/log/httpd/access_log; do
+            rlLogInfo "Waiting for the tar image to be installed."
+            sleep 1m
+            ((t += 1))
+        done
+        rlRun -t -c "grep image-$uuid-installed /var/log/httpd/access_log" 0 "Installation of the image finished."
```
atodorov: You can also add a rlLogInfo inside the if condition below to avoid grepping twice for the same thing.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+
+        # rather check explicitly that the installation finished than wait unnecessarily if the installation failed
+        if grep -q "image-$uuid-installed" /var/log/httpd/access_log; then
+            # wait for the installation process to complete and boot the installed system
+            sleep 1m
```
atodorov: Why sleep again ?

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+        while ((t < timeout)) && ! grep -q "image-$uuid-installed" /var/log/httpd/access_log; do
+            rlLogInfo "Waiting for the tar image to be installed."
+            sleep 1m
+            ((t += 1))
+        done
+        rlRun -t -c "grep image-$uuid-installed /var/log/httpd/access_log" 0 "Installation of the image finished."
+
+        # rather check explicitly that the installation finished than wait unnecessarily if the installation failed
+        if grep -q "image-$uuid-installed" /var/log/httpd/access_log; then
+            # wait for the installation process to complete and boot the installed system
```
atodorov: Comment doesn't look correct. We're not waiting for anything here, are we? From what I can see installation mush have finished already.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+            sleep 1m
+            $QEMU_BIN -m 2048 -drive file=disk.img,format=raw -nographic -net user,id=nic0,hostfwd=tcp::2222-:22 \
+                -net nic &
+            t=0
+            timeout=10
+            while ((t < timeout)) && ! grep -q "image-$uuid-booted" /var/log/httpd/access_log; do
+                rlLogInfo "Waiting for the installed system to boot."
+                sleep 1m
+                ((t += 1))
+            done
```
atodorov: Cherry-pick this commit
https://github.com/weldr/lorax/pull/819/commits/2a23b36150576fffd8c6d6fbad1ddbb0b48d12d6

and use the boot_image function

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,158 @@
...
+            rlRun -t -c "grep image-$uuid-booted /var/log/httpd/access_log" 0 "Installed system booted."
```
atodorov: This will become obsolete with the commit from Brian.

test/check-cli
```diff
@@ -28,6 +28,9 @@ class TestImages(composertest.ComposerTestCase):
     def test_tar(self):
         self.runCliTest("/tests/cli/test_compose_tar.sh")
 
+    def test_tar_kickstart(self):
+        self.runCliTest("/tests/cli/test_compose_tar_kickstart.sh")
+
```
atodorov: Currently the fedora-30 scenario times out so I think tar tests should be separated into their own scenario.

atodorov:
> I saw a comment about IP addresses, and can't tell if it was answered. You can set the subnet to use by passing in net=192.168.103.0/24 as one of the netdev options.

@jikortus what about this comment? where does the `10.0.2.2` IP address come from ?

tests/cli/test_compose_tar_kickstart.sh
atodorov: Actually, can you use `cmdline` in kickstart so we can see the progress in the log/terminal and also not start the qemu process in the background. Isn't it possible to block while the installation is running and then shutdown the VM/qemu ? From what I see this is exactly what you are doing (shutting down the VM, then boot it again) so there should be no need to start qemu in the background.

tests/cli/test_compose_tar_kickstart.sh
jikortus: OK, I'll fix this.

tests/cli/test_compose_tar_kickstart.sh
jikortus: OK, I'll fix this.

tests/cli/test_compose_tar_kickstart.sh
jikortus: OK, I'll fix this.

tests/cli/test_compose_tar_kickstart.sh
jikortus: OK, I'll lower the timeout (however in some cases that I observed, the installation took quite a long time (~45 minutes, possibly even more), so I hope that 30 minutes will be enough). I'll also change the installation type to cmdline.

However, as for the VM execution/teardown changes, I'm not sure if I understand what you suggest. The thing is that I have to start the machine first with direct kernel boot, perform the installation (and wait for the installation process to wholly finish and shutdown the machine). I can't just reboot, as it's not possible (or I couldn't find a suitable way) to boot from disk after the reboot. Technically I could overcome this by once booting from network (TFTP/iPXE) and then from the local drive, however I think this would be an unnecessary complication.

And if I ran the VM in the foreground, then I'd have to find some other way to manage for the timeout (I guess I'd have to run the timeout check in background, would this approach help us in any way?).

tests/cli/test_compose_tar_kickstart.sh
jikortus: This is meant to give the installation some time to finish correctly (and in turn the qemu-kvm process as well), before we can safely start the second qemu-kvm process (with boot from local drive).

tests/cli/test_compose_tar_kickstart.sh
jikortus: Good idea, thanks.

tests/cli/test_compose_tar_kickstart.sh
jikortus: Yes, I could use a better wording here. The installation process is finished at this moment (everything is done, including the %post scripts), but we have to wait for a while for the qemu-kvm process to finish after the installation is complete (as I commented above for the sleep).

tests/cli/test_compose_tar_kickstart.sh
jikortus: Based on our yesterday's conversation, I performed the required changes for the qemu-kvm process to be blocking (and the timeout of the test will be managed by CI).

tests/cli/test_compose_tar_kickstart.sh
jikortus: I believe this is no more relevant, WRT changes made to how the VM is run.

tests/cli/test_compose_tar_kickstart.sh
jikortus: No more relevant as well.

tests/cli/test_compose_tar_kickstart.sh
jikortus: No more relevant as well.

tests/cli/test_compose_tar_kickstart.sh
jikortus: Done.

test/check-cli
jikortus: Done.

tests/cli/test_compose_tar_kickstart.sh
jikortus: > > I saw a comment about IP addresses, and can't tell if it was answered. You can set the subnet to use by passing in net=192.168.103.0/24 as one of the netdev options.
> 
> @jikortus what about this comment? where does the `10.0.2.2` IP address come from ?

Sorry that I missed this comment. 10.0.2.2 is the default host IP address used by the internal qemu-kvm DHCP server (and FTR, 10.0.2.15 is the first guest address in the range), as I found out in the documentation. Of course this can be changed to any custom address range, but I think the defaults should work well for us.

atodorov: OK, if the docs say this is the default it's fine.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,133 @@
...
+
+__EOF__
+        mv $image $image_path
+        restorecon $image_path
+        rlLogInfo "Starting installation from tar image in a VM"
+        $QEMU_BIN -m 2048 -drive file=disk.img,format=raw -nographic -kernel vmlinuz -initrd initrd.img \
+            -append "inst.ks=http://10.0.2.2/ks-tar.cfg inst.stage2=$baseurl console=ttyS0" --no-reboot
+
+        rlLogInfo "Installation of the image finished."
+        boot_image "-drive file=disk.img,format=raw" 600
```
atodorov: You can split this into a separate rlPhase to visually distinguish between installation and boot+verify of the VM

tests/cli/test_compose_live-iso.sh
```diff
@@ -13,8 +13,6 @@ set -e
 . $(dirname $0)/lib/lib.sh
 
 CLI="${CLI:-./src/bin/composer-cli}"
-QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-QEMU="$QEMU_BIN -machine accel=kvm:tcg"
```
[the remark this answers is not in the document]
jikortus: OK, I fixed that. However, is this still needed [in general], as we moved to using the separate function for starting the VM? Btw, now that I looked again at the cherry-picked commit, is `export QEMU="/usr/libexec/qemu-kvm"` what we really want on Fedora? Unforntunately I haven't been able yet to go past the installation step in this test on Fedora (bugs during installation on F-30, issues with packages downloading on rawhide), so that's probably why I haven't noticed this sooner.

tests/cli/test_compose_tar_kickstart.sh
jikortus: OK, good idea.

tests/cli/test_compose_live-iso.sh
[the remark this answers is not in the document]
atodorov: I don't see this to have changed. On Fedora test systems there's `qemu-system-x86_64` but no `qemu-kvm` so the qcow2 test fails.

test/check-cli
```diff
@@ -39,5 +36,13 @@ class TestLiveIso(composertest.ComposerTestCase):
         self.runCliTest("/tests/cli/test_compose_live-iso.sh")
 
 
+class TestTarKickstart(composertest.ComposerTestCase):
```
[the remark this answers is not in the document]
jikortus: OK, I missed this one.

jikortus:
Commits squashed. As for the commented `QEMU*` variable lines in `test_compose_live-iso.sh` - I'm not sure what exactly you meant in your comment.

My guess is that the best option would be to remove those lines, use the new `boot_image` function to boot the iso image, and change `$QEMU_BIN` in `tests/cli/lib/lib.sh`  to `qemu-system-$(uname -m)`, just for the master branch (and leave `/usr/libexec/qemu-kvm` for rhel-7 and 8). Is this correct?

tests/cli/lib/lib.sh
```diff
@@ -16,6 +16,27 @@ if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
   }
 fi
 
+export QEMU="qemu-system-$(uname -m)"
```
atodorov: This is missing the additional qemu options from the test scripts before the cherry-pick. Just copy the deleted lines and paste them here.

tests/cli/test_compose_qcow2.sh
```diff
@@ -13,12 +13,10 @@ set -e
 . $(dirname $0)/lib/lib.sh
 
 CLI="${CLI:-./src/bin/composer-cli}"
-QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-QEMU="$QEMU_BIN -machine accel=kvm:tcg"
 
 rlJournalStart
     rlPhaseStartSetup
-        rlAssertExists $QEMU_BIN
+        rlAssertExists $(which $QEMU)
```
atodorov: If you copy the deleted QEMU and QEMU_BIN variables and adjust their values in the cherry-picked commit all of these assert changes will not be needed.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,133 @@
...
+#####
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+CLI="${CLI:-./src/bin/composer-cli}"
+QEMU="qemu-system-$(uname -m)"
+QEMU_CMD="$QEMU -machine accel=kvm:tcg"
```
atodorov: These should not be needed, right? Why not use QEMU which is defined in the shared library ?

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,133 @@
...
+description = "tar image test"
+version = "0.0.1"
+modules = []
+
+[[packages]]
+name = "openssh-server"
+version = "*"
+
+[[groups]]
+name = "anaconda-tools"
```
atodorov: this shouldn't be needed with the new image type

tests/cli/lib/lib.sh
jikortus: OK, I changed the QEMU* variables to conform to the previous state (before Brian's commit), hope it's fine this way.

tests/cli/test_compose_tar_kickstart.sh
jikortus: Not exactly in accordance to the variables from Brian's commit, but I changed the variables to $QEMU_BIN (path to the qemu-system binary) and $QEMU (qemu-system binary with the accelerator parameters), and therefore there is no need to define them anymore in the test script.

tests/cli/test_compose_tar_kickstart.sh
jikortus: I wasn't 100% sure about this one. But if the reason `@anaconda-tools` group needed to be included was `grub2` and `grub2-tools`, then it can be safely removed from the blueprint.

atodorov: well Brian says the new image type will automatically add everything needed so we don't need to add anything extra on our own. Remove the groups section, tests should continue working.

jikortus:
@atodorov OK, then it's fine to remove the `groups` section. As for the `partitioned_disk` test, we'd need the lorax-composer logs to investigate why it failed...

atodorov:
@bcl with the new image type & test we're getting the following error:

> No such file or directory: 'new-kernel-pkg'

Looks like we also need `grubby` in the package list, no ?

bcl:
> Looks like we also need `grubby` in the package list, no ?

No, that should be fixed by the latest anaconda with the liveimg patches. When using reqpart in the template and the newest anaconda there is no need to add extra packages to handle booting, anaconda will do that automatically.

atodorov:
> No, that should be fixed by the latest anaconda with the liveimg patches. When using reqpart in the template and the newest anaconda there is no need to add extra packages to handle booting, anaconda will do that automatically.

so we just need to wait for a newer anaconda then. Any ETA on that ?

bcl:
> so we just need to wait for a newer anaconda then. Any ETA on that ?

Looks like it should already be available in 31.19 and 30.25.6-4

atodorov:
Indeed the updates repo has 30.25.6-4 but the installation process is using anaconda 30.25.6-2.fc30, probably because initrd image has not been refreshed. 

I don't know if this can be refreshed or if we need to switch to Fedora-31 images/repositories for testing (if those are available).

Otherwise this will continue to fail so we've got to skip it until a newer anaconda is available in stage2.

atodorov:
@bcl this is ready for merge. Few notes:

1) `fedora-30` failure is caused by commit 3b6c6a3 which has been reverted in #836
2) The actual kickstart tar test has been skipped b/c Fedora 30 repositories have an older anaconda in stage2 which doesn't seem likely to be refreshed

atodorov:
@bcl, @larskarlitski ping. Is there anything else to do here ? Please review/merge.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,128 @@
...
+# reboot is used together with --no-reboot qemu-kvm parameter, which makes the qemu-kvm
+# process exit after the installation is complete and anaconda reboots the system
+# (using 'poweroff' ks command just halted the machine without powering it off)
+reboot
+
+liveimg --url http://10.0.2.2/root.tar.xz
+
+__EOF__
+        mv $image $image_path
+        restorecon $image_path
```
larskarlitski: Why is this necessary? Please add a comment.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,128 @@
...
+        rlLogInfo "Starting installation from tar image in a VM"
+        $QEMU -m 2048 -drive file=disk.img,format=raw -nographic -kernel vmlinuz -initrd initrd.img \
+            -append "inst.ks=http://10.0.2.2/ks-tar.cfg inst.stage2=$baseurl console=ttyS0" --no-reboot
+
+        rlLogInfo "Installation of the image finished."
```
larskarlitski: How do we check that the installation succeeded?

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,128 @@
...
+    rlPhaseEnd
+
+    rlPhaseStartTest "Boot and check the installed system"
+        boot_image "-drive file=disk.img,format=raw" 600
+        # run generic tests to verify the instance
+        CHECK_CMDLINE=0 verify_image root localhost "-i $ssh_key_dir/id_rsa -p 2222"
+    rlPhaseEnd
+
+    rlPhaseStartCleanup
+        rlRun -t -c "killall -9 $QEMU_BIN"
```
larskarlitski: Why is this necessary? qemu has exited in the previous phase, no?

tests/cli/test_compose_tar_kickstart.sh
atodorov: if qemu doesn't exit here there's probably an error which happened. 

In most cases when we get here installation would have completed successfully, although not 100% guaranteed. The way we actually check for this is a bit later when we try to boot from the freshly installed image and verify some artifacts inside the running VM.

AFAIK there's no easy way to check that the installation process trully completed without any error. Some of the installation tests by RTT collect data during the install and later verify this data when the system boots. If something goes wrong they usually time out.

tests/cli/test_compose_tar_kickstart.sh
jikortus: The built image doesn't have a proper security context for it to be accessed and served by Apache, so the context has to be restored to a proper one in `/var/www/html`.

atodorov:
Rebased for the failure in `fedora-30` scenario. Fixed in 15818421327dfb46d6d74a65f1a617a641fa5431

tests/cli/test_compose_tar_kickstart.sh
jikortus: Yes, it works exactly how Alex described it. I think this approach is sufficient enough, as any issue during installation (traceback, incomplete spoke, some other error) would most probably lead to a timeout of the test, otherwise it would be quite difficult to check if installation goes on well. It would be rather more useful to check the system after installation, if necessary.

tests/cli/test_compose_tar_kickstart.sh
jikortus: No, because `boot_image` runs qemu as a background process, so that it can be checked afterwards. Therefore we need to kill it in the cleanup phase as it's still running.

jikortus:
@larskarlitski Thank you for the review! I rebased to master, but ATM several tests have failed on me, looks like an issue with the test VM image (image-download).

jikortus:
@larskarlitski All of the tests are green after I restarted them, could you please merge this PR if you don't have any further objections or suggestions? :)

PR #793, opened 2019-07-05
Add seperate validation for blueprint names

jkozol:
The VALID_API_STRING function allows for characters that should not be
allowed in blueprint names. VALID_BLUEPRINT_NAME allows us to
specifically check if a blueprint contains a valid name.

### DIFF
--- a/src/pylorax/api/v0.py
+++ b/src/pylorax/api/v0.py
@@ -68,7 +68,7 @@ from pylorax.api.queue import uuid_tar, uuid_image, uuid_cancel, uuid_log
 from pylorax.api.recipes import RecipeError, list_branch_files, read_recipe_commit, recipe_filename, list_commits
 from pylorax.api.recipes import recipe_from_dict, recipe_from_toml, commit_recipe, delete_recipe, revert_recipe
 from pylorax.api.recipes import tag_recipe_commit, recipe_diff, RecipeFileError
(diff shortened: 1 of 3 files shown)


jkozol:
@atodorov I am going to add a test to check-api to check that a blueprint with an invalid name returns the proper error.

larskarlitski:
> @atodorov I am going to add a test to check-api to check that a blueprint with an invalid name returns the proper error.

Thanks! Please add a check specifically for the bug that this fixes. I think a `,` in the blueprint name confuses some routes that take multiple blueprints separated by comma? Also mention it in the commit message, to explain why this change is necessary.

src/pylorax/api/v0.py
```diff
@@ -401,8 +401,8 @@ def v0_blueprints_new():
             blueprint = recipe_from_toml(request.data)
         else:
             blueprint = recipe_from_dict(request.get_json(cache=False))
-
-        if VALID_API_STRING.match(blueprint["name"]) is None:
+
```
bcl: looks like this added some extra whitespace.

larskarlitski:
Travis failure seems unrelated and should be fixed with #797. Please rebase to make sure.

src/pylorax/api/v0.py
jkozol: Thanks for pointing this out. The whitespace should be removed now.

PR #794, opened 2019-07-07
test: check the number of tests that ran

larskarlitski:
Fail if the number of excuted tests != number of dicovered tests.

This is a patch from @atodorov that I backported from #790.


larskarlitski:
Rebased to include the fix for travis (#797).

PR #795, opened 2019-07-08
lorax: Add debug log of command line options

jcpunk:
I noticed livemedia-creator had debug output for `opts` and thought it would be useful for lorax.

I know I'd get some use out of this.


atodorov:

PR #796, opened 2019-07-08
lorax: provide runtime lorax config in debug log

jcpunk:
While debugging a remote lorax session I thought it would be helpful to have the runtime lorax config within the logs.

### DIFF
--- a/src/sbin/lorax
+++ b/src/sbin/lorax
@@ -187,6 +187,9 @@ def main():
     if opts.sharedir:
         lorax.conf.set("lorax", "sharedir", opts.sharedir)
 
+    with open(lorax.conf.get("lorax", "logdir") + '/lorax.conf', 'w') as f:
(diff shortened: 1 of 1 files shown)


atodorov:

jcpunk:
I've taken a quick stab at writing out to a separate file.

src/sbin/lorax
```diff
@@ -187,6 +187,9 @@ def main():
     if opts.sharedir:
         lorax.conf.set("lorax", "sharedir", opts.sharedir)
 
+    with open(lorax.conf.get("lorax", "logdir") + '/lorax.conf', 'w') as _fd:
```
bcl: We normall just use 'as f'

jcpunk: done

jcpunk:
Squashed and renamed.

PR #797, opened 2019-07-08
tests: set skip_if_unavailable in test repos

larskarlitski:
dnf seems to have changed the default for skip_if_unavailable. Some
mock repositories are still around in later tests, which then fail
because metadata cannot be synced.

Also expose skip_if_unavailable in dnf_repo_to_file_repo(), so that
tests checking for equality of repo files continue to pass.


larskarlitski:
> Looks like that fixed it.

Indeed :tada:  Thanks for the quick review.

PR #798, opened 2019-07-09
Don't refresh metadata on startup

larskarlitski:
dnf repositories are configured with metadata_expire, which is related
to the length of time that old versions of the packages stay in the
repository. Thus, dnf should update the cache when necessary.

Also start lorax-composer once in vm.install to include the cache on the
test image. That should make iterating on tests quicker in the common
case.

### DIFF
--- a/src/pylorax/api/dnfbase.py
+++ b/src/pylorax/api/dnfbase.py
@@ -121,10 +121,6 @@ def get_base_object(conf):
     log.info("Using %s for module_platform_id", platform_id)
     dbc.module_platform_id = platform_id
 
-    # Make sure metadata is always current
-    dbc.metadata_expire = 0
-    dbc.metadata_expire_filter = "never"
-
     # write the dnf configuration file
     with open(dnfconf, "w") as f:
         f.write(dbc.dump())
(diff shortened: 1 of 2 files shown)


larskarlitski:
Travis fails because there's a test which checks that metadata is always current. Is that a real-world use case?

bcl:
I'm really unsure of this change. I spent a *lot* of time trying to get the metadata handling working right.
The big issue is that when anaconda runs it always pulls new metadata, so we have to try really hard in composer to keep it updated. We also force an update right before the final depsolve to catch any changes. 

eg. https://github.com/weldr/lorax/blob/master/src/pylorax/api/compose.py#L713

Is this trying to fix a specific problem?

larskarlitski:
> I'm really unsure of this change. I spent a _lot_ of time trying to get the metadata handling working right.
> The big issue is that when anaconda runs it always pulls new metadata, so we have to try really hard in composer to keep it updated. We also force an update right before the final depsolve to catch any changes.

Makes sense, I was missing this context. I had a feeling there was more to it, but it was easier to start the discussion on a PR, because I already had the patch.

> Is this trying to fix a specific problem?

It's mostly an optimization to not have to wait for updating metadata all the time while developing. I'm realizing though that it needs much more thought if this is ever possible. Closing to revisit later. Thanks!

PR #799, opened 2019-07-09
Use a generic regex for disabled root account check

jikortus:
Now the check is (hopefully) as generic as it can be, so we shouldn't see any more issues in the future when the format of the (disabled) password field for root changes in `/etc/shadow` as it already happened in the past.

### DIFF
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -47,7 +47,7 @@ check_root_account() {
         # ssh returns 255 in case of any ssh error, so it's better to grep the specific error message
         rlRun -t -c "ssh $SSH_OPTS -o PubkeyAuthentication=no root@${SSH_MACHINE} 2>&1 | grep -i 'permission denied ('" \
             0 "Can't ssh to '$SSH_MACHINE' as root using password-based auth"
-        rlRun -t -c "ssh $SSH_OPTS ${SSH_USER}@${SSH_MACHINE} \"sudo grep -E '^root:(\*LOCK\*|!):' /etc/shadow\"" \
+        rlRun -t -c "ssh $SSH_OPTS ${SSH_USER}@${SSH_MACHINE} \"sudo passwd --status root | grep -E '^root\s+LK?'\"" \
             0 "root account is disabled in /etc/shadow"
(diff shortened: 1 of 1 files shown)


tests/cli/lib/lib.sh
```diff
@@ -47,7 +47,7 @@ check_root_account() {
         # ssh returns 255 in case of any ssh error, so it's better to grep the specific error message
         rlRun -t -c "ssh $SSH_OPTS -o PubkeyAuthentication=no root@${SSH_MACHINE} 2>&1 | grep -i 'permission denied ('" \
             0 "Can't ssh to '$SSH_MACHINE' as root using password-based auth"
-        rlRun -t -c "ssh $SSH_OPTS ${SSH_USER}@${SSH_MACHINE} \"sudo grep -E '^root:(\*LOCK\*|!):' /etc/shadow\"" \
+        rlRun -t -c "ssh $SSH_OPTS ${SSH_USER}@${SSH_MACHINE} \"sudo grep -E '^root:.*[:;\*!\\].*(:.*){7}' /etc/shadow\"" \
```
larskarlitski: Checking with a regular expression seems error-prone to me. You can run
```
passwd --status root
```
which outputs  `L` or `LK` as its second field for locked accounts. I wouldn't block this PR based on that. Just FYI :)

jikortus: Thanks for pointing this out, I changed the check to use `passwd --status`, it should be more future-proof :).

larskarlitski:
The aws and images tests succeeded on cockpit's infrastructure. Why did they fail on Jenkins. Do we consider those required?

I'd like to land this to unblock other PRs.

bcl:
> The aws and images tests succeeded on cockpit's infrastructure. Why did they fail on Jenkins. Do we consider those required?
> 
> I'd like to land this to unblock other PRs.

Looks like the path to qemu is wrong for jenkins and right for cockpit. I vote for killing (or ignoring) the jenkins results.

atodorov:
Ignore Jenkins. I will turn it off once we land rhel8 Cockpit CI

larskarlitski:
> Ignore Jenkins. I will turn it off once we land rhel8 Cockpit CI

Ok. Landing this. Thanks!

PR #800, opened 2019-07-09
Leave lscpu in the image for additional debugging

jcpunk:
The RHEL8 images added lscpu for additional debugging.  This would be handy in a lot of cases beyond RHEL.

This patch leaves `lscpu` in the image for folks that would like to use it.

Once this is merged I'll send a patch to anaconda to add lscpu output to the debug logs.


atodorov:

PR #801, opened 2019-07-09
Remove .build-id from install media

jcpunk:
The `/usr/$libdir/.build-id` directory seems to only be used with debuginfo.  I'm not sure it makes sense to put these onto the media.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -363,6 +363,9 @@ removepkg cdparanoia-libs opus libtheora libvisual flac-libs gsm avahi-glib avah
 ## metacity requires libvorbis and libvorbisfile, but enc/dec are no longer needed
 removefrom libvorbis --allbut /usr/${libdir}/libvorbisfile.* /usr/${libdir}/libvorbis.*
 
+## Remove build-id links, they are used with debuginfo
+remove /usr/lib/.build-id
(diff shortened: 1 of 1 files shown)


atodorov:

share/templates.d/99-generic/runtime-cleanup.tmpl
```diff
@@ -363,6 +363,9 @@ removepkg cdparanoia-libs opus libtheora libvisual flac-libs gsm avahi-glib avah
 ## metacity requires libvorbis and libvorbisfile, but enc/dec are no longer needed
 removefrom libvorbis --allbut /usr/${libdir}/libvorbisfile.* /usr/${libdir}/libvorbis.*
 
+## Remove build-id links, they are used with debuginfo
+remove /usr/${libdir}/.build-id
```
bcl: ${libdir} needs to be just plain old /lib/

bcl:
After giving this a try I'm kinda on the fence. It's nice to have less things in the image, but in the end the boot.iso is *exactly* the same size. The install.img file ends up being about 300k smaller though.

share/templates.d/99-generic/runtime-cleanup.tmpl
jcpunk: I'll fix and rebase

jcpunk:
In theory the performance should be a bit better with the smaller install.img

PR #805, opened 2019-07-10
tests: Add UserKnownHostsFile to ssh options for tests

bcl:
When running multiple builds, boots, and ssh attempts in one session it
looks like it is having problems with the known_hosts entry even though
strict checking is turned off. This will not update known_hosts so it
should be in a consistent state for each run.

Related: rhbz#1728571


bcl:
Ends up more is needed, see the commits in #807 which should be picked over to #782

PR #810, opened 2019-07-11
Small progress towards non-native builds

hadess:


atodorov:

hadess:
`Make sure -machine is passed to qemu` is untested as I still didn't manage to force this to do anything, but the other two seem to behave as expected.

src/pylorax/installer.py
```diff
@@ -142,6 +142,17 @@ class QEMUInstall(object):
                  "aarch64": "qemu-system-aarch64",
                  "ppc64le": "qemu-system-ppc64"
                 }
+    COMPATIBLE_ARCHS = {"x86_64": [ "x86_64", "i386" ],
```
bcl: Small style thing, no need for spaces between `[` and the contents of the list. We've (I!) have been inconsistent about this, but no spaces looks nicer.

bcl:
I've added some commits to this. I'm unable to actually get aarch64 to work on my x86_64 system but I think it's close? See pr #1108

hadess:
Thanks for taking the time cleaning this up.

PR #811, opened 2019-07-11
Different patch for reading the end of the logfile

bcl:


bcl:
This has been cherry-picked to #781

PR #813, opened 2019-07-17
don't skip Xorg packages on s390x to allow local GUI installation under KVM

sharkcz:
related to https://bugzilla.redhat.com/show_bug.cgi?id=1727029

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -88,9 +88,7 @@ installpkg systemd-sysv systemd-units
 installpkg rsyslog
 
 ## xorg/GUI packages
-%if basearch != "s390x":
 installpkg xorg-x11-drivers xorg-x11-server-Xorg
-%endif
 installpkg xorg-x11-server-utils xorg-x11-xauth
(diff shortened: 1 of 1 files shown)


atodorov:

bcl:
Looks ok to me, but before I merge it we need to:
* make sure the packages are available in the s390 repo (where is it? I thought it was on alt, but can't track it down this morning)
* make sure someone has done a test build with this change
* make sure it doesn't break existing use cases

sharkcz:
> Looks ok to me, but before I merge it we need to:
> 
>     * make sure the packages are available in the s390 repo (where is it? I thought it was on alt, but can't track it down this morning)

my PRs with the changes have been merged, no builds yet, it will be all in primary koji instance
 
>     * make sure someone has done a test build with this change

yes, will do
 
>     * make sure it doesn't break existing use cases

yes, will do

sharkcz:
- the newly required packages (xorg-x11-drivers, tigervnc-server-module) are in the F-31 repos
- lorax successfully produces an image with the proposed change
- anaconda still starts (and runs) fine with the new content in install.img

PR #814, opened 2019-07-18
Add a lorax_test that builds a boot.iso and boots it with qemu

bcl:
Based on the test_cli tests.
More tests can be added in the ./tests/lorax/ directory as run*sh
files.


bcl:
Work in progress, need to hook it up to the cockpit test code in ./test/ which may require a little re-factoring.

bcl:
Right, this is lorax, so it didn't make sense to use the cli stuff as is. I may re-factor things a bit to share the common parts (especially with the cockpit code).

test/run
```diff
@@ -15,4 +15,5 @@ if [ -n "$TEST_SCENARIO" ]; then
 else
   test/check-cli TestImages
   test/check-api
+  test/check-lorax
```
atodorov: How long does this script take ? From what I can tell it builds a boot.iso so you may want to add a new context/TEST_SCENARIO for it instead.

test/check-lorax
```diff
@@ -0,0 +1,22 @@
...
+
+class TestLorax(LoraxTestCase):
+    """
+        This is the "entry-point" to the test suite when
+        executed in Cockpit CI. If $TEST_SCENARIO==""
+        we end up here.
+
+        New test methods should be added here first!
+        When this target becomes too slow we split out into
+        separate scenarios!
```
atodorov: Doc-string copy-paste. Certainly not accurate.

test/check-api
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/python3
 
-import composertest
+from composertest import ComposerTestCase
+from cockpittest import main
```
larskarlitski: This import style makes it hard for me to read and review code, because it's not immediately apparent that a variable is not local to the file. I also have to jump up to the top of the file to find out which module a definition is coming from. Especially for cases like the one further down, which is just called `TestCase` and could come from `cockpittest` or `unittest`.

test/check-lorax
```diff
@@ -0,0 +1,17 @@
...
+from loraxtest import LoraxTestCase
+
+
+class TestLorax(LoraxTestCase):
+    """
+        This is the "entry-point" to the test suite when
+        executed in Cockpit CI.
+    """
+    def test_lorax_bootiso(self):
+        self.runLoraxTest("/tests/lorax/test_lorax_bootiso.sh")
```
larskarlitski: This is modelled after the `tests/cli` tests, which is quite the complex architecture, because it mixes two programming languages and test programs that run outside and inside the VM. I did it this way because I wanted to re-use the existing beakerlib scripts.

Beakerlib doesn't lend itself very well to what we're doing (for example, the ugly hack we have to do to make it stop on first failure).

This approach also requires a rebuild of the test vm every time the test script is changed.

I'd prefer having no extra scripts on the VM and doing all testing from the outside. Do you think that's feasible?

test/composertest.py
```diff
@@ -1,79 +1,20 @@
...
-        print()
+        super(ComposerTestCase, self).setUp()
 
         print("Waiting for lorax-composer to become ready...")
         curl_command = ["curl", "--max-time", "360",
                                 "--silent",
                                 "--unix-socket", "/run/weldr/api.socket",
                                 "http://localhost/api/status"]
-        r = subprocess.run(self.ssh_command + curl_command, stdout=subprocess.DEVNULL)
+        r = self.execute(curl_command, stdout=subprocess.DEVNULL)
```
larskarlitski: Nice catch :)

test/run
```diff
@@ -9,6 +9,8 @@ if [ -n "$TEST_SCENARIO" ]; then
     test/check-cli TestLiveIso
   elif [ "$TEST_SCENARIO" == "qcow2" ]; then
     test/check-cli TestQcow2
+  elif [ "$TEST_SCENARIO" == "lorax" ]; then
+    test/check-lorax
```
larskarlitski: How long does this test need to run? I'm hesitant to add too many statuses, because `make vm` is costly and we have to make a PR to cockpit every time we add a scenario. We only split out some of them because they take a really long time to run.

test/loraxtest.py
```diff
@@ -0,0 +1,16 @@
...
+        extra_env = []
+        if self.sit:
+            extra_env.append("COMPOSER_TEST_FAIL_FAST=1")
+
+        r = self.execute(["CLI=/usr/sbin/lorax",
+                          "TEST=" + self.id(),
+                          "PACKAGE=lorax",
+                          *extra_env,
+                          "/tests/test_lorax.sh", script])
+        self.assertEqual(r.returncode, 0)
```
larskarlitski: This is very similar to `runCliTest`. Both are running a beakerlib script on the test VM, just with different names. Can we implement both with the same function? (This only applies if we don't change approach as I proposed above.)

tests/lorax/test_lorax_bootiso.sh
```diff
@@ -0,0 +1,54 @@
...
+. "$(dirname $0)/lib/lib.sh"
+
+CLI="${CLI:-./src/sbin/lorax}"
+
+
+# Make up a name (slightly unsafe), should not exist before running lorax so use -u
+OUTPUT_DIR=$(mktemp -d -u '/var/tmp/lorax-output.XXXXX')
+rlJournalStart
+    rlPhaseStartSetup
+        rlAssertExists "$QEMU_BIN"
```
larskarlitski: This test is redundant. We know that qemu is installed on the test machines. Also, you'd get an error when trying to execute it later anyway.

test/check-api
bcl: Understood. In longer files I'd agree. These are really short and easy to see.

test/check-lorax
bcl: Maybe? Right now I just want to get it running. We can optimize later. Copying the existing setup seemed like the best way to start.

test/run
bcl: It took about 30 minutes in the last test run before I moved it into a new scenario (because atodorov said it would be better there since it takes longer to run).

test/loraxtest.py
bcl: Sure, just need to pass in a few more args.

tests/lorax/test_lorax_bootiso.sh
bcl: I think atodorov added this because of the qemu name differences in distributions. Better to find out before you waste the time building the iso only to have to boot fail. I don't think it hurts anything to keep this.

test/check-lorax
larskarlitski: Of course, it was probably easier to get going based on the existing code. It shouldn't be too hard to port over later.

tests/lorax/test_lorax_bootiso.sh
larskarlitski: It doesn't hurt indeed, but we will waste the time anyway, because beakerlib doesn't stop a test when something fails.

atodorov: still it says FAIL when you read the test protocol (the output from beakerlib) and that is far more easy to grasp instead of looking for a "command not found" in the raw output.

larskarlitski: It will - of course - also say FAIL when the command cannot be found.

bcl:
I'll revisit this idea after we merge pr #871

PR #818, opened 2019-07-23
Add optional prelink to postinstall

jcpunk:
I took an initial stab at optionally running prelink against the postinstall.


atodorov:

bcl:
What benefit is this going to have? I suspect it will make the image larger, but speed up startup time. I think image size is likely to be more important to users than how fast things start so I'd like to see a before/after size of the various parts (rootfs, squashfs, final iso).

bcl:
I think I'll have to nak this -- we don't even have prelink in Fedora anymore. And depending on it being installed on the host, instead of explicitly selected, can result in surprises.

PR #822, opened 2019-07-25
Add --squashfs-only support to livemedia-creator

bcl:


bcl:
Still needs some work.

bcl:
Merged this to master.

PR #826, opened 2019-07-31
WIP: Cloud upload

evan-goode:
Thought now would be an appropriate time to start a conversation about the progress on cloud upload so far. I'll start by describing the high-level design and then lay out what I think still needs to be done.

I've managed to keep the bulk of the upload logic separate from the rest of lorax. It currently exists as a package called "lifted" in [src/lifted](https://github.com/evan-goode/lorax/tree/lifted/src/lifted). Uploads are [queued and stored ](https://github.com/evan-goode/lorax/blob/e6431430abcd46acf2dd5034663cb09c5fcaf886/src/lifted/queue.py) in a way similar to how composes are. When an upload is scheduled, an [Upload object](https://github.com/evan-goode/lorax/blob/e6431430abcd46acf2dd5034663cb09c5fcaf886/src/lifted/upload.py#L47) is created with the upload's settings and state, and this object gets serialized and written to `/var/lib/lorax/upload/queue/$upload_uuid`. If an upload is scheduled from a compose that isn't finished yet, the upload starts with a status of "WAITING". As soon as the associated compose finishes, the upload gets marked "READY". A [monitor thread](https://github.com/evan-goode/lorax/blob/e6431430abcd46acf2dd5034663cb09c5fcaf886/src/lifted/queue.py#L242) watches for "READY" uploads, kicks each one off as separate process, and marks them as "RUNNING". Since each running upload gets its own PID, we can [cancel an upload](https://github.com/evan-goode/lorax/blob/e6431430abcd46acf2dd5034663cb09c5fcaf886/src/lifted/upload.py#L144) simply by sending it a SIGINT. 

The actual uploading is done using Ansible playbooks and the [Ansible cloud modules](https://docs.ansible.com/ansible/latest/modules/list_of_cloud_modules.html). I currently have uploads to Azure, vSphere, and OpenStack working. [Each cloud provider](https://github.com/evan-goode/lorax/tree/e6431430abcd46acf2dd5034663cb09c5fcaf886/share/lifted/providers) has both a `provider.toml` file containing some metadata (mostly describing the credentials and settings it expects) and a `playbook.yaml`. When a provider's settings get [saved](https://github.com/evan-goode/lorax/blob/e6431430abcd46acf2dd5034663cb09c5fcaf886/src/pylorax/api/v0.py#L2205) for future use, they're written to `/var/lib/lorax/upload/settings/$provider_name.toml`.

The API should now support most of what we'll need on the front end, including:

- Scheduling an upload to run as soon as a compose finishes, or immediately from a finished compose
- Cancelling uploads
- Retrying a failed upload with new settings
- Fetching an upload's Ansible logs
- Validating, storing, and retrieving settings per cloud provider

Some things we still need to work on:

- Gracefully handling missing dependencies: any upload needs `python3-ansible-runner` installed, and right now lorax-composer just crashes if it's missing. Furthermore, each cloud provider has its own set of dependencies. API routes should return some special error if any required package is missing, and Cockpit would ideally offer a one-click install.
- The Ansible cloud modules themselves need a little work:
     - The Azure modules depend on an older version of the Azure Python SDK than is packaged in Fedora.
     - The AWS modules are missing some functionality we need; see https://github.com/ansible/ansible/issues/53453. I'm not sure the solution mentioned there addresses our problem.
     - We'll need to write a couple new modules if we want to support Alibaba Cloud. There isn't anything yet for uploading or importing VM images. 
- Configuration: I just tacked on an "upload" section to the composer config, so let me know if you think there's a more appropriate way to pass paths to lifted.
- Scheduling and managing uploads from composer-cli
- Uploads should be run as an unprivileged user
- Logs should probably not contain sensitive API tokens

For now, the following installations should be all you need to get Azure, OpenStack, and vSphere working:
```
sudo dnf install -y python3-ansible-runner python3-openstacksdk
sudo pip3 install ansible[azure]
```

### DIFF
--- a/src/lifted/queue.py
+++ b/src/lifted/queue.py
@@ -1,0 +1,270 @@
+#
+# Copyright (C) 2019 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from functools import partial
+from glob import glob
+import logging
+import multiprocessing
+
+# We use a multiprocessing Pool for uploads so that we can cancel them with a
+# simple SIGINT, which should bubble down to subprocesses.
+from multiprocessing import Pool
+
+# multiprocessing.dummy is to threads as multiprocessing is to processes.
+# Since daemonic processes can't have children, we use a thread to monitor the
+# upload pool.
+from multiprocessing.dummy import Process
+
+from operator import attrgetter
+import os
+import pickle
+import stat
+import time
+
+from lifted.upload import Upload, UploadStatus
+from lifted.providers import resolve_playbook_path, validate_settings
+
+# the maximum number of simultaneous uploads
+SIMULTANEOUS_UPLOADS = 1
+
+LOG = logging.getLogger("lifted")
+multiprocessing.log_to_stderr().setLevel(logging.INFO)
+
+
+def _get_queue_path(ucfg):
+    path = ucfg["queue_dir"]
+
+    # create the upload_queue directory if it doesn't exist
+    os.makedirs(path, exist_ok=True)
+
+    return path
+
+
+def _get_upload_path(ucfg, uuid, write=False):
+    path = os.path.join(_get_queue_path(ucfg), uuid)
+    if write and not os.path.exists(path):
+        open(path, "a").close()
+    if os.path.exists(path):
+        # make sure uploads aren't readable by others, as they will contain
+        # sensitive credentials
+        current = stat.S_IMODE(os.lstat(path).st_mode)
+        os.chmod(path, current & ~stat.S_IROTH)
+    return path
+
+
+def _list_upload_uuids(ucfg):
+    paths = glob(os.path.join(_get_queue_path(ucfg), "*"))
+    return [os.path.basename(path) for path in paths]
+
+
+def _write_upload(ucfg, upload):
+    with open(_get_upload_path(ucfg, upload.uuid, write=True), "wb") as upload_file:
+        pickle.dump(upload, upload_file, protocol=pickle.HIGHEST_PROTOCOL)
+
+
+def _write_callback(ucfg):
+    return partial(_write_upload, ucfg)
+
+
+def get_upload(ucfg, uuid, ignore_missing=False, ignore_corrupt=False):
+    """Get an Upload object by UUID
+
+    :param ucfg: upload config
+    :type ucfg: object
+    :param uuid: UUID of the upload to get
+    :type uuid: str
+    :param ignore_missing: if True, don't raise a RuntimeError when the
+    specified upload is missing, instead just return None
+    :type ignore_missing: bool
+    :param ignore_corrupt: if True, don't raise a RuntimeError when the
+    :type ignore_corrupt: bool
+    specified upload could not be unpickled, instead just return None
+    :returns: the upload object or None
+    :rtype: Upload or None
+    :raises: RuntimeError
+    """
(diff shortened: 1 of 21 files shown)


bcl:
Great job! This is looking really nice. Here's my list of comments:
* Have just a provider name isn't enough, there will be cases where users will need multiple accounts per-provider.
* I don't like pickle on disk, TOML is what we're using everywhere else -- it makes it way easier to debug things when they get stuck/broken
* When deleting a compose it needs to look for a running upload and return a new error
* style nitpick: LOG should be log like we use in the other modules.
* Do we want to include the upload logs for a compose in that compose's log directory? Might be nice to have it all in one place so you can download it with the metadata API route
* The image types shouldn't be limited to one provider. A default makes sense, but it should be possible to override and upload qcow2 someplace else.
* When removing an upload it should check to make sure it's not running (and whatever other variations on this theme, I probably didn't spot them all)

It also feels like these new API routes would be a good candidate for a v1 API instead of adding them to v0, which would also make it easier for the front-ends to figure out what's supported.

src/lifted/queue.py
```diff
@@ -31,11 +31,12 @@
 
 from operator import attrgetter
 import os
-import pickle
 import stat
 import time
 
-from lifted.upload import Upload, UploadStatus
+import toml
```
bcl: We should use pylorax.api.toml which is our wrapper around the toml library that acts like pytoml did.

src/lifted/queue.py
```diff
@@ -68,12 +69,12 @@ def _get_upload_path(ucfg, uuid, write=False):
...
     paths = glob(os.path.join(_get_queue_path(ucfg), "*"))
-    return [os.path.basename(path) for path in paths]
+    return [os.path.splitext(os.path.basename(path))[0] for path in paths]
 
 
 def _write_upload(ucfg, upload):
-    with open(_get_upload_path(ucfg, upload.uuid, write=True), "wb") as upload_file:
-        pickle.dump(upload, upload_file, protocol=pickle.HIGHEST_PROTOCOL)
+    with open(_get_upload_path(ucfg, upload.uuid, write=True), "w") as upload_file:
+        toml.dump(upload.serialize(), upload_file)
```
bcl: Which of course means you'll have to add dump to pylorax.api.toml :)

src/lifted/queue.py
```diff
@@ -91,22 +92,21 @@ def get_upload(ucfg, uuid, ignore_missing=False, ignore_corrupt=False):
...
-    specified upload could not be unpickled, instead just return None
     :returns: the upload object or None
     :rtype: Upload or None
     :raises: RuntimeError
     """
     try:
-        with open(os.path.join(_get_queue_path(ucfg), uuid), "rb") as pickle_file:
-            return pickle.load(pickle_file)
+        with open(_get_upload_path(ucfg, uuid), "r") as upload_file:
+            return Upload(**toml.load(upload_file))
```
bcl: and load()

src/pylorax/api/v0.py
```diff
@@ -2191,25 +2191,26 @@ def v0_upload_providers():
...
     except RuntimeError as error:
         return jsonify(status=False, errors=[{"id": UPLOAD_ERROR, "msg": str(error)}])
     return jsonify(status=True, provider=provider)
 
-@v0_api.route("/upload/provider/update", defaults={"provider_name": ""}, methods=["POST"])
-@v0_api.route("/upload/provider/update/<provider_name>", methods=["POST"])
-@checkparams([("provider_name", "", "no provider given")])
-def v0_provider_update(provider_name):
+@v0_api.route("/upload/provider/save", defaults={"provider_name": ""}, methods=["POST"])
+@v0_api.route("/upload/provider/save/<provider_name>/<profile>", methods=["POST"])
```
bcl: For our existing POST methods all information comes from the json we pass in, not from the URL. We should stay consistent with what we've done previously.

evan-goode: I can move that stuff to the POST body, but it looks like we already pass variables in the URL in a few places? e.g. https://github.com/evan-goode/lorax/blob/a7cdb46518242fadfdfb659b2f8fd5603c291755/src/pylorax/api/v0.py#L526

bcl: True, but those aren't also passing in a pile of settings. I think we should follow what's been done with creating new blueprints and sources and include it all in the json body.

evan-goode:
Great, thanks! I'll just rebase on top of your v1 branch.

bcl:
See PR #843 for continued work on this.

PR #830, opened 2019-08-02
lorax-composer: Add liveimg-tar image type

bcl:
This creates a tar suitable for use with the anaconda kickstart liveimg
command. It adds the kernel, grub2, and grub2-tools packages to the tar
template.


atodorov:
Related to #792 - will make that test a tad easier.

PR #834, opened 2019-08-19
When launching AWS instances wait for the one we just launched

atodorov:
because the ec2.instances variable will return all that are
currently running

### DIFF
--- a/tests/cli/playbooks/aws/instance.yml
+++ b/tests/cli/playbooks/aws/instance.yml
@@ -22,11 +22,14 @@
       port: 22
       state: started
     with_items: "{{ ec2.instances }}"
+    when: item.image_id == ami_id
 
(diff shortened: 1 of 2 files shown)


atodorov:
Note: Wait for SSH to come up portion of the playbook seems to timeout. Not sure why ATM, need to debug more.

atodorov:
> Note: Wait for SSH to come up portion of the playbook seems to timeout. Not sure why ATM, need to debug more.

found the cause: `allow-ssh` security group has been deleted

atodorov:
@bcl can I get a merge here with the following notes:

1) `fedora-30` failure is caused by commit 3b6c6a3 which has been reverted in #836
2) `fedora-30/tar` is expected failure. The necessary bits are in #792 which is also ready for merge

PR #835, opened 2019-08-19
Use smarter multipath detection logic during OS installation.

dwlehman:
This new setting for 'find_multipaths' tries to prevent things like
LVM from going ahead and activating LVM on the individual disks/paths
until there is reasonable certainty (via a timeout) that the device
is not a component of a multipath set.


bcl:
Looks like it was added in 0.7.7, updated commit message and pushed to master. Thanks!

PR #836, opened 2019-08-20
New test: assert toml files in git workspace

atodorov:
Related: rhbz#1682992


atodorov:
> I don't really see the point in this, if list returns it then we shouldn't care where it is stored. But Ack.

I agree, OTOH this is a direct reproducer/verifier for the referenced bug and b/c it is a small change hence the PR.

atodorov:
@bcl this is ready for merge. Two notes from my side:

1) I've reverted commit 3b6c6a3 b/c it breaks the tests (see `fedora-30` test job)
2) `fedora-30/tar` is expected failure. The necessary bits are in #792 which is also ready for merge

bcl:
Ok, that looks good! Thanks!

PR #839, opened 2019-08-20
Add support for kickstart repo --ssl* options to livemedia-creator

bcl:
Also switches the kickstart support from appending a cpio to the initrd to using an extra disk image.

### DIFF
--- a/src/pylorax/creator.py
+++ b/src/pylorax/creator.py
@@ -39,7 +39,7 @@ from pylorax.base import DataHolder
 from pylorax.executils import execWithRedirect, runcmd
 from pylorax.imgutils import PartitionMount
 from pylorax.imgutils import mount, umount, Mount
-from pylorax.imgutils import mksquashfs, mkrootfsimg
+from pylorax.imgutils import mkcpio, mksquashfs, mkrootfsimg
 from pylorax.imgutils import copytree
 from pylorax.installer import novirt_install, virt_install, InstallError
 from pylorax.treebuilder import TreeBuilder, RuntimeBuilder
@@ -478,6 +478,7 @@ def make_image(opts, ks, cancel_func=None):
     else:
         tar_img = None
 
+    updates_img = []
     try:
         if opts.no_virt:
             novirt_install(opts, disk_img, disk_size, cancel_func=cancel_func, tar_img=tar_img)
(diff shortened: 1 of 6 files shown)


larskarlitski:
Note that @jkonecny12 is in process of deprecating those options. I don't know if it makes sense to add them to livemedia-creator if he does.

jkonecny12:
Thanks @larskarlitski for pointing me on this. 

Yes, I'm going to deprecate the ssl* pykickstart API (I should have do that already). The point is, it should be handled better by repo files so we would avoid extending our API for every thing which is in the repo file. In other words leave the DNF to handle their work.

About this PR. You can use the `ssl*` pykickstart API for your use-case or composer. It will be available for some time but I really want to discourage users to use this API and get used to that. You know how hard is to remove things from pykickstart if users starts to use them. We will then cooperate on switching from this deprecated API to a new solution.

bcl:
I don't think they should be deprecated. Add support for repo files, yes, but it doesn't hurt to keep this around. Especially since it was just recently added.

jkonecny12:
One reason is that we would end up with two ways to handle the SSL stuff - repo files & this. We don't really want to support two implementations here.

Second reason is the benefit of this API. We won't download the certificate and we don't want to do that really. In that case you have to create the updates image (as you do) to get the certificates in and in that case what is the benefit of this solution? For most use-cases (private certificate) you can just install the certificate to the system paths directly the same way and you don't have to use this API at all. If I'm not mistaken than CA certificates have to be placed on specific place to be able to use them.

bcl:
No, it's not automatic. You have to tell dnf what certificates to use.
Since we already have this method, and it works, I don't see a problem with leaving it in place.
How are you proposing to pass in the full repo file? You really can't embed that in the kickstart.

jkonecny12:
> No, it's not automatic. You have to tell dnf what certificates to use.
> Since we already have this method, and it works, I don't see a problem with leaving it in place.

If I'm not wrong, this is not really required for private HTTPS certificates and that is the biggest use-case. The other use-cases with client-cert and client-key certificates are use basically only by the subscription manager which is the only valid use-case for this API I'm aware of. And that is covered already by composer and that is why we leave the API here but we want to switch to repo files in the future. Less amount of projects/customers using this API will result in less problems during the migration process. 

If we would go the current way we can end up with having also API for gpg keys and basically everything which is supported by the repo files right now. We don't want to double the same functionality or anyhow deal with it.

> How are you proposing to pass in the full repo file? You really can't embed that in the kickstart.

That is a great question. We didn't really discuss this yet but I can imagine using updates image (as you do here) or maybe some kind of updates image which will place the files on predefined paths and then it will be automatically loaded by DNF without our intervention. It's not that hard to create a simple tool which will create this image for a customer. However, this is only my suggestion, maybe others in our team will have better idea how to deal with the repo files.

However, I still didn't deprecated this API in upstream so if you have a strong arguments I'm willing to discuss.

bcl:
That sounds like a good plan (we can discuss it on the list), could we get back to reviewing this PR? :)

PR #840, opened 2019-08-26
Test for --no-system-repos option for lorax-composer

jikortus:
The test seems to work fine. One drawback I'm aware of is the lack of `rlRun` use to run the various commands in the new functions in `lib.sh`, as it resulted in the test(s) failing due to unbound variable in beakerlib:
`/usr/share/beakerlib/testing.sh: line 799: $3: unbound variable`

I haven't figured out (yet) why this happens at that stage, but not later on when beakerlib functions are used.

### DIFF
--- a/test/check-cli
+++ b/test/check-cli
@@ -36,6 +36,10 @@ class TestLiveIso(composertest.ComposerTestCase):
     def test_live_iso(self):
         self.runCliTest("/tests/cli/test_compose_live-iso.sh")
 
+class TestRepos(composertest.ComposerTestCase):
+    def test_repos_sanity(self):
+        self.runCliTest("/tests/cli/test_repos_sanity.sh")
+
 
 class TestTar(composertest.ComposerTestCase):
     def test_tar(self):
--- a/test/run
+++ b/test/run
@@ -17,4 +17,5 @@ if [ -n "$TEST_SCENARIO" ]; then
 else
   test/check-cli TestImages
   test/check-api
+  test/check-cli TestRepos
 fi
--- a/tests/cli/test_repos_sanity.sh
+++ b/tests/cli/test_repos_sanity.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+CLI="${CLI:-./src/bin/composer-cli}"
+
+rlJournalStart
+    rlPhaseStartSetup
+        repodir_backup=$(mktemp -d composerrepos-XXXXX)
+        composer_stop
+        rlRun -t -c "mv /var/lib/lorax/composer/repos.d/* $repodir_backup"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Run lorax-composer with --no-system-repos option and empty repos.d"
+        composer_start --no-system-repos
+
+        # check that there are no composer repos available
+        rlRun -t -c "$CLI sources list | grep -v '^$' | wc -l | grep '^0$'"
+        present_repos=$(ls /var/lib/lorax/composer/repos.d)
+        if [ -z "$present_repos" ]; then
+            rlPass "No repos found in repos.d"
+        else
+            rlFail "The following repos were found in repos.d: $present_repos"
+        fi
+
+        # starting a compose without available repos should fail due to a depsolving error
+        rlRun -t -c "tmp_output='$($CLI compose start example-http-server partitioned-disk 2>&1)'"
+        rlRun -t -c "echo '$tmp_output' | grep -q 'Problem depsolving example-http-server:'"
+        MANUAL=1 composer_stop
+    rlPhaseEnd
+
+    rlPhaseStartTest "Run lorax-composer with --no-system-repos and manually created content in repos.d"
+    echo '[fedora]
+name=Fedora $releasever - $basearch
+failovermethod=priority
+metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
+enabled=1
+metadata_expire=7d
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
+skip_if_unavailable=False
+
+[fedora-modular]
+name=Fedora Modular $releasever - $basearch
+failovermethod=priority
+metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-modular-$releasever&arch=$basearch
+enabled=1
+metadata_expire=7d
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
+skip_if_unavailable=False
+
+[updates]
+name=Fedora $releasever - $basearch - Updates
+failovermethod=priority
+metalink=https://mirrors.fedoraproject.org/metalink?repo=updates-released-f$releasever&arch=$basearch
+enabled=1
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
+metadata_expire=6h
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
+skip_if_unavailable=False
+
+[updates-modular]
+name=Fedora Modular $releasever - $basearch - Updates
+failovermethod=priority
+metalink=https://mirrors.fedoraproject.org/metalink?repo=updates-released-modular-f$releasever&arch=$basearch
+enabled=1
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
+metadata_expire=6h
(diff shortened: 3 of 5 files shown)


atodorov:
> /usr/share/beakerlib/testing.sh: line 799: $3: unbound variable

this is because of `set -u` in our own test scripts

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,61 @@
...
+
+        if [[ -z "$CLI" || "$CLI" == "./src/bin/composer-cli" ]]; then
+            composer_params=$(ps ax | grep -v grep  | grep -m 1 -Po 'bin/lorax-composer \K.*')
+        else
+            composer_params="/var/lib/lorax/composer/blueprints"
+        fi
+        composer_stop
+
+        rlRun -t -c "mv /var/lib/lorax/composer/repos.d/* $repodir_backup"
+        rlRun -t -c "mv /etc/yum.repos.d /etc/yum.repos.d.disabled"
```
atodorov: You don't have to touch `/etc/yum.repos.d/` for this test

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,61 @@
...
+    rlPhaseEnd
+
+    rlPhaseStartTest "Run lorax-composer with --no-system-repos option"
```
atodorov: If `/var/lib/lorax/composer/repos.d/` is removed then the next few lines will report PASS.

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,61 @@
...
+        composer_start --no-system-repos $composer_params &
+        wait_for_composer
+        # check that there are no composer repos available
+        rlRun -t -c "$CLI sources list | grep -v '^$' | wc -l | grep '^0$'"
+        rlRun -t -c "$CLI compose start example-http-server qcow2 2>&1 | grep -q \
+            'Problem depsolving example-http-server:'"
```
atodorov: If you don't rename `/etc/yum.repos.d` then you can iterate over all of the .repo files and assert they are not present here. This will make sure the depsolve error isn't due to something else.

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,61 @@
...
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose info $UUID | grep 'FINISHED\|FAILED'; do
+                sleep 10
+                rlLogInfo "Waiting for compose to finish ..."
+            done;
+            check_compose_status "$UUID"
+        else
+            rlFail "Compose UUID is empty!"
+        fi
```
atodorov: I fail to understand what the lines 35-50 are for. From what I can see:

- restore the `/var/lib/lorax/composer/repos.d` directory with its original content, which is all repos from `/etc/yum.repos.d`, which in turn got there when we started lorax-composer (in `test_cli.sh`) before running this script
- then make sure we can build an image which we do in many other places in the test suite so not really needed to do it here.

The only thing in lines 35-50 which I could think off is to try validating that --no-system-repos will not copy files from `/etc/yum.repos.d` which is already done in the snippet above. You may want to verify that existing files are/are not overwritten but that is a completely different test scenario. 

Please explain lines 35-50 to me b/c as they are right now I don't think they are useful.

Makefile
```diff
@@ -77,6 +77,9 @@ test_vmware:
 test_cli:
 	sudo -E ./tests/test_cli.sh
 
+test_repos:
+	sudo -E ./tests/test_cli.sh tests/cli/test_repos_sanity.sh
```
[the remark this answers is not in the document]
jikortus: OK, I'll remove the proposed target then.

jikortus:
@atodorov, as for the unbound variables - I know that this resulted from the `set -u` in `test_cli.sh`. I rather meant that I fail to see what the real root cause is (why the variable is unbound). OTOH if we're fine if the beaker functions are not used in the new composer start/stop/wait functions, it might not be worth spending more time to investigate this.

jikortus:
@atodorov I made some more changest, mostly based on your suggestions. Could you please have a look?

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,114 @@
...
+    rlPhaseEnd
+
+    rlPhaseStartTest "Run lorax-composer with --no-system-repos option and empty repos.d"
+        composer_start --no-system-repos
+
+        # check that there are no composer repos available
+        rlRun -t -c "$CLI sources list | grep -v '^$' | wc -l | grep '^0$'"
+        unexpected_repos_flag=0
+        for repo in /etc/yum.repos.d/*.repo; do
+            if diff $repo /var/lib/lorax/composer/repos.d/${repo##*/}; then
```
atodorov: Why `diff` instead of just asserting this directory is empty ?

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,114 @@
...
+enabled=1
+repo_gpgcheck=0
+type=rpm
+gpgcheck=1
+metadata_expire=6h
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
+skip_if_unavailable=False
+' > /var/lib/lorax/composer/repos.d/test.repo
+
+        composer_start --no-system-repos
```
atodorov: Here you have to assert that nothing else except `test.repo` exists in the target directory. This will verify that nothing has been copied from `/etc/yum.repos.d`

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,114 @@
...
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose info $UUID | grep 'FINISHED\|FAILED'; do
+                sleep 10
+                rlLogInfo "Waiting for compose to finish ..."
+            done;
+            check_compose_status "$UUID"
+        else
+            rlFail "Compose UUID is empty!"
+        fi
```
atodorov: This entire block is repeated in every single script so maybe move into a helper function ? Not strictly necessary here but while you are at it you can do it (also you have another lib updates in this PR).

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,109 @@
...
+        composer_stop
+        rlRun -t -c "mv /var/lib/lorax/composer/repos.d/* $repodir_backup"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Run lorax-composer with --no-system-repos option and empty repos.d"
+        composer_start --no-system-repos
+
+        # check that there are no composer repos available
+        rlRun -t -c "$CLI sources list | grep -v '^$' | wc -l | grep '^0$'"
+        present_repos=$(cd /var/lib/lorax/composer/repos.d && ls)
```
atodorov: Just ls the directory, no need for cd and &&

tests/cli/test_repos_sanity.sh
```diff
@@ -0,0 +1,109 @@
...
+skip_if_unavailable=False
+' > /var/lib/lorax/composer/repos.d/test.repo
+
+        composer_start --no-system-repos
+        present_repos=$(cd /var/lib/lorax/composer/repos.d/ && ls)
+        if [ "$present_repos" == "test.repo" ]; then
+            rlPass "Only test.repo found in repos.d"
+        else
+            rlFail "The following repos were found in repos.d: $present_repos"
+        fi
```
atodorov: Why not `rlAssertFileExists` ?

tests/cli/test_repos_sanity.sh
atodorov: still an issue, no ?

tests/cli/test_repos_sanity.sh
atodorov: still an issue, no ?

test/check-cli
```diff
@@ -35,6 +35,10 @@ class TestLiveIso(composertest.ComposerTestCase):
     def test_live_iso(self):
         self.runCliTest("/tests/cli/test_compose_live-iso.sh")
 
+class TestRepos(composertest.ComposerTestCase):
+    def test_repos_sanity(self):
+        self.runCliTest("/tests/cli/test_repos_sanity.sh")
+
```
atodorov: Needs to be added in `test/run` somehow, otherwise is not being executed.

jikortus:
@atodorov All of the requested changes should be addressed now.

test/run
```diff
@@ -11,6 +11,8 @@ if [ -n "$TEST_SCENARIO" ]; then
     test/check-cli TestQcow2
   elif [ "$TEST_SCENARIO" == "tar" ]; then
     test/check-cli TestTar
+  elif [ "$TEST_SCENARIO" == "repos" ]; then
+    test/check-cli TestRepos
```
atodorov: This should go after line 20, similar to check-api. VM setup is a heavy operation and if we can squeeze the tests in the available window without them timing out then this is what we want to be doing as the first step.

Otherwise we also need to update `testmap.py` which makes the whole process a bit more cumbersome.

tests/cli/test_repos_sanity.sh
atodorov: Just ls the directory, no need for `cd` and `&&`

tests/cli/test_repos_sanity.sh
atodorov: Why not use `rlAssertFileExists` here ?

PR #841, opened 2019-08-28
Skip Azure test b/c misconfigured infra & creds

atodorov:
Also causes failures on the Cockpit side of things, see
https://github.com/cockpit-project/cockpit/pull/12645


larskarlitski:
All green except for the openstack test, which is definitely an unrelated failure (this patch only adds a skip).

PR #842, opened 2019-08-29
Remove unused cloud resources before testing

atodorov:
by default removes everything older than 24 hrs. This is controlled
via the HOURS_LIMIT environment variable.

### DIFF
--- a/.tasks
+++ b/.tasks
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# When run automated, randomize to minimize stampeding herd
+if [ -t 0 ]; then
+    chance=10
+else
+    chance=$(shuf -i 0-10 -n 1)
+fi
+
+if [ $chance -gt 9 ]; then
+    # command lines representing pending work. one will be executed
+    # randomly by the calling environment
+    echo "./tests/cleanup/remove_old_objects_alibaba.sh"
+    echo "./tests/cleanup/remove_old_objects_aws.sh"
+    echo "./tests/cleanup/remove_old_objects_azure.sh"
+    echo "./tests/cleanup/remove_old_objects_openstack.sh"
+    echo "./tests/cleanup/remove_old_objects_vmware.sh"
+fi
--- a/tests/cleanup/remove_old_objects_aws.sh
+++ b/tests/cleanup/remove_old_objects_aws.sh
@@ -79,7 +79,7 @@ for region in $regions; do
       region: "$region"
       state: absent
     loop: "{{vms_facts.instances}}"
-    when: (item.launch_time < lookup('env','TIMESTAMP')) and (item.tags['keep_me'] is not defined)
+    when: (item.launch_time < lookup('env','TIMESTAMP')) and (item.tags['composer-test'] is defined)
     loop_control:
       label: "{{item.instance_id}}"
 __EOF__
@@ -111,7 +111,7 @@ __EOF__
       state: absent
       delete_snapshot: True
     loop: "{{ami_facts.images}}"
-    when: (item.creation_date < lookup('env','TIMESTAMP')) and (item.tags['keep_me'] is not defined)
+    when: (item.creation_date < lookup('env','TIMESTAMP')) and (item.tags['composer-test'] is defined)
     loop_control:
       label: "{{item.image_id}}"
 __EOF__
@@ -121,8 +121,8 @@ __EOF__
     rlPhaseEnd
 
     rlPhaseStartTest "Delete unused composer key pairs in region $region"
-        # list all key pairs starting with "composer-"
-        keys=`aws ec2 describe-key-pairs --region="$region" --query 'KeyPairs[?starts_with(KeyName, \`composer-\`) == \`true\`].KeyName' --output text`
+        # list all key pairs starting with "Composer-Test"
+        keys=`aws ec2 describe-key-pairs --region="$region" --query 'KeyPairs[?starts_with(KeyName, \`Composer-Test\`) == \`true\`].KeyName' --output text`
         rlLogInfo "Found existing composer keys: $keys"
 
         for key in $keys; do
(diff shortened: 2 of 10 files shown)


bcl:
Can you rebase this on the new master (I just merged pull/836), that should make the tests reflect reality :)

atodorov:
> Can you rebase this on the new master

Done. However it looks like these cleanup scripts don't have access to the cloud credentials but the test scripts which are executed afterwards do have access to the appropriate credentials.

atodorov:
> A test run should remove the objects it created, 

All existing tests do this already.

However they time out from time to time and we're left with resources dangling around. Since we're moving away from Jenkins we'd need some sort of a cron system which will trigger these scripts regularly and do the clean up.

> not ones from previous (or other) runs. This seems to be too brittle and prone to accidentally deleting resources we don't want to delete.

Some scripts (but not all I think) skip everything tagged as "keep-me". However this doesn't address the problem that some jobs could leave VMs/files around which need to be removed at some point.

> I don't think this is the right approach. 

Please share other ideas you have. Without a cron based system or a full-blown reservation system like Beaker I'm out of ideas.

atodorov:
@larskarlitski ping. Can you check the last comment and give me some pointers in which direction to do here? 

We either need to:
- clean up everything that is stale before consuming cloud resources (will extend test execution time, may be deleting resources by other people if not tagged "keep-me") or
- have some sort of triggering system (cron in CI would be great) that will execute these clean up scripts and make sure we're not leaving behind anything.

chris1984:
@larskarlitski can I volunteer to do cleanup?

atodorov:
@chris1984 IMO we need a reliable cron to run all of these cleanup scripts. We killed that when we killed Jenkins. IMO feel free to work on it and share suggestions how to implement it.

marusak:
I've seen the PR in cockpit-project/bots so came here to see what it is used for. I think I misunderstood it purpose, can you please explain it to me?
Personally, I find it weird, that you would have **test** that does not run tests, but would do maintenance. It seems like stretching tests a lot. 
When this would be run? Tests are automatically run on every PR. So you want every PR to do this cleanup? And if it stays in manual, and someone would do the testing/cleaning once in a while, then that is not really automation, is it?
Also, I think I am missing a key point here, but our every test runs separate VM and every change is done in that VM which is destroyed at the end. Also we run tests in many containers - does it mean that every container needs to pick this task so it gets cleaned?

atodorov:
@marusak 

> When this would be run?

We want this to be run on every pull request. The reason being is that we don't have a cron scheduler for these kinds of tasks. 

>  And if it stays in manual

manual is only for testing purposes until I can get this PR merged and also so I can experiment triggering these jobs before final merge.

> Also, I think I am missing a key point here, but our every test runs separate VM and every change is done in that VM which is destroyed at the end. Also we run tests in many containers - does it mean that every container needs to pick this task so it gets cleaned?

I don't understand this section. Like I said above `cloud-cleanup` task is something relevant to the lorax repository so we can cleanup some cloud resources which are left over from when our tests fail and can't remove these resources themselves.

larskarlitski:
Again, I don't think this should be a "test". That's just the wrong framework.

> > A test run should remove the objects it created,
> 
> All existing tests do this already.

When is this a problem, then? How often do the tests fail in such a way that stray images are left over on the clouds?

> However they time out from time to time and we're left with resources dangling around. Since we're moving away from Jenkins we'd need some sort of a cron system which will trigger these scripts regularly and do the clean up.

Tests should always clean up after themselves, even in the event of a timeout.

> > I don't think this is the right approach.
> 
> Please share other ideas you have. Without a cron based system or a full-blown reservation system like Beaker I'm out of ideas.

Sorry, I don't see the problem this is solving. If only a few test resources stay around, it's easy enough to clean those up manually every now and then. If a lot of test resources stay around regularly, the tests need to be fixed.

martinpitt:
> Tests should always clean up after themselves, even in the event of a timeout.

The CI bot hard-kills the entire test after 2 h if it hangs, the test can't run stuff after that. Of course that is only the last line of defence -- tests should have their own timeout that matches what they are running, and they can of course then do cleanup stuff.

> have some sort of triggering system (cron in CI would be great) 

We kinda sorta have that, with .tasks files. They trigger regular stuff like image refreshes or npm updates, they could also run such cleanup things. cockpituous doesn't currently run .tasks for this project, [just for cockpit-composer](https://github.com/weldr/cockpit-composer/blob/master/.tasks), but if you need that it's easy enough to enable.

atodorov:
> We kinda sorta have that, with .tasks files. 

@martinpitt this sounds acceptable.

- On what OS do these task files execute? 
- How often do they trigger ?
- Do they have access to environment (for credentials) ?

atodorov:
@martinpitt - can you inspect the proposed `.tasks` file ?

martinpitt:
Sorry @atodorov , this fell through the cracks.

>  On what OS do these task files execute?

They run directly in the cockpit/tasks container, so Fedora-ish. If some dependency is missing, we can add it to [its Dockerfile](https://github.com/cockpit-project/cockpituous/blob/master/tasks/Dockerfile).

> How often do they trigger ?

Several times a day. It's expected that most tasks do their own frequency check, e. g. npm-update or image-refresh only run once a week. We do this by checking modification times of github issues or PRs with the `bot` label, we have a `task.stale()` helper for that ([example](https://github.com/cockpit-project/bots/blob/master/image-trigger#L103)).

> Do they have access to environment (for credentials) ?

Yes, of course.

.tasks
```diff
@@ -0,0 +1,15 @@
...
+    chance=10
+else
+    chance=$(shuf -i 0-10 -n 1)
+fi
+
+if [ $chance -gt 9 ]; then
+    ./tests/cleanup/remove_old_objects_aws.sh
+    ./tests/cleanup/remove_old_objects_azure.sh
+    ./tests/cleanup/remove_old_objects_openstack.sh
+    ./tests/cleanup/remove_old_objects_vmware.sh
```
martinpitt: This will still run a magnitude of a dozen times a day. If that's too often, you could reduce the chance, and/or randomly pick one of the four cleanup tasks. But in general the structure looks fine.

martinpitt: BTW, the "randomly pick one" can be achieved easily by prepending "echo" in front of all of these. Running ./.tasks is expected to produce a list of command lines to run, and the bot will randomly select one. Due to that you need to be careful -- if these programs output anything to stdout, that will be interpreted as commands to run.

So I'd advise to go with "echo" here to be on the safe side.

marusak:
> Several times a day. It's expected that most tasks do their own frequency check, e. g. npm-update or image-refresh only run once a week. We do this by checking modification times of github issues or PRs with the bot label, we have a task.stale() helper for that (example).

Indeed. But we are going to have a similar problem in near future in Cockpit. We want to test that URLs that we have in source are valid (they still exist). We cannot run this in normal tests, as they don't have internet access. So we need to run it as some kind of task.  I find it pointless that bots would open issue to check the URLs, and it if passes, they would close it again. But yet we need to have something, how bots would know a week (or so) passed since the last check. Also we want to have some visible output - at least to know that the check actually ran, and it is not failing in silent/is not running at all. So we may want to think about this together with weldr and find some optimal solution for this kind of tasks.

martinpitt:
@marusak: We have these transient tasks like https://github.com/cockpit-project/bots/issues/406 indeed, which just get opened and closed again. It's our current way of keeping state, and incidentally it's also a nice place to attach the log output to. We know that every so often they fail, so keeping the issue open is a good place to track this.

I agree that it would be too much noise to create these issues for cleaning up cloud resources several times a day -- but perhaps doing that once per week would suffice?

martinpitt:
I [enabled .tasks for lorax](https://github.com/cockpit-project/cockpituous/pull/311) now, this is rolled out. This PR is conflicty right now, but once it lands, the .tasks will be executed.

atodorov:
@larskarlitski this has been open for a long time and maybe doesn't make much sense at the moment but can you review ?

PR #845, opened 2019-09-01
creator.py: include dmsquash-live-ntfs by default

gmt:
Add dmsquash-live-ntfs to the default dracut modules in livecd builds.

The omission of this is probably why:

  https://bugzilla.redhat.com/show_bug.cgi?id=1449410

didn't work out as intended (I suspect it was closed unfixed).

Without this, initramfs winds up with the ntfs-3g tools lying
around, but no hooks cause them to actually get used.

AFAICT, this is a legit bugfix, not a "proposed enhancement"
so to speak.  To test that this actually fixed the problem (well,
/a/ problem, at least), I used some scripts:

  https://github.com/gmt/test-respin-fedora-ntfsable-iso

Signed-off-by: Greg Turner <gmt@be-evil.net>

### DIFF
--- a/docs/html/_modules/pylorax/creator.html
+++ b/docs/html/_modules/pylorax/creator.html
@@ -206,7 +206,7 @@
 
 
 <span class="c1"># Default parameters for rebuilding initramfs, override with --dracut-args</span>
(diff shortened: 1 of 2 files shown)


gmt:
note: this patch or functional equivalent probably belongs in all branches, or, at least those branches based on dracut newer than 046 (inclusive).  Not sure how multi-branch changes are normally handled here but zillions of PR's seemed wrong.  Note that the documentation hunk doesn't apply to some branches but this is a trivial patch, I'm sure y'all maintainers know what to do with it :)

bcl:
Made a new pr #849 without the html edit, which will also run the tests.

PR #846, opened 2019-09-02
Use osbuild to build tar and qcow2 images

larskarlitski:
See individual commits.

This is a work in progress, but I think it's far enough that others can take a look.


larskarlitski:
Update: Remove `tar.ks` and `qcow2.ks` and don't include `deps.toml` and `frozen.toml` for those composes that use osbuild.

Depends on https://github.com/weldr/lorax/pull/847 now (included in this PR).

bcl:
This is being implemented in https://github.com/osbuild/osbuild-composer

PR #847, opened 2019-09-02
Don't depsolve output types at startup

larskarlitski:
Templates installed into /usr/share ought to depsolve correctly.

This is also redundant, because composer depsolves when a compose is
started. Only then does it have the full list of packages and up to date
metadata.

The returned errors were not used anywhere except in the status call.

### DIFF
--- a/src/pylorax/api/compose.py
+++ b/src/pylorax/api/compose.py
@@ -46,7 +46,7 @@ from pykickstart.version import makeVersion
 
 from pylorax import ArchData, find_templates, get_buildarch
 from pylorax.api.gitrpm import create_gitrpm_repo
-from pylorax.api.projects import projects_depsolve, projects_depsolve_with_size, dep_nevra
+from pylorax.api.projects import projects_depsolve_with_size, dep_nevra
 from pylorax.api.projects import ProjectsError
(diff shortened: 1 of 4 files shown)


larskarlitski:
> NAK. The point of this is so that the UI can check the status and report to the user that their system is not setup correctly for the included output templates, instead of being surprised when trying to compose an image.

The UI doesn't make use of this information and couldn't, because it's not machine-readable.

Files we ship in the package *ought* to be correct, in the same way that all python files we ship don't contain syntax errors.

Most importantly, we're moving away from kickstart-based installations with https://github.com/weldr/lorax/pull/846. This will not be needed at that point, and it's easier to implement the switch to osbuild without this.

bcl:
> Files we ship in the package _ought_ to be correct, in the same way that all python files we ship don't contain syntax errors.

You would hope, but that's not always the case. eg. the way things are divided up for RHEL7 (which may have been why I added this)

> Most importantly, we're moving away from kickstart-based installations with #846. This will not be needed at that point, and it's easier to implement the switch to osbuild without this.

I think we're going to have to discuss that in more detail.

bcl:
Closing this, we need to make sure the templates can be depsolved at runtime.

PR #848, opened 2019-09-03
Use Fedora 30 for testing in Docker

atodorov:
so we don't hit dnf install errors in Travis CI


larskarlitski:
@bcl fixed this on master. Thanks!

PR #851, opened 2019-09-16
WIP: attempt at RHEL7 nested virt issues

atodorov:
I am starting with this old commit which I've kept around from when we were initially working on the CI backport.

### DIFF
--- a/test/check-cli
+++ b/test/check-cli
@@ -1,7 +1,5 @@
 #!/usr/bin/python3
 
-import unittest
-
 import composertest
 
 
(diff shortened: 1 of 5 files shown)


atodorov:
An interesting note from https://wiki.qemu.org/Features/TCG about performance:

> You should disable ASLR (address space layout randomisation). This is a kernel security feature to make it harder for malicious code to guess memory addresses while trying to execute exploits. However for TCG is means code translations that could otherwise be shared end up being re-translated multiple times.

bcl:
`Cannot set up guest memory 'pc.ram': Cannot allocate memory` looks like the host ran out of ram?

bcl:
Same problem we were seeing before - `STDERR: ssh_exchange_identification: read: Connection reset by peer`

bcl:
Instead of lowering the boot memory we should probably increase the VM memory here:
https://github.com/atodorov/lorax/blob/rhel7_nested_kvm/test/composertest.py#L31

atodorov:
> Instead of lowering the boot memory we should probably increase the VM memory here:
https://github.com/atodorov/lorax/blob/rhel7_nested_kvm/test/composertest.py#L31

Trying that as well.

Note: I've been told on IRC that the CI systems have 4GiB memory in total so everything that we do inside must fit or we need to request an increase.

PR #853, opened 2019-09-19
aarch64: Fix live-iso creation on aarch64

bcl:
Drop unneeded uboot-tools, and remove iso-graft from the aarch64.tmpl

Related: rhbz#1752002


atodorov:
FTR we need `-machine` specified for qemu on aarch64.

PR #854, opened 2019-09-20
Keep the zramctl utility from util-linux on boot.iso

M4rtinK:
Anaconda uses zram to allow installation on low memory systems.
We used to have a custom script called "zram-stats", that can be
used to test and debug zram usage during installation.

The script no longer works & zramctl now provides much better
output than our script ever did. So we decided to decommission
the old Anaconda provided script & use zramctl instead.

So change the cleanup rule in the Lorax boot.iso template
to keep the zramctl utility.

Related: rhbz#1561773


M4rtinK:
Anaconda PR removing the old zram-stats script: https://github.com/rhinstaller/anaconda/pull/2138

PR #860, opened 2019-09-30
WIP: tests: Pass extra params to QEMU on secondary arches

atodorov:
For s390x and power there are default values for `-machine` so I'm not adding them explicitly.

### DIFF
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -19,12 +19,21 @@ if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
   }
 fi
 
+
 export QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
-export QEMU="$QEMU_BIN -machine accel=kvm:tcg"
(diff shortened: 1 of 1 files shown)


atodorov:
@bcl:

> STDERR: KVM is not supported for this guest CPU type

which one should I try next ? 

FTR I saw this on machine where I could do modprobe kvm and another one which says such module doesn't exist. Not sure how to select aarch machines which support KVM virtualization.

tests/cli/lib/lib.sh
```diff
@@ -19,12 +19,22 @@ if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
   }
 fi
 
+
+if [ `rlGetArch` == "ppc64le" ]; then
+    # because there is only qemu-system-ppc64 !
+    ln -s /usr/bin/qemu-system-ppc64 /usr/bin/qemu-system-ppc64le
```
larskarlitski: I'd set the `QEMU_BIN` variable here, instead of changing the underlying system.

atodorov:
Nested virt will no longer be in use so closing.

PR #861, opened 2019-10-01
test: Disable pylint subprocess check check

bcl:
The return value for subprocess is already being checked, no need to use
check=True.


atodorov:
On one side this makes pylint happy, OTOH I don't see how using check=True can hurt. We can also remove the explicit asserts.

If you prefer going with the disable directive I suggest placing this on the 2 offending lines. Written like that it disables this check until the end of the current file and in theory it is possible for us to use another subprocess.run somewhere in this file without checking the return value and pylint not being able to catch that.

bcl:
I did it that way so that it didn't touch the existing code. I don't think it really matters one way or the other.

PR #862, opened 2019-10-03
tests: unskip Azure scenario

atodorov:


bcl:
Any idea why all those other tests failed?

Also, do you have any ideas as to how we could better manage things like this? I don't like seeing patches like this, it's making the history messy.

chris1984:
rekicking jenkins

atodorov:
> rekicking jenkins, +1 can we just override the all tests pass and leave a comment or something?

Jenkins is no more, not sure what you did.

atodorov:
> Any idea why all those other tests failed?

Travis failed pylint which was related to my changes. Fixed.

A number of the other scenarios failed with 503 from the mirrors. I saw live-iso timed out.

> Also, do you have any ideas as to how we could better manage things like this? I don't like seeing patches like this, it's making the history messy.

Not really. If we move the code outside of the repository then PR-CI integration will become much harder. I don't see the problem here but this is something better discussed on the ML.

chris1984:
> > rekicking jenkins, +1 can we just override the all tests pass and leave a comment or something?
> 
> Jenkins is no more, not sure what you did.

Sorry meant travis

larskarlitski:
Nice to see this work again finally!

PR #865, opened 2019-10-07
mkksiso tool to add a kickstart to an existing boot.iso

bcl:
This tool will add the kickstart to the boot.iso, edit the kernel boot
arguments so that the kickstart is used when the iso boots, as well as
allow adding extra files and directories to the / of the iso which can
then be used by the kickstart (they are found under /run/install/repo
while Anaconda is running).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,10 +33,7 @@ all: src/pylorax/version.py src/composer/version.py
 install: all
 	$(PYTHON) setup.py install --root=$(DESTDIR) --prefix=$(PREFIX)
 	mkdir -p $(DESTDIR)/$(mandir)/man1
-	install -m 644 docs/man/lorax.1 $(DESTDIR)/$(mandir)/man1
-	install -m 644 docs/man/livemedia-creator.1 $(DESTDIR)/$(mandir)/man1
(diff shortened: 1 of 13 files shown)


bcl:
Note that this is still in progress, but I have tested it with Fedora and RHEL8 boot.iso on x86.
Still needs docs, other arches for mkisofs, and testing (likely manual, not CI based).

jkonecny12:
Just an idea, don't you want to use and extend existing `pungi-patch-iso`? 
https://pagure.io/pungi/blob/master/f/bin/pungi-patch-iso

Some time ago I wanted to extend the patch-iso from pungi to add a kickstart file to the ISO but I didn't found a time to do that, unfortunately. However, it shouldn't be hard thing to do.

bcl:
> Just an idea, don't you want to use and extend existing `pungi-patch-iso`?
> https://pagure.io/pungi/blob/master/f/bin/pungi-patch-iso

Ha! Because I didn't know about it. It looks like it does support kickstart as long as it is named ks.cfg.
It isn't using xorrisofs yet though, its dependency list is a bit heavier than lorax's, and I'm almost done with this so I think I'll continue with this, thanks.

PR #866, opened 2019-10-11
Backport 871 & 885, removal of nested virt in tests

atodorov:
cc @martinpitt - in case the `rhel-8` scenario fails can you rerun on cockpit-7 manually and see if that makes a difference.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,15 @@ $(VM_IMAGE): srpm bots
 vm: $(VM_IMAGE)
 	echo $(VM_IMAGE)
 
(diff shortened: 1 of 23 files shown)


martinpitt:
@atodorov: I sent PR #870 to fix bots properly (your first commit included here, but that's buggy -- not your fault).

live-iso timed out, and qcow2 timed out in ssh -- are these a sign of nested virt failure? qcow ran on cockpit-10 -- I went into the CI pod, started a rhel-8-1 image (the same as the test uses), and launched a nested qemu-kvm without problems.

atodorov:
vmware failure is unrelated so merging

PR #871, opened 2019-10-16
Split testing of built image into separate step

bcl:
Still need to deal with the liveimg-tar test, pushing this PR to see if it runs in cockpit CI.

### DIFF
--- a/tests/cli/test_boot_live-iso.sh
+++ b/tests/cli/test_boot_live-iso.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+#####
+#
+# Test the live-iso image
+#
+#####
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+rlJournalStart
+    rlPhaseStartTest "Verify live iso"
+        # Just the fact that this is running means the image can boot and ssh is working
+        rlRun -t -c "passwd --status root | grep -E '^root\s+NP?'" 0 "root account has no password set"
+        rlAssertGrep "liveuser" /etc/passwd
+        rlAssertGrep "custom_cmdline_arg" /proc/cmdline
+    rlPhaseEnd
+rlJournalEnd
+rlJournalPrintText
--- a/tests/cli/test_boot_live-iso.sh
+++ b/tests/cli/test_boot_live-iso.sh
@@ -15,7 +15,7 @@ set -e
 rlJournalStart
     rlPhaseStartTest "Verify live iso"
         # Just the fact that this is running means the image can boot and ssh is working
-        rlRun -t -c "passwd --status root | grep -E '^root\s+NP?'" 0 "root account has no password set"
+        ROOT_ACCOUNT_LOCKED=0 verify_image liveuser localhost "-p 22"
         rlAssertGrep "liveuser" /etc/passwd
         rlAssertGrep "custom_cmdline_arg" /proc/cmdline
     rlPhaseEnd
--- a/tests/cli/test_boot_tar_kickstart.sh
+++ b/tests/cli/test_boot_tar_kickstart.sh
@@ -15,6 +15,7 @@ set -e
 rlJournalStart
     rlPhaseStartTest "Verify VM instance"
         # Just the fact that this is running means the image can boot and ssh is working
+        CHECK_CMDLINE=0 verify_image root localhost "-p 22"
         rlAssertExists "/root/.ssh/authorized_keys"
     rlPhaseEnd
 rlJournalEnd
--- a/tests/cli/test_boot_tar_kickstart.sh
+++ b/tests/cli/test_boot_tar_kickstart.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+#####
+#
+# Test the liveimg installed tar disk image
+#
+#####
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+rlJournalStart
+    rlPhaseStartTest "Verify VM instance"
+        # Just the fact that this is running means the image can boot and ssh is working
+        rlAssertExists "/root/.ssh/authorized_keys"
+    rlPhaseEnd
+rlJournalEnd
+rlJournalPrintText
--- a/tests/cli/test_compose_live-iso.sh
+++ b/tests/cli/test_compose_live-iso.sh
@@ -15,40 +15,66 @@ set -e
 CLI="${CLI:-./src/bin/composer-cli}"
 
 rlJournalStart
-    rlPhaseStartSetup
-        rlAssertExists $QEMU_BIN
-    rlPhaseEnd
-
     rlPhaseStartTest "compose start"
         rlAssertEquals "SELinux operates in enforcing mode" "$(getenforce)" "Enforcing"
 
+        TMP_DIR=$(mktemp -d /tmp/composer.XXXXX)
+        SSH_KEY_DIR=$(mktemp -d /tmp/composer-ssh-keys.XXXXXX)
+
+        rlRun -t -c "ssh-keygen -t rsa -N '' -f $SSH_KEY_DIR/id_rsa"
+        PUB_KEY=$(cat "$SSH_KEY_DIR/id_rsa.pub")
+
+        cat > "$TMP_DIR/with-ssh.toml" << __EOF__
+name = "with-ssh"
+description = "HTTP image with SSH"
+version = "0.0.1"
+
+[[packages]]
+name = "httpd"
+version = "*"
+
+[[packages]]
+name = "openssh-server"
+version = "*"
+
+[[packages]]
+name = "beakerlib"
+version = "*"
+
+[customizations.services]
+enabled = ["sshd"]
+
+[[customizations.user]]
+name = "root"
+key = "$PUB_KEY"
+
+[customizations.kernel]
+append = "custom_cmdline_arg"
+__EOF__
+
+        rlRun -t -c "$CLI blueprints push $TMP_DIR/with-ssh.toml"
+
         # NOTE: live-iso.ks explicitly disables sshd but test_cli.sh enables it
-        UUID=`$CLI compose start example-http-server live-iso`
+        UUID=$($CLI compose start with-ssh live-iso)
         rlAssertEquals "exit code should be zero" $? 0
 
-        UUID=`echo $UUID | cut -f 2 -d' '`
+        UUID=$(echo "$UUID" | cut -f 2 -d' ')
     rlPhaseEnd
 
     rlPhaseStartTest "compose finished"
-        wait_for_compose $UUID
+        wait_for_compose "$UUID"
 
-        rlRun -t -c "$CLI compose image $UUID"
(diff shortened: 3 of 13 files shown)


tests/cli/test_compose_live-iso.sh
```diff
@@ -15,40 +15,66 @@ set -e
...
+
+[customizations.services]
+enabled = ["sshd"]
+
+[[customizations.user]]
+name = "root"
+key = "$PUB_KEY"
+
+[customizations.kernel]
+append = "custom_cmdline_arg"
```
atodorov: You don't assert on this parameter in test_boot_live-iso.sh. Copy&paste error maybe ?

tests/cli/test_compose_live-iso.sh
```diff
@@ -15,40 +15,66 @@ set -e
...
 
     rlPhaseStartTest "compose finished"
-        wait_for_compose $UUID
-
-        rlRun -t -c "$CLI compose image $UUID"
-        IMAGE="$UUID-live.iso"
-    rlPhaseEnd
-
-    rlPhaseStartTest "Start VM instance"
-        boot_image "-boot d -cdrom $IMAGE" 120
```
atodorov: this breaks the existing behavior of being able to execute this script on a bare metal system, build the image, boot a VM from it and ssh into the test VM to verify a few settings. Preferably this would be disabled with an if statement when running inside Cockpit CI. If you want you can pass additional variables from the Python caller.

tests/cli/test_compose_live-iso.sh
```diff
@@ -15,40 +15,66 @@ set -e
...
-    rlPhaseEnd
+        wait_for_compose "$UUID"
 
-    rlPhaseStartTest "Verify VM instance"
-        # run generic tests to verify the instance
-        ROOT_ACCOUNT_LOCKED=0 verify_image liveuser localhost "-p $SSH_PORT"
```
atodorov: Again on the topic of being able to execute these tests scripts easily by hand: Instead of removing this section can you instead surround with an if condition and maybe call test_boot_live-iso.sh from inside here.

tests/cli/test_compose_qcow2.sh
```diff
@@ -51,31 +51,26 @@ __EOF__
...
-        IMAGE="$UUID-disk.qcow2"
-    rlPhaseEnd
-
-    rlPhaseStartTest "Start VM instance"
-        boot_image "-boot c -hda $IMAGE" 60
-    rlPhaseEnd
-
-    rlPhaseStartTest "Verify VM instance"
-        # run generic tests to verify the instance
-        verify_image root localhost "-i $SSH_KEY_DIR/id_rsa -p $SSH_PORT"
```
atodorov: Same comment here for booting the test VM and verifying the VM instance. It is important to preserve existing behavior and be able to execute these scripts by hand, like the rest of them.

tests/test_image.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+. $BEAKERLIB_DIR/TestResults
+
+if [ $TESTRESULT_RESULT_ECODE != 0 ]; then
+  echo "Test failed. Leaving log in $BEAKERLIB_DIR"
+  exit $TESTRESULT_RESULT_ECODE
+fi
+
+rm -rf $BEAKERLIB_DIR
```
atodorov: test_cli.sh should probably call this one directly to reduce the amount of duplicated code. However that is not related to the current effort of ditching nested virt.

bcl:
Addressing the issue of running these manually:

yes it breaks that. And no, I don't have a good solution for bare metal yet. 

The goal here is to get the PR tests running reliably without nested virt. It may make sense to add another test/script runs the compose script on the host and boots the image using the cockpit ci virtual machine code. At least that's the direction I am leaning right now -- I do not think we should continue to try to run qemu ourselves from inside tests given the issues with nested, and the other problems you are running into with it requiring different parameters on different arches. May as well let the libvirt based cockpit ci code take care of all of that.

tests/cli/test_boot_live-iso.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
```
atodorov: Could be a good idea to actually mention that this needs to be executed from inside the VM. 

Or even better group the existing files in different directories to indicate where they will be executed.

tests/test_image.sh
atodorov: I still think it can be called from inside test_cli but doesn't matter that much. Renaming, groupping into directories may help with making it explicit which part executes where.

atodorov:
@bcl, @larskarlitski please review additional changes in #885. It is mostly docs updates but also adds a make target to build a VM containing all repos from the host and updates the asserts inside the running VM to match what was there before.

atodorov:
OTOH nested virt seems to be working right now, see tar scenario in #883.
@martinpitt is leading work in this space and may comment more.

FTR: I do prefer the straight-forward approach  where we use nested virt during testing but we'll need libvirt for secondary arches anyway so I'm not against Brian's approach as well. From what I can tell we may have some issues running the Cockpit CI wrappers on RHEL 7 (missing Python3) but maybe we don't need to (just supply the VM with all the repo files needed).

bcl:
> OTOH nested virt seems to be working right now, see tar scenario in #883.

Ha! We must have frightened it into working ;)

I'm not ready to push this yet, I haven't figured out a clean way to support running on bare metal yet, and I currently have a bunch of other things to work on so hopefully nested virt continues to work until we sort this out.

atodorov:
> I'm not ready to push this yet, I haven't figured out a clean way to support running on bare metal yet,

What exaclty do you mean? Doesn't `make vm && ./test/check-cli ....` work for you ?

bcl:
> What exaclty do you mean? Doesn't `make vm && ./test/check-cli ....` work for you ?

I mean running the ./tests/cli/ scripts without the cockpit ci, isn't that how you are testing things on other arches?

atodorov:
>> What exaclty do you mean? Doesn't make vm && ./test/check-cli .... work for you ?
> I mean running the ./tests/cli/ scripts without the cockpit ci,

This is what I am proposing in #885. I gave it a try and it seems to do the job for QE. The rest is cosmetic changes and doc updates. The part where we collect the repo files from the host into the base vm image is important b/c we need to make sure we're as close as possible to the internal snapshot. But basically we're advocating the same thing right now.

> isn't that how you are testing things on other arches?

No!

Not yet - we're still to get a new snapshot and test with the new command line instructions
Not before - previously the way we've been testing was by calling `./tests/test_cli.sh ./tests/cli/test_*.sh` direclty onto bare metal installed with the appropriate OS snapshot + export CLI=/usr/bin/composer-cli.

So if you think this PR is good to go let's merge it. My concerns have been proven unreasonable in the mean time.

tests/test_cli.sh
```diff
@@ -15,6 +15,22 @@ function setup_tests {
...
+
+    # Make the live-iso boot more quickly (grub.cfg)
+    for cfg in "$share_dir"/templates.d/99-generic/live/config_files/*/grub.conf; do
+        sed -i.orig 's/^timeout.*/timeout 2/' "$cfg"
+    done
+
+    # Make the live-iso boot more quickly (grub2-efi.cfg)
+    for cfg in "$share_dir"/templates.d/99-generic/live/config_files/*/grub2-efi.cfg; do
+        sed -i.orig 's/^set timeout.*/set timeout=2/' "$cfg"
+    done
```
atodorov: I don't like the fact that we're not restoring the changes in `teardown_tests`. They are harmless but still better restore back to what was there before. You don't know if the same system won't be used for some manual testing afterwards.

tests/cli/test_boot_tar_kickstart.sh
```diff
@@ -0,0 +1,21 @@
...
+#####
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+rlJournalStart
+    rlPhaseStartTest "Verify VM instance"
+        # Just the fact that this is running means the image can boot and ssh is working
```
atodorov: I prefer having the same asserts before and after the change so we can still make use of the helper functions in more importantly extend the asserts in the future with validations that apply to all image types.

The assert you currently have  will become a bit redundant b/c if we can ssh into localhost using the existing keys there's not much point in checking if `authorized_keys` exists.

Adding `CHECK_CMDLINE=0 verify_image root localhost "-p 22` should work here, see https://github.com/weldr/lorax/pull/885/commits/5af8fb1ef99cbb70b54b21bd4ace96d0f20e2059. Also note the part where we upload the existing keys to the VM in `check-cli:L45` in above commit.

atodorov:
`fedora-31/tar` fails with `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`. Can be fixed with the migration to podman, cherry-pick
https://github.com/weldr/lorax/pull/887/commits/6389b26b999e8edc9ba0cc5d7ffb032fd0b7d21d

tests/cli/test_boot_tar_kickstart.sh
bcl: Cherry picked the docker->podman commit, 5af8fb, and added the changes to the new boot script as part of that commit.

atodorov:
vmware fails with 

```
Traceback (most recent call last):
  File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 285, in <module>
    main()
  File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 281, in main
    cluster.resourcePool)
  File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 185, in create_vm
    portKey = search_port(dvs, portgroup.key)
  File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 93, in search_port
    return search_portkey[0]
IndexError: list index out of range
```

while starting the VM instance. This appears to be some sort of misconfiguration in the environment. I've asked @chris1984 to take a look.

Other than that - ship it!

chris1984:
> vmware fails with
> 
> ```
> Traceback (most recent call last):
>   File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 285, in <module>
>     main()
>   File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 281, in main
>     cluster.resourcePool)
>   File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 185, in create_vm
>     portKey = search_port(dvs, portgroup.key)
>   File "/tmp/composer-vmware.CcPpR/pyvmomi-community-samples/samples/create_vm.py", line 93, in search_port
>     return search_portkey[0]
> IndexError: list index out of range
> ```
> 
> while starting the VM instance. This appears to be some sort of misconfiguration in the environment. I've asked @chris1984 to take a look.
> 
> Other than that - ship it!

@atodorov did anything change in the scripts? I see a ton of vms getting created.

bcl:
> @chris1984 did anything change in the scripts? I see a ton of vms getting created.

No, we haven't changed that part of things, at least in this PR.

bcl:
All tests except vmware were passing, vmware failure is unrelated so merging this to master.

PR #872, opened 2019-10-16
docs: Override macboot/nomacboot documentation

bcl:
These are two sides of the same variable (domacboot) so sphinx-argparse
generated docs are a bit confusing. Override the docs for them to clear
things up.


bcl:
(Note that the RHEL version of this will need to invert the Default, we do not build macboot enabled isos by default).

PR #875, opened 2019-10-17
Improve lorax-composer disk size estimate

bcl:


bcl:
Figured out why the host metadata and anaconda metadata were so different in size. This should be robust enough to work in all cases.

bcl:
Merged into rhel7-extras for lorax-19.7.40-1

PR #876, opened 2019-10-17
Improve size estimation in lorax-composer

bcl:


bcl:
Actually, dnf picks the host for these since it has the larger filesystem. No need for these on master.

PR #881, opened 2019-10-18
lorax: Drop unused --title option

### LINKED ISSUE #863: livemedia-creator --title
Hi - according to the [documentation](https://weldr.io/lorax/livemedia-creator.html), _livemedia-creator_ option `--title <...>`, is used in bootloader config files:

    Substituted for @TITLE@ in bootloader config files

    Default: "Linux Live Media"

However, none of the config files use `@TITLE@`, nor is it referenced anywhere in the code. I would recommend we remove it from the docs.

bcl:
It has never been hooked up to anything, let alone substitution @TITLE@
(which also hasn't ever been used in the config files).

Closes #863


chris1984:
[test fedora-30]

PR #882, opened 2019-10-21
Add new test to run 'test' and 'check' in cockpit ci

jrusz:
I want to move the tests that ran in travis CI to cockpit CI. I created a new test for that and I want it to do the same as "make check" and "make test" which is what was running in the docker container in travis.

Right now I am stuck on "check". It gives me this traceback:

:: [ 04:41:07 ] :: [   LOG    ] :: --------------- OUTPUT START ---------------
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR: Traceback (most recent call last):
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR:   File "/tests/pylint/runpylint.py", line 34, in <module>
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR:     rc = linter.run()
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR:   File "/usr/lib/python3.7/site-packages/pocketlint/__init__.py", line 339, in run
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR:     files = self._files
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR:   File "/usr/lib/python3.7/site-packages/pocketlint/__init__.py", line 166, in _files
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR:     lines = fo.readlines()
:: [ 04:41:07 ] :: [   LOG    ] :: STDERR: OSError: [Errno 22] Invalid argument
:: [ 04:41:07 ] :: [   LOG    ] :: ---------------  OUTPUT END  ---------------

I am not sure what causes the error. I am running the test by first building the vm "make vm" and then  "./test/check-cli TestPylint.test_ci"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ PYTHON ?= /usr/bin/python3
 DESTDIR ?= /
 PREFIX ?= /usr
 mandir ?= $(PREFIX)/share/man
-DOCKER ?= docker
+DOCKER ?= podman
 DOCS_VERSION ?= next
 RUN_TESTS ?= ci
 
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,9 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 IMAGE_RELEASE = $(shell awk -F: '/FROM/ { print $$2}' Dockerfile.test)
 
 ifeq ($(TEST_OS),)
-TEST_OS = fedora-30
+OS_ID = $(shell awk -F= '/^ID/ {print $$2}' /etc/os-release)
+OS_VERSION = $(shell awk -F= '/^VERSION_ID/ {print $$2}' /etc/os-release)
+TEST_OS = $(OS_ID)-$(OS_VERSION)
 endif
 export TEST_OS
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
--- a/tests/pylorax/test_projects.py
+++ b/tests/pylorax/test_projects.py
@@ -330,7 +330,7 @@ def singlerepo_v0():
         "check_gpg": True,
         "check_ssl": True,
         "gpgkey_urls": [
-            "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-32-x86_64"
+            "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-" + os.environ['TEST_OS'] + "-x86_64"
         ],
         "name": "single-repo",
         "system": False,
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: python
 services:
   - docker
 
+env:
+  - DOCKER=docker
+
 script:
   - make test-in-docker
 
--- a/tests/composer/test_blueprints.py
+++ b/tests/composer/test_blueprints.py
@@ -232,7 +232,7 @@ class BlueprintsTest(unittest.TestCase):
     @unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Test requires a running API server")
     def test_tag(self):
         """blueprints tag"""
-        rc = blueprints_tag("/run/weldr/api.socket", 0, ["glusterfs"], show_json=False)
+        rc = blueprints_tag("/run/weldr/api.socket", 0, ["example-http-server"], show_json=False)
         self.assertTrue(rc == 0)
 
     @unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Test requires a running API server")


test/run
```diff
@@ -11,6 +11,8 @@ if [ -n "$TEST_SCENARIO" ]; then
     test/check-cli TestQcow2
   elif [ "$TEST_SCENARIO" == "tar" ]; then
     test/check-cli TestTar
+  elif [ "$TEST_SCENARIO" == "ci" ]; then
+	test/check-cli TestPylint
```
atodorov: Why not call `make ci` (aka `make check test`) here like we discussed instead of starting a VM ?

atodorov: You also need the corresponding test scenario in `testmap.py` under `_manual` because we want to kick off this scenario and make sure it passes before merging this PR.

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,59 @@
...
+  rsync \
+  e2fsprogs \
+  xz-lzma-compat \
+  pbzip2 \
+  squashfs-tools \
+  qemu-img \
+  which 
+
+mkdir /lorax
+git clone https://github.com/weldr/lorax /lorax/
```
atodorov: this isn't paying attention to the current branch or PR commit that is being tested.

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,59 @@
...
+  python3-pyvmomi \
+  beakerlib \
+  sudo \
+  rsync \
+  e2fsprogs \
+  xz-lzma-compat \
+  pbzip2 \
+  squashfs-tools \
+  qemu-img \
+  which
```
atodorov: @larskarlitski please see my comment above for running `make ci` directly into `test/run`. Possibly lots of the packages above will end up in vm.install which may turn out to be just too much. What do you think ?

tests/cli/test_ci.sh
atodorov: If we continue going the same way as now there should be some way to mount the current git checkout into the VM and just call make ci from there. The CI slave system (be it Travis or Cockpit) already has a local git checkout before even executing any of this code.

test/run
jrusz: Well I thought we want to run the make ci inside the VM which will have a specific fedora/rhel compose running...

atodorov: You are right. Scratch my previous comment.

tests/cli/test_ci.sh
bcl: In my split the tests branch I ended up adding a bit of code to the class that copies the tests into the VM using `self.machine.upload(["../tests"], "/")` that could be used to copy the whole checkout if needed.

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,59 @@
...
+# Note: execute this file from the project root directory
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+CLI="${CLI:-./src/bin/composer-cli}"
+
+dnf -y install \
```
bcl: We need a better way to manage these test dependencies, probably a file similar to a pip requirements file. There should be a single place to edit the list.

bcl:
Something else to keep in mind here is that the cockpit vm is currently Fedora 30. A primary reason for using podman/docker containers to run this test is that it can easily be run on the actual release that it is targeting.

test/check-cli
```diff
@@ -49,6 +49,9 @@ class TestTar(composertest.ComposerTestCase):
     def test_tar_kickstart(self):
         self.runCliTest("/tests/cli/test_compose_tar_kickstart.sh")
 
+class TestPylint(composertest.ComposerTestCase):
+    def test_ci(self):
+        self.runCliTest("/tests/cli/test_ci.sh")
```
atodorov: This must be `runImageTest()` after cherry-picking the changes in #871, at least the 1st commit which makes it possible to run anything inside the VM without starting lorax-composer.

Otherwise we're starting lorax-composer before running `make ci` and it likely causes the composer process started by the Python test suite to fail.

test/check-cli
```diff
@@ -104,7 +104,7 @@ class TestTar(composertest.ComposerTestCase):
 
 class TestPylint(composertest.ComposerTestCase):
     def test_ci(self):
-        self.runCliTest("/tests/cli/test_ci.sh")
+        self.runImageTest("/tests/cli/test_ci.sh")
```
atodorov: The 3rd commit must be squashed with the first one.

tests/cli/test_ci.sh
atodorov: +1 for the single place for dependencies. 

However many of the test jobs don't need everything, which means they will still incur the penalty of network downloads and dnf install for things not needed in the particular job. If this is acceptable then we've already got this  place, which is `test/vm.install`.

test/check-cli
jrusz: I believe this is solved now.

tests/cli/test_ci.sh
atodorov: FTR I'm starting to like the fact that the test script makes sure it has all the packages it needs and this doesn't affect the rest of the test suite (i.e. not in vm.install and affecting all scenarios). I think this is acceptable for now so marking this comment as resolved.

tests/cli/test_ci.sh
atodorov: FTR I'm starting to like the fact that the test script makes sure it has all the packages it needs and this doesn't affect the rest of the test suite (i.e. not in vm.install and affecting all scenarios). I think this is acceptable for now so marking this comment as resolved.

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+. $(dirname $0)/lib/lib.sh
+
+CLI="${CLI:-./src/bin/composer-cli}"
```
atodorov: This isn't needed I think. So is line 7 above. Probably copy&paste error.

tests/cli/test_ci.sh
atodorov: Running against HEAD instead of the current checkout is not good. Even when the failures reported in the logs are fixed this job will still report FAIL b/c it won't pick up the changes.

@jrusz `self.machine.upload(["../"], "/lorax")` should be able to copy the whole git checkout from the CI host to the VM. Then inside the VM you just `cd /lorax && make ci`. Give it a try.

jrusz: I solved this in commit 0a486514375f79eb79d0355946ab78bcb53c9495 .. There are two reasons for why I didn't use exactly self.machine.upload(["../"], "/lorax") .. First is that when I tried it I got an error saying that "scp: error: unexpected filename: ..". What could work is self.machine.upload(["../../lorax"], "/lorax") but I can't really do that on my machine because I have the bots repo there too. So I decided to just upload the relevant files I hope I got all of them.

test/check-cli
```diff
@@ -102,6 +102,20 @@ class TestTar(composertest.ComposerTestCase):
...
+    def test_ci(self):
+
+        # Create an empty folder for the upload bellow
+        self.execute(["mkdir /lorax"])
+
+        # Upload contents of repo for running the test
+        self.machine.upload(["../tests"], "/lorax/tests")
+        self.machine.upload(["../src"], "/lorax/src")
+        self.machine.upload(["../share"], "/lorax/share")
+        self.machine.upload(["../Makefile"], "/lorax")
```
atodorov: @bcl what do you think about this ? Also see comment in 
https://github.com/weldr/lorax/pull/882#discussion_r343122821

It's not particularly elegant as would be copying the entire directory or even better bind-mount the git checkout into the running VM. However it seems to work.

tests/cli/test_ci.sh
atodorov: > but I can't really do that on my machine because I have the bots repo there too.

I don't understand what the problem is and how does having bots repo on your development machine prevents you from doing something in CI ?

jrusz: It doesn't prevent me from doing something in CI but it prevents me from testing it on my machine first. What I mean is that I can't upload my whole local lorax repo because I have the bots repo inside for running the vm locally. Meaning that when I would want to run some test locally and then upload the whole lorax folder inside, it would also try to copy the whole bots folder along with the images. AFAIK I need to have it there if I want to run it locally first.

atodorov: You can always upload the entire dir inside the VM and remove the bots/ directory from the VM if it causes failures (may probably affect only pylint). If it doesn't cause any problems when running inside the VM I see no reason to worry about it.

tests/cli/test_ci.sh
jrusz: You are right, I removed it and squashed to the first commit.

jrusz:
- I made the necessary changes to my test_ci.sh file and squashed it with the first commit.

- I removed .travis.yml, moved the `docker build` and removed `test-in-docker` target in 31bb3a8b1f06e7e3395406b37d571cdfec97da18

- docs test failures were solved by also copying the `docs` directory like you suggested

WRT the failing test, I though it was caused by what @bcl fixed in PR https://github.com/weldr/lorax/pull/895 but after rebasing and running the tests it still failed, so the problem is somewhere else.

Makefile
```diff
@@ -121,15 +121,8 @@ test-in-copy:
 	make -C /lorax/ $(RUN_TESTS)
 	cp /lorax/.coverage /test-results/
 
-test-in-docker:
```
bcl: Please don't remove this. I use it to run the tests locally.
You could change it to use podman by default though, then I could stop typing DOCKER=podman make test-in-docker :)

tests/pylorax/test_projects.py
```diff
@@ -330,7 +330,7 @@ def singlerepo_v0():
         "check_gpg": True,
         "check_ssl": True,
         "gpgkey_urls": [
-            "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-32-x86_64"
```
bcl: This will break running things in podman. Ideally the cockpit tests for master branch should be running against a rawhide image, but I realize that's hard to do. The test setup should probably make the change, possibly based on the TEST_OS value?

test/check-cli
```diff
@@ -102,6 +102,21 @@ class TestTar(composertest.ComposerTestCase):
             # Run the test, on the booted image
             self.runImageTest("/tests/cli/test_boot_tar_kickstart.sh")
 
+class TestPylint(composertest.ComposerTestCase):
+    def test_ci(self):
+
+        # Create an empty folder for the upload bellow
+        self.execute(["mkdir /lorax"])
```
bcl: This should be ["mkdir", "/lorax"]

test/check-cli
bcl: That's fairly awkward. But bind mounting would be worse, allowing changes to the host. Copying all of ./lorax/ would also include any .git tree and extra artifacts, so I guess this is the least bad option.
upload can take a list of things to upload, you should be able to do this instead:
`self.machine.upload([../tests", "../src", "../share", "../Makefile"], "/lorax")`

tests/pylorax/test_projects.py
atodorov: Right, there is no Rawhide image available in Cockpit CI but there is `fedora-testing` (I don't know which version this is). Whatever we decide to do Dockerfile.test (Travis CI) and the new scenario added by Jakub must use the same Fedora version. Otherwise we'll start seeing differences. 

One option, which I don't like, is to have this switch merged only for RHEL branches and leave master as-is.

jrusz: I changed the way `TEST_OS` variable is defined in the Makefile and changed this test to use it. I don't know if it's the most elegant way but it works. I'll be glad to hear any suggestions to improve this.

test/check-cli
jrusz: Yes it is awkward. I changed it to the list now.

test/check-cli
jrusz: Thanks, I changed it and squashed with the first commit.

Makefile
jrusz: What do you think @atodorov ? I could just drop this commit and just change the `DOCKER ?= docker` line in Makefile to `DOCKER ?= podman` if that's any helpful.

tests/pylorax/test_projects.py
```diff
@@ -330,7 +330,7 @@ def singlerepo_v0():
...
+            "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-" + os.environ['TEST_OS'] + "-x86_64"
```
bcl: That only works if TEST_OS is defined, which in the case of running it with podman, it won't.
I'm also not sure that the key name will always match the TEST_OS value (eg. RHEL, CentOS)

jrusz: Well I just tried running it in podman and those blueprints tests are skipped there (SKIP: Test requires a running API server). I don't know if this is the correct behavior but it seems that this will not cause problems for docker.

atodorov: hmmm, I see `@unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Test requires a running API server")` only in `test_blueprints.py`, not in `test_projects.py`

jrusz: You're right, I mixed the two tests I was fixing together in my head. I need to check again, but it seems that this really won't work for RHEL...

Makefile
```diff
@@ -14,7 +14,9 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 IMAGE_RELEASE = $(shell awk -F: '/FROM/ { print $$2}' Dockerfile.test)
 
 ifeq ($(TEST_OS),)
-TEST_OS = fedora-30
+OS_ID = $(shell awk -F= '/^ID/ {print $$2}' /etc/os-release)
+OS_VERSION = $(shell awk -F= '/^VERSION_ID/ {print $$2}' /etc/os-release)
+TEST_OS = $(OS_ID)-$(OS_VERSION)
```
atodorov: @bcl I kind of like this. I will also be helpful for testing internal snapshots. It just sets TEST_OS to the OS version on the host, which for me is 1 less thing to think about. In CI TEST_OS is explicitly defined anyway.

Makefile
jrusz: Alright I did that but it seems that travis doesn't know podman... See [travis log](https://travis-ci.org/weldr/lorax/builds/610809398?utm_source=github_status&utm_medium=notification)

jrusz: Nevermind, I added the variable to .travis.yml and now it's running. Apparently there is no podman package in Ubuntu repos.

tests/pylorax/test_projects.py
bcl: With the changes above I think it is probably good enough.

jrusz:
Thank you for the approvals. I don't know why are the tests stuck in Expected state... Also I was thinking, maybe it would be better to change the actual name of the test? It's just called `test_ci` which I don't think is descriptive enough. Also when I was creating the bots target, pitti said that this is not a good name. I am not good with naming so I would appreciate some suggestion. Or we can just leave it as is, I don't mind.

atodorov:
@bcl I just remember I saw CentOS 8 being released last week or so. Which means that is also a viable option for Travis CI on rhel8-branch . (On RHEL 7 the docker image already uses CentOS 7 + EPEL).

What do you prefer:

1) Close this PR entirely (maybe cherry-pick the commits related to podman) & try CentOS 8 for rhel8-branch, keeping Travis CI and Dockerfile.test mostly as-is or 
2) Merge this PR & backport both to RHEL 8 & 7 branches. 

3) With/without keeping Travis CI on rhel8/7 branches.

jrusz:
@bcl There is a problem though. I can not find `pylint` or `pocketlint` package in any rhel8.1 repo.

bcl:
It's really up to you. As long as travis is still working and I can run 'make tests-in-docker' locally I don't care if it is merged or not.

As far as naming goes, I'm fine with 'ci', that describes the Makefile target that all of this is wrapping. But then I guess the class TestPylint isn't really a good name and it should be TestCI instead :)

As for backporting and running on Centos8, that is going to need the test dependencies that aren't in RHEL8 to be added to EPEL8. That's why the Dockerfile for rhel8-branch is still using Fedora 28.

atodorov:
In light of the missing packages in RHEL 8 / EPEL 8 here's what I prefer to do:

1) @jrusz  drop the first commit from the PR; also drop the "ci" scenario from bot's testmap.py; leave everything else. I will merge once CI reports pass again

2) for all packages missing in RHEL 8 @jrusz should open bugs in Bugzilla so these packages will be added to EPEL 8

3) Backporting to RHEL 8 isn't possible ATM because of the missing RPMs and it doesn't make sense to pursue this for the time being. Once we've got the missing packages in EPEL it will be either easy to backport the 1st commit or change Dockerfile to use CentOS 8 (I prefer this option)

atodorov:
@bcl please merge if you have the permissions to merge with the failing fedora-31/vmware scenario b/c none of us can. There's nothing else to be done here.

PR #883, opened 2019-10-22
Backport kickstart tar image installation test to RHEL-8

jikortus:
I'll execute the tests once the related update in `testmap.py` is merged in cockpit/bots.

### DIFF
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -46,7 +46,9 @@ verify_image() {
     SSH_OPTS="-o StrictHostKeyChecking=no $3"
     rlLogInfo "verify_image: SSH_OPTS:'$SSH_OPTS' SSH_USER:'$SSH_USER' SSH_MACHINE: '$SSH_MACHINE'"
     check_root_account "$@"
-    check_kernel_cmdline "$@"
+    if [ "$CHECK_CMDLINE" != 0 ]; then
+        check_kernel_cmdline "$@"
+    fi
 }
 
 check_root_account() {
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -16,6 +16,28 @@ if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
   }
 fi
 
+export QEMU_BIN="/usr/libexec/qemu-kvm"
+export QEMU="$QEMU_BIN -machine accel=kvm:tcg"
+export SSH_PORT=2222
+
+boot_image() {
+    QEMU_BOOT=$1
+    TIMEOUT=$2
+    rlRun -t -c "$QEMU -m 2048 $QEMU_BOOT -nographic -monitor none \
+                 -net user,id=nic0,hostfwd=tcp::$SSH_PORT-:22 -net nic &"
+    # wait for ssh to become ready (yes, http is the wrong protocol, but it returns the header)
+    tries=0
+    until curl -sS -m 15 "http://localhost:$SSH_PORT/" | grep 'OpenSSH'; do
+        tries=$((tries + 1))
+        if [ $tries -gt $TIMEOUT ]; then
+            exit 1
+        fi
+        sleep 1
+        echo "DEBUG: Waiting for ssh become ready before testing ..."
+    done;
+}
+
+
 # a generic helper function unifying the specific checks executed on a running
 # image instance
 verify_image() {
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -23,7 +23,7 @@ export SSH_PORT=2222
 boot_image() {
     QEMU_BOOT=$1
     TIMEOUT=$2
-    rlRun -t -c "$QEMU -m 2048 $QEMU_BOOT -nographic -monitor none \
+    rlRun -t -c "$QEMU -m 2048 $QEMU_BOOT -nographic \
                  -net user,id=nic0,hostfwd=tcp::$SSH_PORT-:22 -net nic &"
     # wait for ssh to become ready (yes, http is the wrong protocol, but it returns the header)
     tries=0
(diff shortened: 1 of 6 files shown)


jikortus:
@bcl Could you please grant devel_ack+ for RHBZ1733975?

Makefile
```diff
@@ -55,6 +55,7 @@ test_images:
 	sudo -E ./tests/test_cli.sh tests/cli/test_compose_ext4-filesystem.sh  \
 				    tests/cli/test_compose_partitioned-disk.sh \
 				    tests/cli/test_compose_tar.sh              \
+				    tests/cli/test_compose_tar_kickstart.sh    \
```
atodorov: FYI not really needed. In fact this target is being removed in #885.

tests/cli/lib/lib.sh
```diff
@@ -16,14 +16,14 @@ if [ "$COMPOSER_TEST_FAIL_FAST" == "1" ]; then
   }
 fi
 
-export QEMU_BIN="/usr/bin/qemu-system-$(uname -m)"
+export QEMU_BIN="/usr/libexec/qemu-kvm"
```
atodorov: this change should go together with the 1st commit in the sequence

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,139 @@
...
+        rlAssertEquals "SELinux operates in enforcing mode" "$(getenforce)" "Enforcing"
+        uuid=$($CLI compose start $bp_name tar)
+        rlAssertEquals "exit code should be zero" $? 0
+
+        uuid=$(echo $uuid | cut -f 2 -d' ')
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose finished"
+        if [ -n "$uuid" ]; then
+            until $CLI compose status | grep -E "$uuid (FINISHED|FAILED)"; do
```
atodorov: Missing `\` before `|`. (Same error also on rhel7-extras branch).

Makefile
jikortus: OK, I'll get rid of it.

tests/cli/lib/lib.sh
jikortus: OK, will move that to the 1st commit.

tests/cli/test_compose_tar_kickstart.sh
jikortus: I fail to see why the backslash should be put there? If it was just `grep` without `-E`, I'd need to escape the parentheses and pipe, but as it is an extended regexp, this is not the case. Or is it something else you are getting at?

atodorov: Hmm, I didn't actually try to see if the backslash was required. I just noticed that was different from the other code snippets that reference the same code. There is actually a helper function but if you want to use it then you'll have to cherry-pick it from master which is outside the scope of this change. 

You can still do it if you want. IMO having the test suite not diverge too much between branches is a good thing.

jikortus: You're right that we should strive for the lowest divergence possible across branches. I had a look at the state of the aforementioned piece of code checking for compose finish on master is different from `rhel7-extras` and `rhel8-branch`, plus that the function you mentioned hasn't been backported from master. I totally agree that using the `wait_for_compose` function is the best approach, however, at this state, I'd rather consider backporting this in different PRs (both for `rhel7-extras` and `rhel8-branch`) later on.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -0,0 +1,139 @@
...
+name = "sudo"
+version = "*"
+
+[[packages]]
+name = "audit"
+version = "*"
+
+[[groups]]
+name = "anaconda-tools"
+version = "*"
```
atodorov: Also missing package kernel which is stated as required in the description of this test case. The variant which was on master branch was missing anaconda-tools+kernel and was running into some issues. Retest is still in progress (#887)

tests/cli/test_compose_tar_kickstart.sh
atodorov: While testing https://github.com/weldr/lorax/pull/887 on Fedora 31 the compose itself failed and b/c it wasn't using the `wait_for_compose()` function we didn't have logs. I've made the necessary change and will open tickets to cherry-pick these changes for the rhel branches.

bcl:
Should we backport #871 first?

tests/cli/test_compose_tar_kickstart.sh
jikortus: The test works for me as it is, without `kernel` package. IIRC the need for `anaconda-tools`/`kernel` differs across the main branches, doesn't it? Also most probably `audit` should go away if #871 is backported, as Brian suggests.

atodorov:
> Should we backport #871 first?

No. #871 modifies the file ` tests/cli/test_compose_tar_kickstart.sh` (among others) but this PR adds it in the first place. #883 should be reviewed and merged first then #871 and #885 should be backported downstream.

atodorov:
@jikortus given the latest changes in 871 and subsequent pull requests I think the following is a reasonable course of action:

1) Drop the last commit "tests: Increase test VM memory to 3 GB" from this pull request b/c nested virt seems to fail anyway (so although this change is fine it doesn't matter in the long run) and we'll start the backport immediately.

2) No other changes to the PR, will merge the rest as is b/c the backport of 871 will change it anyway. 

As for needing kernel & other packages: 871 adds what it needs in the blueprint and IMO we should keep it the same during the packport. This is also the reason we can merge as is ATM.


What do you think ?

jikortus:
@atodorov: OK, what you are suggesting makes sense to me. I dropped the commit "tests: Increase test VM memory to 3 GB" and will work on backporting #871 after this PR is merged.

PR #885, opened 2019-10-23
WIP: Addons after 871 changes about qcow2/live-iso

atodorov:
This is mostly documentation and clean up of the scripts & trying to reuse existing functions.


Makefile
```diff
@@ -56,31 +56,6 @@ test:
...
-test_azure:
-	sudo -E ./tests/test_cli.sh tests/cli/test_build_and_deploy_azure.sh
-
-test_openstack:
-	sudo -E ./tests/test_cli.sh tests/cli/test_build_and_deploy_openstack.sh
-
-test_vmware:
-	sudo -E ./tests/test_cli.sh tests/cli/test_build_and_deploy_vmware.sh
-
-test_cli:
```
bcl: This test is actually useful when running in podman. I usually run `RUN_TESTS="ci test_cli" DOCKER=podman make test-in-docker` to test my changes locally.

bcl:
Rebase this on master which now includes #871 and go from there.

tests/test_cli.sh
```diff
@@ -105,19 +106,9 @@ if [ -e "/var/tmp/test-results" ]; then
...
-else
-    # execute other cli tests which need more adjustments in the calling environment
-    # or can't be executed inside Travis CI
-    for TEST in "$@"; do
-        $TEST
-    done
-fi
+setup_beakerlib_env
 
+run_beakerlib_tests "$@"
```
atodorov: @bcl with this change `make test_cli` changes behavior and that will affect how you run things locally. Do you want to have the previous if-else block back ?

bcl: No, when I'm running it locally I pass in the list of tests via the RUN_TESTS variable in the makefile so it will be set.

PR #886, opened 2019-10-23
tests: Disable mirrors

atodorov:


atodorov:
```
pyanaconda.errors.NonInteractiveError: Non interactive installation failed: Problems in request:
missing packages: systemd-243-4.gitef67743.fc31.x86_64, systemd-libs-243-4.gitef67743.fc31.x86_64, systemd-pam-243-4.gitef67743.fc31.x86_64, systemd-rpm-macros-243-4.gitef67743.fc31.noarch
```

I'm beginning to wonder if we switched prematurely to Fedora-31.

bcl:
> I'm beginning to wonder if we switched prematurely to Fedora-31.

That looks suspiciously like a problem I hit in the container. The mirrors are not very good at staying in sync, and if composer hits one mirror and anaconda hits a different mirror, you will see depsolve fail. The safest thing is to not use mirrors in Fedora.

test/vm.install
```diff
@@ -17,6 +17,10 @@ if ! rpm -q beakerlib; then
     fi
 fi
 
+# disable mirrors & re-enable direct download
+sed -i "s/metalink.*//" /etc/yum.repos.d/*.repo
+sed -i "s/#\(baseurl=.*\)/\1/" /etc/yum.repos.d/*.repo
```
bcl: That's a good start, but on Fedora at least, the url they have in there by default is backed by some sort of random/round-robin/load-balancing DNS so you still don't get the same server every time. When I really want the exact same thing I use dl.fedoraproject.org instead of download.fedoraproject.org

bcl:
Rebase this on master which now includes #871 and see if we can get the repo behavior to be consistent.

PR #887, opened 2019-10-23
Unskip tar kickstart test

atodorov:
b/c we've switched to using Fedora-31 for testing and the known
bugs should be fixed

### DIFF
--- a/tests/cli/test_compose_tar_kickstart.sh
+++ b/tests/cli/test_compose_tar_kickstart.sh
@@ -67,15 +67,7 @@ __EOF__
     rlPhaseEnd
 
     rlPhaseStartTest "compose finished"
-        if [ -n "$uuid" ]; then
-            until $CLI compose info $uuid | grep 'FINISHED\|FAILED'; do
-                sleep 60
-                rlLogInfo "Waiting for compose to finish ..."
-            done;
-            check_compose_status "$UUID"
-        else
-            rlFail "Compose uuid is empty!"
-        fi
+        wait_for_compose "$uuid"
 
         rlRun -t -c "$CLI compose image $uuid"
         image="$uuid-root.tar.xz"
--- a/tests/cli/test_compose_tar_kickstart.sh
+++ b/tests/cli/test_compose_tar_kickstart.sh
@@ -37,6 +37,13 @@ description = "tar image test"
 version = "0.0.1"
 modules = []
 
+[[groups]]
+name = "anaconda-tools"
+
+[[packages]]
+name = "kernel"
+version = "*"
+
 [[packages]]
 name = "openssh-server"
 version = "*"
--- a/tests/cli/test_compose_tar_kickstart.sh
+++ b/tests/cli/test_compose_tar_kickstart.sh
@@ -104,8 +104,10 @@ __EOF__
         mv $image $image_path
         restorecon $image_path
         rlLogInfo "Starting installation from tar image in a VM"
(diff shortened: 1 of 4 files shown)


atodorov:
On a local Fedora 31 system I see:
```
Oct 23 14:04:42 hpe-dl360g5-01.hpe2.lab.eng.bos.redhat.com dockerd[14246]: Error starting daemon: Devices cgroup isn't mounted
Oct 23 14:04:42 hpe-dl360g5-01.hpe2.lab.eng.bos.redhat.com systemd[1]: docker.service: Main process exited, code=exited, status=1/FAILURE
Oct 23 14:04:42 hpe-dl360g5-01.hpe2.lab.eng.bos.redhat.com systemd[1]: docker.service: Failed with result 'exit-code'.
Oct 23 14:04:42 hpe-dl360g5-01.hpe2.lab.eng.bos.redhat.com systemd[1]: Failed to start Docker Application Container Engine.
```

which currently breaks the first test (unrelated to changed here). Documented solution reports changing grub command line options:
https://stackoverflow.com/questions/32002882/error-starting-docker-daemon-on-ubuntu-14-04-devices-cgroup-isnt-mounted

Not sure if we need anything else before running the tar test. Any pointers ?

bcl:
How about we switch to using podman? Since this is running on F31 and not in travis, there's no reason to depend on docker for testing it.

atodorov:
> How about we switch to using podman? Since this is running on F31 and not in travis, there's no reason to depend on docker for testing it.

It doesn't depend on where this is running. The original test case called for importing the tar into Docker as a quick way to make an image and run a container from it. On some branches we did have docker proper (now called moby-engine) while on others we used podman-docker and this is what I switched to.

atodorov:
Looks like we're having issues with booting the test VM at the end (tar & qcow2) - either nested virt started acting up or network doesn't come up in the VM. Trying to debug locally what is going on.

atodorov:
> Trying to debug locally what is going on.

I see qemu trying to boot the generated disk.img, I see boot menu prompt and then 
```
Probing EDD (edd=off to disable) ... ok
```

and nothing else.

bcl:
Try changing the test liveimg kickstart to use `autopart --type plain`, the root.tar.xz doesn't contain the lvm tools so it has no way to boot from lvm. Users have to make sure that they have installed all the needed tools in their tar if they are going to be using liveimg with it.

atodorov:
@bcl: I've added `kernel` & `anaconda-tools` which were missing but originally stated as required in the description of this test case. Let's see how this goes.

atodorov:
@bcl  even after adding anaconda-tools group in the blueprint, which contains lvm2 as mandatory package I get a failure. The logs at:
https://209.132.184.41:8493/logs/pull-887-20191029-151616-2fac5ff5-weldr-lorax--fedora-31-tar/
show that lvm2 wasn't installed. e2fsprogs either.

I'm not quite certain what's going on right now.

Unrelated: I think my commit 2fac5f is not what you meant above so I will remove it and give it a try again.

tests/cli/test_compose_tar_kickstart.sh
```diff
@@ -37,6 +37,14 @@ description = "tar image test"
 version = "0.0.1"
 modules = []
 
+[[groups]]
+name = "anaconda-tools"
+version = "*"
```
bcl: groups have no version, but it also shouldn't really matter, it will just be ignored.

bcl:
Fixed by PR #871

PR #891, opened 2019-10-29
Disable some compose types on other architectures

bcl:
The 'enabled' field in the /compose/types output now reflects whether or
not the type is supported on the current architecture. Disabled types
are not allowed to be built, and will raise an error like:

Compose type 'alibaba' is disabled on this architecture

Resolves: rhbz#1751998


bcl:
Note: tar test is failing because pr #883 hasn't been merged yet. It is still run as part of the generic image tests.

PR #895, opened 2019-11-06
composer-cli: Return int from handle_api_result not bool

bcl:
The callers, and the documentation, all expect int 0/1 to use as the
exit status for the program. Not True/False, even though that works most
of the time.


bcl:
Merged to master

PR #897, opened 2019-11-08
Backport of nested virt related changes for RHEL 7

atodorov:
Very similar to #866 but for the other branch.

### DIFF
--- a/tests/test_cli.sh
+++ b/tests/test_cli.sh
@@ -78,6 +78,10 @@ else
     composer_start
 fi
 
+# Clean out the test-results directory
+if [ -e "/var/tmp/test-results" ]; then
+    rm -rf "/var/tmp/test-results"
+fi
 
 export BEAKERLIB_JOURNAL=0
 export PATH="/usr/local/bin:$PATH"
--- a/tests/test_cli.sh
+++ b/tests/test_cli.sh
@@ -3,6 +3,8 @@
 
 set -eu
 
+. $(dirname $0)/cli/lib/lib.sh
+
 export BEAKERLIB_DIR=$(mktemp -d /tmp/composer-test.XXXXXX)
 CLI="${CLI:-}"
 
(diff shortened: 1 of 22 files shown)


bcl:
Looks like it needs a bit of adjustment for rhel7 :)

tests/test_cli.sh
```diff
@@ -13,6 +16,22 @@ function setup_tests {
     # explicitly enable sshd for live-iso b/c it is disabled by default
     # due to security concerns (no root password required)
     sed -i.orig 's/^services.*/services --disabled="network" --enabled="NetworkManager,sshd"/' $share_dir/composer/live-iso.ks
+
+    # Make the live-iso boot more quickly (isolinux.cfg)
+    for cfg in "$share_dir"/templates.d/99-generic/live/config_files/*/isolinux.cfg; do
```
atodorov: @bcl where are these files located on RHEL 7 ? I'm not able to find them.

bcl: `./share/live/config_files/x86/isolinux.cfg` -- rhel7 doesn't support adding new templates so it has a slightly shorter directory tree.

atodorov:
@bcl can you check the live-iso output. I get this:
```
2019-11-11 10:01:44,224: Problem depsolving with-ssh: The following package(s) had problems: grub2-efi-*-cdboot (No package name matching grub2-efi-*-cdboot)
```

while in the log you can see I have the optional repo enabled and you can see yum finds these packages.

I remember there was a bug in this space but wasn't that fixed ?

bcl:
> I remember there was a bug in this space but wasn't that fixed ?

In RHEL8 I moved those into live-install.tmpl because they are arch-dependent, but that's not the problem here. Ah, I think I see it:

commit b38a358ea3c15542dfa07573850df9158c8b3ca9 adds the optional repo, but doesn't restart the service, so composer doesn't know about it. It would probably be better to just set it up before the service is started.

atodorov:
@bcl in light of the latest problems with live-iso I've left the `@skip` decorator. All tests should report PASS now and this PR is ready to review & merge. I'll continue working on live-iso separately.

atodorov:
@bcl this PR is ready for review. vmware failure is unrelated.

@chris1984 cc, vmware started acting up again. I will try testing with the env you provided to me.

PR #899, opened 2019-11-09
lorax-composer: Switch vhd image to use cloud-init

bcl:
This switches the vhd image to use cloud-init instead of WALinuxAgent.
Also limits the cloud-init datasource to Azure

Resolves: rhbz#1754711


yuxisun1217:
Hi,

Please consider to add cloud-utils-growpart package in, which is for growpart feature of cloud-init. It's not a dependency of cloud-init package so it cannot be installed automatically follow cloud-init.

bcl:
cherry-picked to master and rhel8-branch

PR #900, opened 2019-11-11
Add dmidecode on supported architectures

bcl:
Currently supported on i386, x86_64, aarch64

Related: rhbz#1714793


bcl:
Merged.

PR #908, opened 2019-11-21
Always remove RPMs before installing them & allow using arbitrary repositories, incl. system repos

atodorov:
1) Remove whatever packages happen to be there in the image so they can be installed later from the locally built RPMs

2) I also want to adjust Makefile (vm-local-repos)/vm.install so that when testing downstream snapshots we don't build & install from source but instead install from the repos copied from the host machine.

cc @bcl, @martinpitt, @jikortus

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ ci: check test
 
 $(VM_IMAGE): TAG=HEAD
 $(VM_IMAGE): srpm bots
+	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
 	srpm=$(shell rpm --qf '%{Name}-%{Version}-%{Release}.src.rpm\n' -q --specfile lorax.spec | head -n1) ; \
 	bots/image-customize -v \
 		--resize 20G \
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ endif
 export TEST_OS
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 
+ifeq ($(REPOS_DIR),)
+REPOS_DIR = /etc/yum.repos.d
+endif
+
 default: all
 
 src/composer/version.py: lorax.spec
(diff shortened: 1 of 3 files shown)


jikortus:
@atodorov I like this idea. On top of that, I also thought about implementing some way (a new makefile target?) to use some arbitrary repo file instead of system repos, which would allow for running the tests on a host with different system and have the image in a state very close to a tested compose (e. g. running tests for RHEL-7.x particular compose on a Fedora system). What do you think about that?

atodorov:
> On top of that, I also thought about implementing some way (a new makefile target?) to use some arbitrary repo file instead of system repos

Good idea, refreshed latest commit. You can configure the `REPOS_DIR` variable in the same way you configure `TEST_OS` before calling `make vm-local-repos`. See README.

Makefile
```diff
@@ -165,7 +169,13 @@ vm: $(VM_IMAGE)
 # sure VM_IMAGE is as close as possible to the host!
 vm-local-repos: vm
 	bots/image-customize -v \
-		--upload /etc/yum.repos.d:/etc/yum.repos.d/ \
+		--run-command "rm -rf /etc/yum.repos.d" \
+		$(TEST_OS)
+	bots/image-customize -v \
```
martinpitt: There's no need to call it twice. This means an extra shutdown and boot for no reason, you can specify arbitrarily many command line options to image-customize.

Makefile
```diff
@@ -165,7 +169,13 @@ vm: $(VM_IMAGE)
...
+		--upload $(REPOS_DIR):/etc/yum.repos.d \
+		--run-command "yum -y remove composer-cli lorax-composer" \
+		--run-command "yum -y install composer-cli lorax-composer" \
+		--run-command "systemctl enable lorax-composer" \
 		--run-command "yum -y update" \
```
martinpitt: I'd recommend to first remove, then update, then install. Otherwise the update may change the version again to something unintended?

Makefile
atodorov: Actually there is. When I group the options into 1 command first new repos are uploaded and then the `rm` is executed deleting the newly uploaded files. We want to make sure that before continuing with the tests the VM contains only the specified repositories and nothing else!

martinpitt: OK -- you could upload it to somewhere else, and move it afterwards, but if you don't mind the extra waiting this works as well. We recently discussed that in PM about whether i-customize should always start fresh or use an existing overlay (the latter happens now)

atodorov:
I did some changes to this PR since it was initially approved. Built a VM with the host repos manually and verified that tests still pass. @jikortus independently verified the patches work fine on the rhel7-extras branch.

This is ready for final review, merge & backport.

test/vm.install
```diff
@@ -2,6 +2,10 @@
 
 SRPM="$1"
 
+# always remove older versions of these RPMs if they exist
+# to ensure newly built packages have been installed
+yum -y remove lorax lorax-composer composer-cli
```
bcl: Does this make the `rpm -e ...` further down obsolete? Seems like it's doing the same thing in a less general way.

atodorov: Indeed you are right. I've removed the rpm -e and left the explicit yum remove

PR #910, opened 2019-11-22
tests: enable compose_ext4 test to use CDN repos

jrusz:
This enables to run compose_ext4 test with CDN repos only. The idea is that the system gets subscribed against stage candlepin and consumes from production CDNs if the required environmental variables are set.

### DIFF
--- a/tests/cli/test_compose_ext4-filesystem.sh
+++ b/tests/cli/test_compose_ext4-filesystem.sh
@@ -16,8 +16,22 @@ set -e
 
 CLI="${CLI:-./src/bin/composer-cli}"
 
-
 rlJournalStart
+    rlPhaseStartSetup
+        if [[ -n "$SUBSCRIPTION_HOSTNAME" && -n "$SUBSCRIPTION_USERNAME" && -n "$SUBSCRIPTION_PASSWORD" ]]; then
+            rlRun -t -c "cp -r /etc/yum.repos.d /etc/yum.repos.backup"
+            rlRun -t -c "rm -f /etc/yum.repos.d/*.repo"
+            rlRun -t -c "subscription-manager config --server.hostname=$SUBSCRIPTION_HOSTNAME"
+            subscription-manager register --username=$SUBSCRIPTION_USERNAME --password=$SUBSCRIPTION_PASSWORD
+            rlAssert0 "'subscription-manager register' succeeded" $?
+            rlRun -t -c "subscription-manager attach --auto"
+            rlRun -t -c "systemctl restart lorax-composer"
+            rlLogInfo "System is subscribed"
+        else
+            rlLogInfo "System is not subscribed"
+        fi
+    rlPhaseEnd
+
     rlPhaseStartTest "compose start"
         rlAssertEquals "SELinux operates in enforcing mode" "$(getenforce)" "Enforcing"
         UUID=`$CLI compose start example-http-server ext4-filesystem`
@@ -32,6 +46,11 @@ rlJournalStart
 
     rlPhaseStartCleanup
         rlRun -t -c "$CLI compose delete $UUID"
+        if [[ -n "$SUBSCRIPTION_HOSTNAME" && -n "$SUBSCRIPTION_USERNAME" && -n "$SUBSCRIPTION_PASSWORD" ]]; then
+            rlRun -t -c "subscription-manager unregister"
+            rlRun -t -c "rm -rf /etc/yum.repos.d"
+            rlRun -t -c "mv /etc/yum.repos.backup /etc/yum.repos.d"
+        fi
     rlPhaseEnd
 
 rlJournalEnd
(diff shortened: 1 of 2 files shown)


tests/cli/test_compose_ext4-filesystem.sh
```diff
@@ -16,6 +16,13 @@ set -e
 
 CLI="${CLI:-./src/bin/composer-cli}"
 
+if [[ -v SUB_HOSTNAME && -v SUB_USER && -v SUB_PASS ]]; then
```
atodorov: Change to `SUBSCRIPTION_`, no need to be terse when you can be verbose.

Also make sure you are actually using variables, e.g. `$SUBSCRIPTION_HOSTNAME`.

What does `-v` do? I can't find it documented and I've never seen it before?

atodorov: Maybe also `rlLogInfo` because I can't tell if the current test execution used subscriptions or not.

jrusz: > What does `-v` do? I can't find it documented and I've never seen it before?
I found it here https://www.tutorialkart.com/bash-shell-scripting/check-if-variable-is-set/  It works but like you said it does not seem to be documented.  I'll change it to `-z` or `-n`

bcl:
What about if they aren't set? Shouldn't this be a separate test?

martinpitt:
The failed tests ran on the temporary AWS instance of cockpit CI. This was an experiment, and I'll roll back to the e2e cluster later today. It should structurally look like the others, so I'd appreciate if you could have a look why they failed. If it was just a flake, then feel free to retry, of course.

(I may have an idea what's going wrong)

martinpitt:
Actually, the rhel8-1/azure test ran on AWS, and that succeeded.

tests/cli/test_compose_ext4-filesystem.sh
```diff
@@ -16,8 +16,21 @@ set -e
...
 CLI="${CLI:-./src/bin/composer-cli}"
 
-
 rlJournalStart
+    rlPhaseStartSetup
+        if [[ -n $SUBSCRIPTION_HOSTNAME && -n $SUBSCRIPTION_USERNAME && -n $SUBSCRIPTION_PASSWORD ]]; then
+            rlRun "mkdir /etc/yum.repos.d/disabled"
+            rlRun "mv /etc/yum.repos.d/*.repo /etc/yum.repos.d/disabled"
+            rlRun "subscription-manager config --server.hostname=$SUBSCRIPTION_HOSTNAME"
+            rlRun "subscription-manager register --username=$SUBSCRIPTION_USERNAME --password=$SUBSCRIPTION_PASSWORD"
```
atodorov: This is exposing the username & password in the logs. See current log. Run the command without rlRun and examine the exit code to figure out if it succeeded or not.

tests/cli/test_compose_ext4-filesystem.sh
```diff
@@ -16,8 +16,21 @@ set -e
 
 CLI="${CLI:-./src/bin/composer-cli}"
 
-
 rlJournalStart
+    rlPhaseStartSetup
+        if [[ -n $SUBSCRIPTION_HOSTNAME && -n $SUBSCRIPTION_USERNAME && -n $SUBSCRIPTION_PASSWORD ]]; then
+            rlRun "mkdir /etc/yum.repos.d/disabled"
+            rlRun "mv /etc/yum.repos.d/*.repo /etc/yum.repos.d/disabled"
```
atodorov: FYI you can copy the entire directory and make the backup in a single operation.

tests/cli/test_compose_ext4-filesystem.sh
```diff
@@ -32,6 +45,11 @@ rlJournalStart
 
     rlPhaseStartCleanup
         rlRun -t -c "$CLI compose delete $UUID"
+        if [[ -n $SUBSCRIPTION_HOSTNAME && -n $SUBSCRIPTION_USERNAME && -n $SUBSCRIPTION_PASSWORD ]]; then
+            rlRun -t -c "subscription-manager unregister"
+            rlRun -t -c "mv -f /etc/yum.repos.d/disabled/*.repo /etc/yum.repos.d/"
+            rlRun -t -c "rm -rf /etc/yum.repos.d/disabled"
```
atodorov: Same here, wrt restoring. You can overwrite the entire directory in a single operation. Probably it would be easier if both locations are on the same level in the file system.

atodorov:
> so I'd appreciate if you could have a look why they failed

all of the failures were related to jobs that need to talk to internal machines. I suspect the AWS infra isn't connected to VPN.

atodorov:
> What about if they aren't set? Shouldn't this be a separate test?

@bcl these variables are used only in 1 of the existing scripts so the rest of them still do not connect to CDN. IMO we're good here.

atodorov:
I've adjsuted this PR to resolve the comments from code review and also to reference a rhbz number in the commit logs so we can merge. Will merge once CI reports pass.

PR #912, opened 2019-11-26
Backport PR908 to RHEL 8

atodorov:
We need these changes here b/c snapshots testing happens on downstream branches, not on master.

PR #915, opened 2019-11-27
tests: If TEST_OS isn't specified then match the host OS

atodorov:
this will help with downstream snapshots testing making
it easier to match the host OS snapshot

Cherry-picked from 6839390be262f5a51f17e1b8d728f88106385f39

Related: rhbz#1769525

This is an addition to #912

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,9 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 IMAGE_RELEASE = rhel8-latest
 
 ifeq ($(TEST_OS),)
-TEST_OS = rhel-8-1
+OS_ID = $(shell awk -F= '/^ID=/ {print $$2}' /etc/os-release)
+OS_VERSION = $(shell awk -F= '/^VERSION_ID/ {print $$2}' /etc/os-release | tr '.' '-')
+TEST_OS = $(OS_ID)-$(OS_VERSION)
 endif
 export TEST_OS
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)


Makefile
```diff
@@ -10,7 +10,9 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 IMAGE_RELEASE = rhel8-latest
 
 ifeq ($(TEST_OS),)
-TEST_OS = rhel-8-1
+OS_ID = $(shell awk -F= '/^ID/ {print $$2}' /etc/os-release)
```
jikortus: The `^ID` expression matches to `ID` and `ID_VERSION` at the same time, so it needs to be more specific to just match only `ID` (`/^ID=/` should be enough).

jikortus: I can see this came from master branch, but this issue doesn't exist there as `/etc/os-release` in Fedora doesn't contain the `ID_LIKE` variable.

Makefile
```diff
@@ -10,7 +10,9 @@ TAG = lorax-$(VERSION)-$(RELEASE)
 IMAGE_RELEASE = rhel8-latest
 
 ifeq ($(TEST_OS),)
-TEST_OS = rhel-8-1
+OS_ID = $(shell awk -F= '/^ID=/ {print $$2}' /etc/os-release)
+OS_VERSION = $(shell awk -F= '/^VERSION_ID/ {print $$2}' /etc/os-release)
+TEST_OS = $(OS_ID)-$(OS_VERSION)
```
martinpitt: Note that a much simpler approach  is to use something like `$(shell . /etc/os-release; echo "$$ID-$$VERSION_ID")`; but I don't want to quibble over such implementation details, particularly because this is a backport (i. e. if you want to change it, change it in master first).

However, this breaks on our RHEL images. For hysterical raisins they are *not* called "rhel-8.2", but "rhel-8-2". I. e. you need to add an additional `tr '.' '-'` (or some equivalent `sed`).

atodorov: fixed

atodorov:
> Does the master branch need a similar fix, as this was backported?

@martinpitt it's backported

PR #917, opened 2019-11-29
Test updates before switching to osbuild-composer backend

atodorov:
FTR the idea here is to keep this test suite active in the beginning and just swap lorax-composer with osbuild-composer. 

The benefit is that the test suite matches the existing test plan, everything is fully automated and we've got CI, creds, etc all available.

For now we are going to change the BACKEND only when doing downstream testing!


tests/test_cli.sh
```diff
@@ -10,8 +10,9 @@ set -eu
 CLI="${CLI:-}"
 
 function setup_tests {
+    [ "$BACKEND" == "osbuild-composer" ] && return 0
```
bcl: Related to all the above changes from example-http-server.toml to test-http-server.toml wouldn't it be cleaner to instead have osbuild-composer push the copy of the expected example blueprints (example-http-server, example-development, example-atlas) from tests/pylorax/blueprints/? That way there's less change to account for.

atodorov: In the general case of working with osbuild-composer we don't have access to the pylorax source code so I'd say no. It can be done ATM but will require changes very soon IMO.

atodorov:
@martinpitt it looks like CI slaves don't have access to internal resources. Can you confirm ? (e.g. slaves are running in another data center)

larskarlitski:
Added three more commits, which together with the fixes in https://github.com/osbuild/osbuild-composer/pull/156 make most of the tests run through cleanly with `osbuild-composer`.

There's a minor issue with the tar test, which needs https://github.com/osbuild/osbuild/pull/178

atodorov:
@larskarlitski I've added more stuff into your last 3 commits and/or adjusted them because they were missing some of the scenarios and/or broke CI. Should be better now. I will retest locally with osbuild-composer and adjust if needed.

atodorov:
This should be now the same as #1007. Still needs:

- an approved rhbz
- the osbuild-composer RPM inside the latest 8.3 images so we can test it

atodorov:
@bcl I am seeing: `RHSM secrets not found on host`  which is causing `composer-cli compose start` to fail. Doesn't seem related to the current changes though.

atodorov:
@bcl I've added I more commit to override the repositories configuration and that seems to resolve the RHSM issue. IMO this is ready to be merged. Please review.

PR #918, opened 2019-12-04
Add API v1 commands to composer-cli

bcl:
Tested against RHEL8 and rawhide lorax-composer. Needs testing against osbuild-composer


bcl:
Pushed to rhel8-branch

PR #919, opened 2019-12-05
mkksiso: Add the option to set the ISO volume label

flyingflo:
To distinguish from the base images, a different 
volume label is useful. Add a commandline option, 
pass it to the iso tool, and replace the occurences 
of the label in isolinux and grub config files.

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -16,6 +16,7 @@ URL:            https://github.com/weldr/lorax
 Source0:        %{name}-%{version}.tar.gz
 
 BuildRequires:  python3-devel
+BuildRequires:  make
 
 Requires:       lorax-templates
 
--- a/src/sbin/mkksiso
+++ b/src/sbin/mkksiso
@@ -368,7 +368,7 @@ MkisofsTools = [Mkisofs_aarch64, Mkisofs_ppc, Mkisofs_ppc64le, Mkisofs_s390, Mki
 
 
 class MakeKickstartISO():
-    def __init__(self, ks, input_iso, output_iso, add_paths, cmdline=""):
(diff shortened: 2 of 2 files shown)


flyingflo:
Tested on x86_84 with Centos 8, BIOS and UEFI.
I couldn't verify the change in `_edit_ppc`, though.

bcl:
I understand wanting to do that, but the problem you will run into is that it now assumes inst.stage2= formatting that isn't always consistent between releases/distributions. I'm trying to keep the changes as simple and generic as possible so that this doesn't end up being a big pile of tweaks like livecd-iso-to-disk did :)

flyingflo:
Ok, instead of assuming the `inst.stage2=` prefix, we could replace the volume label regardless of the prefix. But, I think I understand that you just don't want such tweaks at all, right?

chris1984:
ok to test

bcl:
> Ok, instead of assuming the `inst.stage2=` prefix, we could replace the volume label regardless of the prefix. But, I think I understand that you just don't want such tweaks at all, right?

oh, good idea! There's really no need to change the inst.stage.* part at all, you're just wanting to use a new volid so yes, that would be just fine with me.

flyingflo:
> oh, good idea! There's really no need to change the inst.stage.* part at all, you're just wanting to use a new volid so yes, that would be just fine with me.

I simplified the patch so search only for the original label and replace it. It doesn't assume `inst.stage2=`, but it assumes the volume label to be somewhere in these files.

flyingflo:
The second patch `spec: Add missing BuildRequires: make`  is just a minor fix.
Most builders will already have make, and won't notice.

bcl:
Thanks! Merged to master with an additional commit to update the documentation in mkksiso.rst

PR #920, opened 2019-12-06
lorax-composer: Change cockpit.socket examples to cockpit

bcl:
The RHEL7 version of Anaconda only allows .service units to be used with
the kickstart services command. Update the documentation, examples, and
tests to reflect this.

Related: rhbz#1718473


bcl:
Tests mostly passed, vmware and openstack issues are unrelated to the changes.

PR #921, opened 2019-12-10
composer-cli: Only display the available compose types

bcl:
The enabled bool is now being used so the cli should only show the types
actually available on the architecture.

Also modifies the test in test_compose_sanity.sh

Related: rhbz#1751998


bcl:
Merging this despite openstack and aws failures which looks like infrastructure issues not related to this change.

PR #923, opened 2019-12-12
Rhel7 cockpit ci

bcl:


atodorov:
@bcl do we need to port the serial console changes to master & rhel8-branch? Looks like they work without these changes.

bcl:
> @bcl do we need to port the serial console changes to master & rhel8-branch? Looks like they work without these changes.

no, this appears to be a problem/quirk of the RHEL7 environment. I think it may be the bootloader hanging because of the virt environment that the CI sets up. Newer versions have not had these problems.

As for adding the serial commands everywhere I did that so they are consistent, and to make it easier to debug problems in the future (you can manually edit the test to copy the image on the side and then boot with qemu and serial. It's easier if serial is already setup).

PR #925, opened 2019-12-13
Fix typo in warning and add missing requirement

bcl:


bcl:
Fixed title in commit, should be python3-pytoml. Not sure when/if github will pick it up.

PR #926, opened 2019-12-13
lorax: Remove vmlinuz from install.img /boot

bcl:
The kernel in /boot is not needed. Keep the .vmlinuz*hmac file so that
fips mode can check it (this requires an updated dracut, see
https://github.com/dracutdevs/dracut/pull/696).

Related: rhbz#1782737

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -41,7 +41,7 @@ Requires:       xz-lzma-compat
 Requires:       xz
 Requires:       pigz
 Requires:       pbzip2
-Requires:       dracut >= 030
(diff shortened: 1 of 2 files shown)


bcl:
Note that this is currently only for comments and testing -- I am not 100% sure that fips is the only user of the kernel in /boot of the install.img, need to make sure before removing it.

bcl:
The dracut change has been merged upstream.

sharkcz:
LGTM, the installer "chroots" into the install.img, so kernel (vmlinuz) is not needed there

PR #927, opened 2019-12-13
Make the boot.iso smaller

bcl:


jwboyer:
We should hold off on merging the wireless change.  That's not confirmed yet.

share/runtime-cleanup.tmpl
```diff
@@ -344,6 +344,9 @@ removefrom redhat-logos /usr/share/{plymouth,backgrounds}*
 removefrom subscription-manager --allbut /etc/rhsm/* /usr/share/rhsm/* /var/log/rhsm/* \
     /etc/yum/pluginconf.d/product-id.conf /usr/lib/yum-plugins/product-id.py
 
+## OSCAP cleanups
+removefrom scap-security-guide --allbut /usr/share/xml/scap/ssg/content/ssg-rhel7-ds.xml
+
```
jwboyer: Could add:

removefrom scap-security-guide /usr/share/scap-security-guide/

as well.

bcl: That already removes those files as well. --allbut removes every file in the packages except those matching the argument glob.

bcl:
Closing this one, the oscap files were removed but the firmware left as is.

PR #933, opened 2020-01-03
Add test for canceling a running compose

jrusz:
This is a modification of an existing test for compose sanity. This should now create the same test scenario like in https://bugzilla.redhat.com/show_bug.cgi?id=1656691

### DIFF
--- a/tests/cli/test_compose_sanity.sh
+++ b/tests/cli/test_compose_sanity.sh
@@ -22,6 +22,26 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "compose start"
+        UUID=`$CLI compose start example-http-server ami`
+        rlAssertEquals "exit code should be zero" $? 0
+        UUID=`echo $UUID | cut -f 2 -d' '`
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose info $UUID | grep 'RUNNING'; do
+                sleep 20
+                rlLogInfo "Waiting for compose to start running..."
+            done;
+        else
+            rlFail "Compose UUID is empty!"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "cancel compose"
+        rlRun -t -c "$CLI compose cancel $UUID"
+        rlRun -t -c "$CLI compose info $UUID" 1 "compose is canceled"
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose start again"
         UUID=`$CLI --test=2 compose start example-http-server tar`
         rlAssertEquals "exit code should be zero" $? 0
 


tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,30 @@ rlJournalStart
...
+        sleep 20
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose info"
+        if [ -n "$UUID" ]; then
+            rlRun -t -c "$CLI compose info $UUID | egrep 'RUNNING|WAITING'"
+        else
+            rlFail "Compose UUID is empty!"
+        fi
+    rlPhaseEnd
```
atodorov: this section is not needed. Instead it should be moved above and executed within a loop until the test detects a RUNNING status. Should look like `wait_for_compose` function (or even reuse the function by passing parameter RUNNING)

tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,30 @@ rlJournalStart
...
+
+    rlPhaseStartTest "cancel compose"
+        `$CLI compose cancel $UUID`
```
atodorov: should execute via `rlRun -t -c` if this returns exit code 0. As-is it doesn't make sense to quote in backticks b/c exit code is not checked.

tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,30 @@ rlJournalStart
...
+        `$CLI compose status | grep $UUID`
+        rlAssertNotEquals "compose is canceled" $? 0
+        rlAssertNotExists "/var/lib/lorax/composer/results/$UUID/"
```
atodorov: We're going to have to remove the last assert when migrating to the osbuild-composer backend so better to remove it now. Just assert on current behavior which is that the compose UUID is nowhere to be found and the exit codes of the commands above. That is sufficient to verify the original bug which this test covers.

jrusz: Right, makes sense, so I'm just checking that that the compose UUID is not there when running composer-cli status now.

tests/cli/test_compose_sanity.sh
jrusz: This returns and exit code 0 only if I cancel it right after starting the compose. If I wait for ~20 seconds the cancel takes longer and does not exit with 0. I can make the cancel either way but I am not sure which approach covers the scenario better.

tests/cli/test_compose_sanity.sh
jrusz: Ok I took the `wait_for_compose` function and reused it for detecting the running compose. I am just not sure about the sleep duration and sometimes it didn't even sleep for me because the compose was already running and that then changes the cancel behavior. We'll need Brian's input on this.

atodorov: The above looks good in terms of how we wait for the compose to start running IMO.

tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,27 @@ rlJournalStart
...
 
     rlPhaseStartTest "compose start"
+        UUID=`$CLI compose start example-http-server ami`
+        rlAssertEquals "exit code should be zero" $? 0
+        UUID=`echo $UUID | cut -f 2 -d' '`
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose info $UUID | grep 'RUNNING'; do
+                sleep 20
+                rlLogInfo "Waiting for compose to start ..."
```
atodorov: `start ...` is `running`

tests/cli/test_compose_sanity.sh
atodorov: Cancelling right after in many cases will happen before the compose has even started (e.g. it is still WAITING). The original bug report is about cancel hanging when there is a running compose so this is what we want to cover.  

If I understand the script correctly the current implementation will wait until the status becomes RUNNING  and then cancel it, which will take some time to kill anaconda and cleanup files from disk. This is what you want.

The fact that return code is not 0 is a separate issue.

tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,27 @@ rlJournalStart
...
+            done;
+        else
+            rlFail "Compose UUID is empty!"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "cancel compose"
+        rlRun -t -c "$CLI compose cancel $UUID"
+        `$CLI compose status | grep $UUID`
+        rlAssertNotEquals "compose is canceled" $? 0
```
atodorov: if you grep for something which doesn't exist the exist code of the entire pipeline is 1, b/c grep didn't find it.

tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,27 @@ rlJournalStart
...
+                rlLogInfo "Waiting for compose to start ..."
+            done;
+        else
+            rlFail "Compose UUID is empty!"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "cancel compose"
+        rlRun -t -c "$CLI compose cancel $UUID"
+        `$CLI compose status | grep $UUID`
```
atodorov: backticks not needed

jrusz:
It still needs a change. The probably best way to check if the cancel was successful is to run `composer-cli compose info <UUID>` and that should not find the compose. I just have to add a check that the command failed, which in this case means it's ok.

I came to a conclusion though that this whole test cancel scenario requires it's own test. If you look at this bz https://bugzilla.redhat.com/show_bug.cgi?id=1788501 this would not have been found by the test as it is now. The test should go like this:
1. Start a compose
1. Wait for ~20-30 seconds for it to start anaconda
1. Cancel the compose
1. Start another compose
1. Let it finish and check that it was successful

Right now the compose is being started with `--test=2` which does not start anaconda and would not cover the issue in the bz above.

atodorov:
> I came to a conclusion though that this whole test cancel scenario requires it's own test. If you look at this bz https://bugzilla.redhat.com/show_bug.cgi?id=1788501 this would not have been found by the test as it is now. The test should go like this:

>    Start a compose
    Wait for ~20-30 seconds for it to start anaconda
    Cancel the compose
    Start another compose
    Let it finish and check that it was successful

> Right now the compose is being started with --test=2 which does not start anaconda and would not cover the issue in the bz above.

Good catch but this is an entirely different test scenario. It is related but different. The current one we are concerned about is that cancelling the compose doesn't hang the host system.  

After this PR is merged & backported the changes needed to cover rhzb#1788501 can be blended in very smoothly into the existing script. Feel free to open tickets for yourself as to not forget about this other scenario.

jrusz:
This should be the final version for this test now.

bcl:
Yeah, the coverage/coveralls thing is a different issue. I'm just waiting for coverage 5.0.3 to be released before we can change the requirements back.

PR #935, opened 2020-01-07
composer-cli: Increase DELETE timeout to 120s

bcl:
When the timeout is too short the http library sends the request again.
We return the last response to the user so even if the cancel works they
get an error about the UUID not being valid.

Resolves: rhbz#1788461


bcl:
pushed to master. failures unrelated to the change.

PR #938, opened 2020-01-09
Add test for canceling a running compose

jrusz:
This is a modification of an existing test for compose sanity.
Cherry-picked from master.

### DIFF
--- a/tests/cli/test_compose_sanity.sh
+++ b/tests/cli/test_compose_sanity.sh
@@ -22,6 +22,30 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "compose start"
+        UUID=`$CLI compose start example-http-server ami`
+        rlAssertEquals "exit code should be zero" $? 0
+        UUID=`echo $UUID | cut -f 2 -d' '`
+
+        if [ -n "$UUID" ]; then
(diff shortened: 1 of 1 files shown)


jrusz:
Oh, didn't know that, should be ok now.

atodorov:
1656691 is CLOSED, you need another one which is approved for the current release before we can merge.

jrusz:
I got another related rhbz#1788461. This one should be ok.

tests/cli/test_compose_sanity.sh
```diff
@@ -22,6 +22,26 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "compose start"
+        UUID=`$CLI compose start example-http-server ami`
+        rlAssertEquals "exit code should be zero" $? 0
+        UUID=`echo $UUID | cut -f 2 -d' '`
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose info $UUID | grep 'RUNNING'; do
```
bcl: There is some possibility that this could return FAILED and then it would be stuck in the loop.

jrusz: Thanks for noticing this. I added a check for FAILED compose. I guess this should go to master branch too.

PR #939, opened 2020-01-09
[7.9] Add test for canceling a running compose

jrusz:
This is a modification of an existing test for compose sanity.
Cherry picked from 8c9f528cd42fd8dd61e64572eda89e6168c409a0

Related: rhbz#1789451

### DIFF
--- a/tests/cli/test_compose_sanity.sh
+++ b/tests/cli/test_compose_sanity.sh
@@ -16,6 +16,30 @@ rlJournalStart
     rlPhaseEnd
 
     rlPhaseStartTest "compose start"
+        UUID=`$CLI compose start example-http-server ami`
+        rlAssertEquals "exit code should be zero" $? 0
+        UUID=`echo $UUID | cut -f 2 -d' '`
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose details $UUID | grep 'RUNNING'; do
+                sleep 20
+                rlLogInfo "Waiting for compose to start running..."
+                if $CLI compose info $UUID | grep 'FAILED'; then
+                    rlFail "Compose FAILED!"
+                    break
+                fi
+            done;
+        else
+            rlFail "Compose UUID is empty!"
+        fi
+    rlPhaseEnd
+
+    rlPhaseStartTest "cancel compose"
+        rlRun -t -c "$CLI compose cancel $UUID"
+        rlRun -t -c "$CLI compose details $UUID" 1 "compose is canceled"
+    rlPhaseEnd
+
+    rlPhaseStartTest "compose start again"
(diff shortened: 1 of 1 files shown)


atodorov:
@jrusz what is the justification for closing this PR ?

jrusz:
@atodorov Omg, I'm sorry I was deleting old branches and this got mixed in there, can I restore this?

jrusz:
I added a new commit that is a backport of https://github.com/weldr/lorax/pull/950

tests/cli/test_compose_sanity.sh
```diff
@@ -16,7 +16,34 @@ rlJournalStart
...
     rlPhaseStartTest "compose start"
-        UUID=`$CLI --test=2 compose start example-http-server tar`
+        UUID=`$CLI compose start example-http-server ami`
+        rlAssertEquals "exit code should be zero" $? 0
+        UUID=`echo $UUID | cut -f 2 -d' '`
+
+        if [ -n "$UUID" ]; then
+            until $CLI compose details $UUID | grep 'RUNNING'; do
+                sleep 20
+                rlLogInfo "Waiting for compose to start running..."
```
atodorov: Missing the check for failed compose.

jrusz: Right, thanks for noticing. I added it and squashed with the first commit.

jrusz:
The test fails on rhel7 because `anaconda.pid` is left in `/var/run/` but the next compose can be started and finished because anaconda does not care that it has its `.pid` file there already... I could either change it to just check if it's stopped with `ps` command, but that would make this test a bit different from the other branches or just remove the check for the `.pid` file and only test the after a successful cancel the next compose can be finished successfully... I don't know when or if anaconda is going to change this behavior on rhel7, but until then we need to change this test... What do you think @atodorov ?

tests/cli/test_compose_sanity.sh
```diff
@@ -16,7 +16,38 @@ rlJournalStart
...
+        until ps -axo comm,pid | grep '^anaconda'; do
+            sleep 10
+            rlLogInfo "Waiting for anaconda to start running..."
+        done;
+    rlPhaseEnd
+
+    rlPhaseStartTest "cancel compose"
+        rlRun -t -c "$CLI compose cancel $UUID"
+        rlRun -t -c "$CLI compose details $UUID" 1 "compose is canceled"
+        rlAssertNotExists "/var/run/anaconda.pid"
```
atodorov: IMO remove this line and don't change anything else.

jrusz: I agree, this line is not really necessary for covering the intended test scenario.

PR #941, opened 2020-01-09
lorax-composer: Enable ami on aarch64

bcl:
Resolves: rhbz#1789308

### DIFF
--- a/tests/cli/test_compose_sanity.sh
+++ b/tests/cli/test_compose_sanity.sh
@@ -14,6 +14,10 @@ rlJournalStart
         if [ "$(uname -m)" = "x86_64" ]; then
             rlAssertEquals "lists all supported types" \
                     "`$CLI compose types | xargs`" "alibaba ami ext4-filesystem google live-iso openstack partitioned-disk qcow2 tar vhd vmdk"
+        elif [ "$(uname -m)" == "aarch64" ]; then
+            # ami is supported on aarch64
+            rlAssertEquals "lists all supported types" \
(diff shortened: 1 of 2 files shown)


tests/cli/test_compose_sanity.sh
```diff
@@ -14,6 +14,11 @@ rlJournalStart
         if [ "$(uname -m)" = "x86_64" ]; then
             rlAssertEquals "lists all supported types" \
                     "`$CLI compose types | xargs`" "alibaba ami ext4-filesystem google live-iso openstack partitioned-disk qcow2 tar vhd vmdk"
+        elif [ "$(uname -m)" == "aarch64" ]; then
+            # ami is supported on aarch64
+            rlAssertEquals "lists all supported types" \
+                    "`$CLI compose types | xargs`" "ami ext4-filesystem live-iso liveimg-tar openstack partitioned-disk qcow2 tar"
+>>>>>>> 77eec0d0... lorax-composer: Enable ami on aarch64
```
atodorov: leftover from conflict resolution.

atodorov: I like better the approach in f0b60f4c3fcc8c9319e79a2de6b787e59e4c65d7 instead of listing all compose types separately for each architecture. However the linked commit doesn't exactly do what I am advocating for so either way is fine for now.

bcl: Thanks for catching that.

bcl: > I like better the approach in [f0b60f4](https://github.com/weldr/lorax/commit/f0b60f4c3fcc8c9319e79a2de6b787e59e4c65d7) instead

That seems harder to read. I have to assemble the string in my head in order to understand what's going on.

PR #948, opened 2020-01-14
tests: Check for cloud-init presence in azure image

jrusz:
This tests if the cloud-init is installed and running.


jrusz:
Yeah, sorry, copy&paste error. Should be ok now.

PR #955, opened 2020-01-21
lorax: Catch rootfs out of space failures

bcl:
It isn't always obvious what happened when the rootfs runs out of space,
especially when using lorax via pungi. So this checks for the out of
space error string when building the runtime image and logs it to the
primary logfile and console as an error with the rootfs size.

eg.
2020-01-20 18:52:58,920: The rootfs ran out of space with size=1

### DIFF
--- a/src/pylorax/treebuilder.py
+++ b/src/pylorax/treebuilder.py
@@ -23,6 +23,7 @@ logger = logging.getLogger("pylorax.treebuilder")
 import os, re
 from os.path import basename
 from shutil import copytree, copy2
+from subprocess import CalledProcessError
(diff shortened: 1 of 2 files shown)


AdamWill:
The test looks wrong or incomplete - it doesn't seem to check that the correct message is logged, AFAICS.

bcl:
> The test looks wrong or incomplete - it doesn't seem to check that the correct message is logged, AFAICS.

Testing the guts of treebuilder are for another time, it's something I'd like to add eventually. This test just makes sure that writing too much data to a small rootfs raises the right error.

PR #956, opened 2020-01-22
tests: run ssh commands in batch mode

jrusz:
This should ensure that during manual testing there won't be any password prompts.

### DIFF
--- a/tests/cli/lib/lib.sh
+++ b/tests/cli/lib/lib.sh
@@ -132,7 +132,7 @@ composer_stop() {
 verify_image() {
     SSH_USER="$1"
     SSH_MACHINE="$2"
-    SSH_OPTS="-o StrictHostKeyChecking=no $3"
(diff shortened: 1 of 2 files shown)


tests/cli/lib/lib.sh
```diff
@@ -132,7 +132,7 @@ composer_stop() {
 verify_image() {
     SSH_USER="$1"
     SSH_MACHINE="$2"
-    SSH_OPTS="-o StrictHostKeyChecking=no $3"
+    SSH_OPTS="-o StrictHostKeyChecking=no -o BatchMode=yes $3"
```
atodorov: Also add to `tests/cli/test_build_and_deploy_azure.sh` for cloud-init checks.

jrusz: Added. Does it actually do anything if the identity file is specified as well?

atodorov: I think the only difference is when you run the commands on interactive terminal and for commands which are supposed not to be able to authenticate.

PR #909, opened 2019-11-21
[7.9]: Unskip AWS test scenario

atodorov:
Related: rhbz#1670032

PR #957, opened 2020-01-22
Add workaround for AWS deployment test

jikortus:
This is a workaround for the 'RequestTimeTooSkewed' issue we sometimes get with AWS uploads. It's only meant as a temporary solution until the related BZ has been fixed. I tested it on a machine where I was hitting that issue and it seems to perform as expected.

I also wondered if I should modify the code to fail on RHEL-7.9+ when the issue is hit to make the test hit our attention again after RHEL-7.8 (otherwise the test would pass despite the presence of the time skew, and would only log the issue if present).

### DIFF
--- a/test/check-cloud
+++ b/test/check-cloud
@@ -1,12 +1,9 @@
 #!/usr/bin/python3
 
-import unittest
-
 import composertest
 
 
 class TestCloud(composertest.ComposerTestCase):
-    @unittest.skip('Disabled b/c of RequestTimeTooSkewed issue')
     def test_aws(self):
         self.runCliTest("/tests/cli/test_build_and_deploy_aws.sh")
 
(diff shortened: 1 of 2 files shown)


jikortus:
It looks like the workaround doesn't necessarily work as expected and just resyncing the time via NTP may not bring the desired result (although this worked for me when I tested in the first time outside of CI). It seems I'll also need to have a look about what happens in terms of localtime/utc and timezone settings and make that part of the workaround.

atodorov:
> It looks like the workaround doesn't necessarily work as expected 

Given that information and the fact that we can't merge the workaround to enable testing in CI before 7.9 has been approved I am closing this pull request.

There is #909 which will be available for merge at the same time where the anaconda bug will be fixed. 

This issue has been primarily problem within CI, not when executing the test suite by hand so I don't see a reason to spend more resources on it. Once the anaconda bug has been resolved we will enable the AWS test scenario in CI and everything should be fine.

PR #959, opened 2020-01-28
Test & cleanup script for Alibaba cloud

atodorov:
Cherry-picked from 718ac31c47a8f4f6779e2773811771067709beb7

Related: rhbz#1785154

### DIFF
--- a/tests/cleanup/remove_old_objects_alibaba.sh
+++ b/tests/cleanup/remove_old_objects_alibaba.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+# Script removes ECS Instances, Custom Images and OSS files older than
+# HOURS_LIMIT (24 hours by default) from Alibaba cloud
+#
+. /usr/share/beakerlib/beakerlib.sh
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        ALI_DIR=`mktemp -d /tmp/alicloud.XXXXX`
+
+        if [ -z "$ALICLOUD_ACCESS_KEY" ]; then
+            rlFail "ALICLOUD_ACCESS_KEY is empty!"
+        else
+            rlLogInfo "ALICLOUD_ACCESS_KEY is configured"
+        fi
+
+        if [ -z "$ALICLOUD_SECRET_KEY" ]; then
+            rlFail "ALICLOUD_SECRET_KEY is empty!"
+        else
+            rlLogInfo "ALICLOUD_SECRET_KEY is configured"
+        fi
+
+        ALICLOUD_BUCKET="${ALICLOUD_BUCKET:-composer-test}"
+        rlLogInfo "ALICLOUD_BUCKET=$ALICLOUD_BUCKET"
+
+        # VMs older than HOURS_LIMIT will be deleted
+        HOURS_LIMIT="${HOURS_LIMIT:-24}"
+        TIMESTAMP=`date -u -d "$HOURS_LIMIT hours ago" '+%FT%T'`
+        rlLogInfo "HOURS_LIMIT=$HOURS_LIMIT"
+        rlLogInfo "TIMESTAMP=$TIMESTAMP"
+
+        for package in jq; do
+            if ! rlCheckRpm "$package"; then
+                rlRun -t -c "dnf -y install $package"
+                rlAssertRpm "$package"
+            fi
+        done
+
+        # use the CLI b/c Ansible modules are not yet upstream and are unreliable
+        TAR_FILE="aliyun-cli-linux-3.0.32-amd64.tgz"
+        curl -L https://github.com/aliyun/aliyun-cli/releases/download/v3.0.32/$TAR_FILE > $ALI_DIR/$TAR_FILE
+        tar -C $ALI_DIR/ -xzvf $ALI_DIR/$TAR_FILE
+        chmod a+x $ALI_DIR/aliyun
+
+        # configure
+        [ -d ~/.aliyun/ ] || mkdir ~/.aliyun/
+
+        if [ -f ~/.aliyun/config.json ]; then
+            rlLogInfo "Reusing existing ~/.aliyun/config.json"
+        else
+            rlLogInfo "Creating ~/.aliyun/config.json"
+            cat > ~/.aliyun/config.json << __EOF__
+{
+    "current": "",
+    "profiles": [
+        {
+            "mode": "AK",
+            "access_key_id": "$ALICLOUD_ACCESS_KEY",
+            "access_key_secret": "$ALICLOUD_SECRET_KEY",
+            "region_id": "$ALICLOUD_REGION",
+            "output_format": "json",
+            "language": "en"
+        }
+    ],
+    "meta_path": ""
+}
+__EOF__
+        fi
+
+    rlPhaseEnd
+
+# Check all regions
+for REGION_ID in `$ALI_DIR/aliyun Ecs DescribeRegions | jq -r '.Regions.Region[] | .RegionId'`; do
+    rlPhaseStartTest "Delete old VMs in region $REGION_ID"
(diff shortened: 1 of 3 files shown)


jrusz:
Maybe this test should be called `test_build_and_deploy_alibaba.sh` ? Just like the aws, azure etc...

atodorov:
> Maybe this test should be called test_build_and_deploy_alibaba.sh ? Just like the aws, azure etc...

I wanted to change the naming scheme

tests/cli/test_compose_alibaba.sh
```diff
@@ -0,0 +1,197 @@
...
+        rlAssertEquals "Instance $INSTANCE_ID is Running" \
+            "$($ALI_DIR/aliyun ecs DescribeInstanceAttribute --InstanceId $INSTANCE_ID | jq -r .Status)" "Running"
+        rlRun -t -c "$ALI_DIR/aliyun ecs DescribeInstanceAttribute --InstanceId $INSTANCE_ID"
+
+        IP_ADDRESS="null"
+        while [ "$IP_ADDRESS" == "null" ]; do
+            rlLogInfo "IP_ADDRESS is not assigned yet ..."
+            sleep 30
+            IP_ADDRESS=`$ALI_DIR/aliyun ecs DescribeInstanceAttribute --InstanceId $INSTANCE_ID | jq -r .PublicIpAddress.IpAddress | jq -r '.[0]'`
+        done
```
jrusz: I can see the `IP_ADDRESS` already assigned when line 161 is being run. Maybe it can be retrieved from there and check if it's `null` instead of setting it to `null`,  waiting and then retrieving it?

tests/cli/test_compose_alibaba.sh
```diff
@@ -0,0 +1,197 @@
...
+    rlPhaseStartCleanup
+        rlRun -t -c "$ALI_DIR/aliyun ecs DeleteInstance --Force True --InstanceId $INSTANCE_ID"
+        rlRun -t -c "$ALI_DIR/aliyun ecs DeleteImage --Force True --ImageId $IMAGE_ID"
+        rlRun -t -c "$ALI_DIR/aliyun oss rm oss://$ALICLOUD_BUCKET/$IMAGE --force"
+        rlRun -t -c "$CLI compose delete $UUID"
+        rlRun -t -c "rm -rf $IMAGE $SSH_KEY_DIR $ALI_DIR"
+        # do this here to give time for the VM instance to be removed properly
+        # also don't fail if the key is still attached to an instance which is waiting
+        # to be desroyed. We're going to remove these keys in cleanup afterwards
+        $ALI_DIR/aliyun ecs DeleteKeyPairs --KeyPairNames "['$KEY_NAME']" || echo
```
jrusz: I don't understand what line 193 does. The `$ALI_DIR` is removed on line 189, so can this ever do something?

tests/cli/test_compose_alibaba.sh
atodorov: line 161 is like an info dump after we're certain the VM instance is in status running so we've got a reference to all of its attributes.

Then the script proceeds to get the IP address. I can instead assign the output of line 161 to a variable, rlLogInfo that variable and then echo & grep for the IP address but it is not any different.  So I'm not inclided to do that unless you insist too much.

tests/cli/test_compose_alibaba.sh
atodorov: This deletes keypairs from the Alicloud console, not from the local filesystem.

tests/cli/test_compose_alibaba.sh
jrusz: Ok good, I understand line 161, but what I did not understand is why not do line 163 the same way like line 167, why assing it `null` first? But yes the results would be the same, so you don't have to change it I was just curious.

tests/cli/test_compose_alibaba.sh
jrusz: Ok... I don't know how the Alicloud console works, I just saw this `/tests/cli/test_compose_alibaba.sh: line 193: /tmp/alicloud.c6k0i/aliyun: No such file or directory` in the log and thought that it does nothing.

atodorov: Crap, that is a bug. Good catch. This line should be move 1 line up.

tests/cli/test_compose_alibaba.sh
```diff
@@ -0,0 +1,197 @@
...
+                rlRun -t -c "dnf -y install $package"
+                rlAssertRpm "$package"
+            fi
+        done
+
+        ALI_DIR=`mktemp -d /tmp/alicloud.XXXXX`
+        # use the CLI b/c Ansible modules are not yet upstream and are unreliable
+        TAR_FILE="aliyun-cli-linux-3.0.32-amd64.tgz"
+        curl -L https://github.com/aliyun/aliyun-cli/releases/download/v3.0.32/$TAR_FILE > $ALI_DIR/$TAR_FILE
+        tar -C $ALI_DIR/ -xzvf $ALI_DIR/$TAR_FILE
```
jikortus: I'd suggest to download and use the latest version here as well (as in `remove_old_objects_alibaba.sh`).

tests/cli/test_compose_alibaba.sh
```diff
@@ -0,0 +1,197 @@
...
+
+        rlLogInfo "ALICLOUD_BUCKET=$ALICLOUD_BUCKET"
+        rlLogInfo "ALICLOUD_REGION=$ALICLOUD_REGION"
+
+        for package in jq; do
+            if ! rlCheckRpm "$package"; then
+                rlRun -t -c "dnf -y install $package"
+                rlAssertRpm "$package"
+            fi
+        done
```
jikortus: Is the loop necessary here (same question as for `remove_old_objects_alibaba.sh`).

tests/cleanup/remove_old_objects_alibaba.sh
```diff
@@ -0,0 +1,124 @@
...
+        ALICLOUD_BUCKET="${ALICLOUD_BUCKET:-composer-test}"
+        rlLogInfo "ALICLOUD_BUCKET=$ALICLOUD_BUCKET"
+
+        # VMs older than HOURS_LIMIT will be deleted
+        HOURS_LIMIT="${HOURS_LIMIT:-24}"
+        TIMESTAMP=`date -u -d "$HOURS_LIMIT hours ago" '+%FT%T'`
+        rlLogInfo "HOURS_LIMIT=$HOURS_LIMIT"
+        rlLogInfo "TIMESTAMP=$TIMESTAMP"
+
+        for package in jq; do
```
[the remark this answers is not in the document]
atodorov: it's not strictly necessary, that is a left over from how I've copied the script initially. When we're able to migrate to ansible we'll need to install some packages here but not for the time being.

tests/cleanup/remove_old_objects_alibaba.sh
```diff
@@ -0,0 +1,124 @@
...
+            if ! rlCheckRpm "$package"; then
+                rlRun -t -c "dnf -y install $package"
+                rlAssertRpm "$package"
+            fi
+        done
+
+        # use the CLI b/c Ansible modules are not yet upstream and are unreliable
+        TAR_FILE="aliyun-cli-linux-3.0.32-amd64.tgz"
+        curl -L https://github.com/aliyun/aliyun-cli/releases/download/v3.0.32/$TAR_FILE > $ALI_DIR/$TAR_FILE
```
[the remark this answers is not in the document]
atodorov: :-1: on that idea. 

- using the latest version is prone to breaking the script when the cli decides to implement some of these commands differently. Not very likely but still possible
- the tar file names are not consistent between versions. There are slight differences in suffix of these files between older versions (since when this PR was originally opened) and the current ones. Seeing this history of release changes makes me pretty confident that would change again sometime in the future. 
- using GitHub API is overkill IMO. The right course of action is getting the ansible modules supported upstream so we don't have to rely on the cli. 

With a version lock the only thing that could go wrong is someone deleting the tarball from GitHub. We can always host this somewhere else (e.g. in our own S3) however that doesn't guarantee that the files won't become missing over time. We've seen breaking changes in our infra in the past (incl changing access scope).

PR #960, opened 2020-01-28
Add new test to run 'test' and 'check' in cockpit ci

jrusz:
This is a work in progress with some comments in code and possible placeholder names. I will create a manual trigger in `testmap.py` and see how this works so far.

### DIFF
--- a/tests/cli/test_ci.sh
+++ b/tests/cli/test_ci.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+# Note: execute this file from the project root directory
+
+set -e
+
+. /usr/share/beakerlib/beakerlib.sh
+
+yum-config-manager --enable epel 
+
+dnf -y install $(cat /lorax/test-packages)
+
+pip3 install pocketlint 
+
+cd /lorax
+
+rlJournalStart
+    rlPhaseStartTest "check"
+        # running with -k because of possible errors that would make "make test" to not run
+    	rlRun "make ci -k"
+    rlPhaseEnd
+
+rlJournalEnd
+rlJournalPrintText
--- a/test/check-cli
+++ b/test/check-cli
@@ -103,6 +103,18 @@ class TestTar(composertest.ComposerTestCase):
             # Run the test, on the booted image
             self.runImageTest("/tests/cli/test_boot_tar_kickstart.sh")
 
+class TestPylint(composertest.ComposerTestCase):
+    def test_ci(self):
+
+        # Create an empty folder for the upload bellow
+        self.execute(["mkdir", "/lorax"])
+
+        # Upload contents of repo for running the test
+        self.machine.upload(["../tests", "../src", "../share", "../docs", "../Makefile",
+        "../test-packages"], "/lorax")
+
+        # Running the test
+        self.runImageTest("/tests/cli/test_ci.sh")
 
 if __name__ == '__main__':
     composertest.main()
--- a/test/run
+++ b/test/run
@@ -11,6 +11,8 @@ if [ -n "$TEST_SCENARIO" ]; then
     test/check-cli TestQcow2
   elif [ "$TEST_SCENARIO" == "tar" ]; then
     test/check-cli TestTar
+  elif [ "$TEST_SCENARIO" == "ci" ]; then
+	test/check-cli TestPylint 
   else
     test/check-cloud TestCloud.test_$TEST_SCENARIO
   fi


jrusz:
@bcl could you please take a look at the test? I can see pylint is failing with some error and I don't know what seems to be the problem.

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,75 @@
...
+
+. /usr/share/beakerlib/beakerlib.sh
+
+# I guess that all of this could be moved to setup phase
+
+cat >> /etc/yum.repos.d/nightly.repo << __EOF__
+
+[RHEL-8-NIGHTLY-CRB]
+name=CRB
+baseurl=http://download-ipv4.eng.brq.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.2/compose/CRB/x86_64/os/
```
bcl: Seems like a fragile URL, and possibly not local to wherever the test runs. Is there a more general one that can be used?

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,75 @@
...
+enabled=1
+gpgcheck=0
+
+__EOF__
+
+yum-config-manager --enable epel epel-playground
+
+dnf -y install \
```
bcl: This is exactly like the Dockerfile setup, it's probably time to move it to someplace common so that we are not keeping multiple file lists in sync. See PR #961

jrusz: Yes, that would be nice. I'll wait for a PR on rhel8-branch to be merged and rebase this PR then.

tests/cli/test_ci.sh
jrusz: Thanks for noticing, I guess I got redirected when I was looking for it. I changed it to the one you proposed.

tests/composer/test_blueprints.py
```diff
@@ -233,7 +233,7 @@ def test_freeze(self):
     @unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Test requires a running API server")
     def test_tag(self):
         """blueprints tag"""
-        rc = blueprints_tag("/run/weldr/api.socket", 0, ["glusterfs"], show_json=False)
+        rc = blueprints_tag("/run/weldr/api.socket", 0, ["example-http-server"], show_json=False)
```
atodorov: I see where this commit came from but I don't understand the reason behind it. Can't we use example-glusterfs.toml instead ?

tests/composer/test_blueprints.py
```diff
@@ -173,20 +173,20 @@ def test_changes(self):
...
-        """blueprints save example-development"""
+        """blueprints save"""
         blueprints_save("/run/weldr/api.socket", 0, ["example-http-server"], show_json=False)
         self.assertTrue(os.path.exists("example-http-server.toml"))
 
     @unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Test requires a running API server")
     def test_save_1(self):
-        """blueprints delete example-development"""
+        """blueprints delete"""
         rc = blueprints_delete("/run/weldr/api.socket", 0, ["example-development"], show_json=False)
```
atodorov: Don't you have to delete `example-http-server` which is now referenced above and below ? 

@bcl I don't quite understand the behavior of these 3 test methods and why their order matters. Looks fragile to me. Can you explain what is the intended behavior here ?

tests/composer/test_blueprints.py
jrusz: Well, the way I undertand it, the example-glusterfs doesn't exist anymore? I think it's a thing from the past.. These tests were skipped in travis all the time so I guess that's why this was not updated earlier..

tests/composer/test_blueprints.py
jrusz: If I delete `example-http-server` then the last test called `blueprints-diff` fails because it won't find this blueprint. But I guess Brian would know better what's exactly happening here I was also confused when trying to figure this out.

jrusz:
@bcl @atodorov Are we going to stop using travis after this is merged? I personally don't see a point in running it if this works and it also test more because there were some tests skipped in travis which are now running in cockpit CI. The reason I'm asking now is because in the `Makefile` under `make test` is also `coverage3` command which on rhel8 is `coverage-3` for some reason... I can either create an alias or symlink in the test so it can run on rhel8 or just change it in the `Makefile` but that breaks travis.

tests/cli/test_ci.sh
```diff
@@ -0,0 +1,75 @@
...
+
+. /usr/share/beakerlib/beakerlib.sh
+
+# I guess that all of this could be moved to setup phase
+
+cat >> /etc/yum.repos.d/nightly.repo << __EOF__
+
+[RHEL-8-NIGHTLY-CRB]
+name=CRB
+baseurl=http://download.devel.redhat.com/nightly/latest-RHEL-8/compose/CRB/x86_64/os/
```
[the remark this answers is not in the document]
jrusz: It should be CodeReady Linux Builder repo that was introduced in rhel8. It is also available when you subscribe your rhel8 system. The reason I'm adding it here is that it contains `python3-sphinx` for example and that way I don't have to use `pip` to install it. 

When making this test I aimed to install as few packages as possible using `pip`.

tests/composer/test_blueprints.py
bcl: Yes, these are fragile because they are running against an actual API server so the changes persist and may effect later test results, that's why I used different blueprints for some of them, so that the changes wouldn't cause problems with example-http-server which is used for most of them.

If we need to add another test blueprint to replace glusterfs or whatever that's fine with me but the changes here should be as minimal as possible.

bcl:
> @bcl @atodorov Are we going to stop using travis after this is merged?

I'd like to keep using travis, it usually runs faster than the cockpit-ci based setup. I also run the tests locally using test-in-docker so that target needs to keep working.

tests/cli/test_ci.sh
atodorov: #961 has been merged. rebase and use the common list of packages to be installed.

tests/composer/test_blueprints.py
atodorov: @jrusz from what I can tell by Brian's comment you shouldn't have to make changes here. As of now I'm not sure we know this piece of test code well enough to be confident that these changes result in the test methods exercising the same functionality that was tested before the change. 

If it did work in Travis then you shouldn't have to change anything here or am I missing something ?

tests/composer/test_blueprints.py
atodorov: I do see `./tests/pylorax/blueprints/example-glusterfs.toml` which is what gets used by the test suite I believe so it definitely exists. Again, not sure if the name used in the code didn't exist how this test method even passed in Travis.

tests/composer/test_blueprints.py
jrusz: Well that's because these tests never ran in Travis, they were just skipped. I first encountered this when trying to run `make ci` here on master branch. I fixed those tests but it never made it to rhel8-branch because there were missing packages in epel8 and we didn't want to use pip. Now that we know we have to use pip regardless I decided to go back to this and make it work.

tests/composer/test_blueprints.py
jrusz: I can try `example-glusterfs.toml` but I guess it would have to be uploaded first, I'm not sure now. Again this was skipped in travis so that's why this was not an issue until now.

tests/composer/test_blueprints.py
atodorov: ok, I see what's happening now. In Travis there was no running server that's why these test methods were skipped. 

Regardless of that, some of the blueprints should still be available. I see `example-development.toml` and `example-glusterfs.toml` under `./tests/pylorax/blueprints`. Is this the directory used by the server API while these tests are running ? IOW are we starting the API server from the git checkout or relying on the system wide installation to be running via systemd ? This is what I still can't understand.

jrusz: > IOW are we starting the API server from the git checkout or relying on the system wide installation to be running via systemd ? This is what I still can't understand.

After some investigation I believe that this depends on whether or not is the `$CLI` set. 
The following is from `tests/test_cli.sh`
```
 82 if [ -z "$CLI" ]; then
 83     export top_srcdir=`pwd`
 84     . ./tests/testenv.sh
 85 
 86     export BLUEPRINTS_DIR=`mktemp -d '/tmp/composer-blueprints.XXXXX'`
 87     cp ./tests/pylorax/blueprints/*.toml $BLUEPRINTS_DIR
 88 
 89     export SHARE_DIR=`mktemp -d '/tmp/composer-share.XXXXX'`
 90     cp -R ./share/* $SHARE_DIR
 91     chmod a+rx -R $SHARE_DIR
 92 
 93     setup_tests $SHARE_DIR $BLUEPRINTS_DIR
 94     # start the lorax-composer daemon
 95     composer_start
 96 else
 97     export PACKAGE="composer-cli"
 98     export BLUEPRINTS_DIR="/var/lib/lorax/composer/blueprints"
 99     composer_stop
100     setup_tests /usr/share/lorax /var/lib/lorax/composer/blueprints
101     composer_start
102 fi

```
So since I don't set `$CLI` in the test I would say this runs with the git checkout.

tests/composer/test_blueprints.py
jrusz: I changed it to `example-glusterfs` and it passed. So I guess it depends if there's any difference in which blueprints is used, test is passing as long as the blueprints exists it seems.

atodorov: ok, in this case the change is fine.

tests/composer/test_blueprints.py
atodorov: So in this case we're starting the lorax server from the git checkout and instruct it to serve the blueprints which are available in the source code. There definitely is an `example-development.toml` file in there so all of these changes shouldn't be necessary. 

From what I can tell the sequence of steps is: upload a blueprint and make sure it exists; then delete it and make sure it doesn't exist and finally upload it again, maybe as a means to restore the environment like it was before.

tests/cli/test_ci.sh
jrusz: I just realized that the list of packages is actually different for rhel-8 and for fedora-28 meaning we can't simply have one file list unless we either drop the testing on fedora-28 or make a conditional list or something...

tests/composer/test_blueprints.py
jrusz: Right that seems to be what's happening here, but if I leave all this with `example-development` another test fails and I get this:
```
======================================================================                                                                                                                                                                                                          
FAIL: blueprints diff                                                                                                                                                                                                                                                           
----------------------------------------------------------------------                                                                                                                                                                                                          
Traceback (most recent call last):                                                                                                                                                                                                                                              
  File "/lorax/tests/composer/test_blueprints.py", line 274, in test_z_diff                                                                                                                                                                                                     
    self.assertTrue("Changed Version" in output)                                                                                        
AssertionError: False is not true                                                                                                                                                                                                                                               
-------------------- >> begin captured logging << --------------------                                                                                                                                                                                                          
urllib3.connectionpool: DEBUG: http://localhost:None "GET /api/v0/blueprints/changes/example-http-server?limit=0 HTTP/1.1" 200 163                                                                                                                                              
urllib3.connectionpool: DEBUG: http://localhost:None "GET /api/v0/blueprints/changes/example-http-server?limit=1 HTTP/1.1" 200 401                                                                                                                                              
urllib3.connectionpool: DEBUG: http://localhost:None "GET /api/v0/blueprints/diff/example-http-server/14c14d18d3713c909e68e03026ba5a0869528a5b/HEAD HTTP/1.1" 200 17                                                                                                            
--------------------- >> end captured logging << ---------------------                                                                                                   
```
But I can make it pass by making changes in that particular test and the one above it. So this seems to be a question of what we want to change, I can change either and this will pass... 

@bcl could you please decide which do you prefer? I can see these options:

1. Leave it as it is now, meaning have the `test_save_0, test_save_1, test_save_2` use `example-http-server`
1. Have those tests use `example-development` and then change `test_workspace, test_z_diff` to use `example-development` as well

tests/cli/test_ci.sh
bcl: Well crud. And it looks like CentOS8 still doesn't have all the things we need either.
Which also made me realize that the CRB repo is internal only. We need to keep in mind that this code will eventually make its way to CentOS8 and it needs to be able to run in that environment as well.

The original plan was to use Fedora 28 to run things until CentOS8 was available and then to switch over to it.

If we are going to use RHEL repos they need to be passed to the tests somehow, not hard-coded into them.

bcl:
Please take a look at PR #964, I think I've worked the bugs out of using CentOS8 for testing which should help this PR.

atodorov:
@jrusz - let's put this PR on hold for the moment. #964 brings in considerable changes and also modifies some of the test methods that you touch in this PR. IMO once 964 gets merged you will have a more closer environment between Travis CI (CentOS 8) and Cockpit CI (rhel8) hence your own changes should be smaller.

jrusz:
Thanks a lot @bcl for solving all the failing tests and travis dockerfile.
@atodorov I rebased and modified the test.

bcl:
> Thanks a lot @bcl for solving all the failing tests and travis dockerfile.
> @atodorov I rebased and modified the test.

You're welcome. Looks like it made things a lot simpler :)

PR #962, opened 2020-01-29
tests: remove ALI_DIR after we're done using the cli

atodorov:
small fix


bcl:
> Looks good. I wonder what happened to travis, maybe something changed in rawhide that causes the failure?

glib2 had a bug, which should be fixed when the new build hits the repos:

https://bugzilla.redhat.com/show_bug.cgi?id=1795052

PR #964, opened 2020-02-05
Update testing code to use CentOS8 instead of Fedora 28 for testing.

bcl:


bcl:
> @bcl when you run them locally do they get skipped or do these test methods reports PASS (OK)?
> 
> If these methods are skipped even locally that means @jrusz will have to rebase #960 and adjust after this PR gets merged.

Those methods run when there is an active API on /run/weldr/api.socket, I tested it using `test-in-docker` and on a RHEL8 system with a running API and it passes for me.

atodorov:
> Those methods run when there is an active API on /run/weldr/api.socket, I tested it using test-in-docker and on a RHEL8 system with a running API and it passes for me.

:+1: from me then.

PR #967, opened 2020-02-13
Update TestApi skip message because of a bug

atodorov:


bcl:
The first thing I see is that `r = self.request("GET", "/api/v0/status")` should be `/api/status`

atodorov:
The API test starts  ssh forward from localhost:8080 (that is the CI host executing the test scripts) to /run/weldr/api.socket on the VM (which is running lorax-composer). 

@bcl I can see the port available on the CI host but for some reason all connections terminate immediately..

```
# curl -v http://localhost:8080/api/status
*   Trying ::1:8080...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /api/status HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.66.0
> Accept: */*
> 
* Recv failure: Connection reset by peer
* Closing connection 0
curl: (56) Recv failure: Connection reset by peer
```

On the VM host I see 
```
Feb 14 12:25:34 m1.cockpit.lan sshd[1401]: refused streamlocal port forward: originator ::1 port 46504, target /run/weldr/api.socket
```

I tried to add these in `/etc/ssh/sshd_config` on the VM host & restart sshd without success

```
AllowTCPForwarding yes
AllowStreamLocalForwarding yes
PermitOpen any
```

@larskarlitski any ideas what may be blocking us on RHEL 7 ?

bcl:
I don't think I've ever managed to get ssh port forwarding from TCP to a domain socket working, so I don't know if it's even possible in this setup.

atodorov:
@bcl, @larskarlitski  - turns out this is a bug in ssh documented here https://bugzilla.redhat.com/show_bug.cgi?id=1805238

Do you want to workaround by using an ssh package from a scratch build or remove this test scenario altogether ?  It doesn't seem very likely backporting the ssh patch will make it into RHEL 7 and the test is still failing with other issues.

atodorov:
@bcl, @larskarlitski ping. How do you want to proceed here, see previous comment?

Either ask for https://bugzilla.redhat.com/show_bug.cgi?id=1805238 to get fixed (unlikely) or use an ssh package from a scratch build or leave as-is (skipped) or remove ?

larskarlitski:
I’d skip before messing around with scratch builds, but would leave the final decision to @bcl.

bcl:
> I’d skip before messing around with scratch builds, but would leave the final decision to @bcl.

I agree, let's just leave it skipped. Maybe add a comment with a pointer to that bug.

atodorov:
@bcl feel free to merge this despite the failures. None of them are related to the proposed changes.

PR #971, opened 2020-02-21
[WIP, don't merge] Fix tests which break on secondary arches

atodorov:
cc @jrusz

opening here so I anyone wanting to test secondary arch support while lorax-composer is still the default backend can do so.

@bcl no need to merge b/c this test suite will be soon obsolete.


jrusz:
Yeah, that's right, didn't think of that, thanks.

PR #976, opened 2020-03-05
[tests] Enable handling of rel-eng images

jikortus:
In order to enable testing with rel-eng images, we need some slight changes in `Makefile` and `vm.install`. This could actually be done by using the already existing `vm-local-repos` target, however this involves some unnecessary steps (building and installing lorax-composer RPMs that are in turn removed and reinstalled by RPMs from additionally defined custom repos - which itself is another unnecessary step). This approach however can't be used on a host system other than RHEL-7, as one of the steps involved is building a RHEL-7 lorax-composer SRPM.

Thus I think the best way is to define a new Makefile target and introduce a `BUILD_SRPM` environment variable in `vm.install` script that bypasses RPM building and installation in case of a rel-eng image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,20 @@ vm-local-repos: vm
 		--run-command "systemctl enable lorax-composer" \
 		$(TEST_OS)
 
+vm-releng:
+	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
+	bots/image-customize -v \
+		--resize 20G \
+		--upload $(CURDIR)/test/vm.install:/var/tmp/vm.install \
+		--upload $(realpath tests):/ \
+		--run-command "chmod +x /var/tmp/vm.install" \
+		--run-command "cd /var/tmp; BUILD_SRPM=0 /var/tmp/vm.install" \
+		$(TEST_OS)
+	[ -f ~/.config/lorax-test-env ] && bots/image-customize \
+		--upload ~/.config/lorax-test-env:/var/tmp/lorax-test-env \
+		$(TEST_OS) || echo
+
+
 vm-reset:
 	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
 
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,6 @@ endif
 export TEST_OS
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 
-ifeq ($(REPOS_DIR),)
-REPOS_DIR = /etc/yum.repos.d
-endif
-
 default: all
 
 src/composer/version.py:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -149,6 +149,20 @@ vm-local-repos: vm
 		--run-command "systemctl enable lorax-composer" \
 		$(TEST_OS)
 
+vm-releng:
```
atodorov, line 152: That should depend on the `vm` target or no ?

atodorov: If this is the target intended for usage downstream then `vm-local-repos` should be removed, right ?

jikortus: The process of creating & customizing the rel-eng image is a bit different from fedora/nightly images. The first step is to manually create an image using `bots/image-create-rhel-compose` script, together with environment variables determining URLs to the base and extras compose (yes, it can default to -latest, but will likely fail due to the changes in URL structures we've seen lately). The second step is to run `make vm-releng`, which customizes the image in an analogous way to what we've used for the stock bots images.

However, if you think I should change that workflow to be aligned with `vm` target, I'd propose creating a new target (`vm-releng-image`) that would check for presence of the necessary env variables and then would fire up the rel-eng creation script. This target could then be a dependency of `vm-releng` target.

As for removal of `vm-local-repos`, I don't have a strong opinion about it. It purpose is/was a bit different, but on the other hand it will likely find no more use when the `vm-releng` target is introduced, I guess.

atodorov: > The process of creating & customizing the rel-eng image is a bit different from fedora/nightly images. 

Then you need to add a few lines to `test/README` to share the knowledge or link to the bots repo, etc.

> However, if you think I should change that workflow to be aligned with vm target, I'd propose creating a new target (vm-releng-image) 

There is a single-use case here from what I can see. We don't need any more make targets. Any related changes should go into the current make target depending on how you decide to evolve this PR.

> As for removal of vm-local-repos, I don't have a strong opinion about it. It purpose is/was a bit different

It's a very similar make target which I believe is unused ATM. In the case that you are adding a new target either remove the previous one or maybe try merging the implementation so 1 target can do 2 very closely related things.

jikortus: OK, thanks for you insights. I decided to remove the `vm-local-repos` target altogether. I think that this target and the new one do two different things (but to achieve mostly the same goal) and it would be quite cumbersome to merge them together in a seamless manner.

Makefile
```diff
@@ -134,20 +134,19 @@ $(VM_IMAGE): srpm bots
...
-# and update the image. Mostly used when testing downstream snapshots to make
-# sure VM_IMAGE is as close as possible to the host!
-vm-local-repos: vm
-	bots/image-customize -v \
-		--run-command "rm -rf /etc/yum.repos.d" \
-		$(TEST_OS)
+vm-releng:
+	rm -f $(VM_IMAGE) $(VM_IMAGE).qcow2
 	bots/image-customize -v \
-		--upload $(REPOS_DIR):/etc/yum.repos.d \
```
atodorov, line 145: Dangling section for REPOS_DIR remaining at the top of Makefile.

atodorov:
@bcl rhel-7-9 and rhel-7-9/openstack failed previously but the failure isn't related to the current changes. I've rescheduled them but I suspect they will fail/timeout again. IMO we can merge this.

PR #980, opened 2020-03-19
ltmpl: Fix epoch value in package logs

bcl:
If it is 0 it should not be included.

Resolves: rhbz#1815000


rvykydal:
Another difference related to the BZ is that architecture has disappeared, is this expected?

bcl:
> Another difference related to the BZ is that architecture has disappeared, is this expected?

Nope. I'll grab an old iso and do a compare again this time :) Thanks!

bcl:
Ok, triple checked against an old boot.iso this time. epoch actually isn't included at all, except in debuginfo, so I removed that and added arch.

PR #982, opened 2020-04-01
tests: pip install future

atodorov:
Resolves:
https://github.com/ansible/ansible/issues/68616
https://github.com/cockpit-project/bots/pull/636

Another posibility would be to `yum install python2-future` from EPEL in vm.install instead.

@bcl - do I need to create a RHBZ for this ? If yes I could probably reuse some of my other bugzillas.


bcl:
> do I need to create a RHBZ for this

Yes. But you cannot reuse closed bugs, need new ones for 7.9

martinpitt:
Thanks @atodorov! This looks straightforward and blocks tests, so approving.

atodorov:
@bcl How about using 1785159 for this PR or you want a new one? Either way I need a devel_ack+

PR #983, opened 2020-04-08
System release

fepitre:


src/pylorax/cmdline.py
```diff
@@ -111,6 +112,8 @@ def lorax_parser(dracut_default=""):
                           help="Enable a DNF plugin by name/glob, or * to enable all of them.")
     optional.add_argument("--squashfs-only", action="store_true", default=False,
                           help="Use a plain squashfs filesystem for the runtime.")
+    optional.add_argument("--system-release", default=None, type=str,
```
bcl, line 115: I *really* don't want to add yet another cmdline switch to lorax. That's why I added commit 91d71b09a4b54d269492a22e4fb2d3f1e8d4e727 to use the variant name to select the release package. Is there a reason why that won't work for you?

bcl:
pep8 isn't part of the tests/checks that we run, so while it is an improvement it doesn't do much good if it isn't enforced so I'm not sure it really adds any value at this point.

src/pylorax/cmdline.py
fepitre: I understand your point. I'll try then variant possibility.

fepitre:
> pep8 isn't part of the tests/checks that we run, so while it is an improvement it doesn't do much good if it isn't enforced so I'm not sure it really adds any value at this point.

Sure no problem. Mostly I'm used to try to PEP8 every code I'm going to modify and sometimes it's not clear if a project enforces PEP8, so in case, I did it.

src/pylorax/cmdline.py
fepitre: I checked the code and for example, in case of Fedora, 'system-release' is 'fedora-release' and variant could be 'server'. In this case it fits for catching the good package. In our case, Qubes, we provide, 'qubes-release' and it's not really a variant of fedora as is. You commit would allow something like 'fedora-release-qubes'. @marmarek an comment about this?

bcl: Do you have more than one release file in your repos? This really is only a problem if more than one thing provides 'system-release'. If you have 'qubes-release' and 'qubes-logos' then it should be working fine.

marmarek: They are sorted based on name, right? The issue is that we use Fedora repos as a base directly, and we add our repositories on top. This makes `fedora-release` package chosen as it's before `qubes-release`.
If you don't want to add more options, here are some alternative ideas:
 - honor `exclude`/`includepkgs` options in repo definition
 - sort them not by name, but by version (or even version of `system-release` provides)
 - look for variant not only in package name, but also its provides
Is any of the above acceptable?

bcl: Can you change the name of your package to end in -qubes and pass --variant=qubes?
If that doesn't work, then I like applying exclude to it as a fallback. Sorry this is such a pain, here I thought the variant solution was an elegant one :)

marmarek: Is the variant suffix used in any product/variant/something in production? I have an impression that some places (not only lorax) relies on `*-release` naming schema. For example [few lines below](https://github.com/weldr/lorax/pull/983/files#diff-0472b55fe32084b66f441cb34168273eL120-L122) there is `-logos` name constructed:

        # logos
        release, _suffix = release.split('-', 1)
        self._runner.installpkg('%s-logos' % release)

This would work with `qubes-release-qubes`, but not with `release-qubes`.

bcl:
Ok, I think I've come up with a way to make things work for everyone, take a look at PR #996

fepitre:
> Ok, I think I've come up with a way to make things work for everyone, take a look at PR #996

Yes indeed. That will solve the problem. Thank you for your work. Feel free to close the PR

bcl:
This will be fixed by #996

PR #985, opened 2020-04-09
Update datastore for VMware testing

chris1984:


chris1984:
@atodorov @jikortus updated the commit message with the BZ and why the change.

chris1984:
@atodorov how do I kick off the tests again?

atodorov:
See https://github.com/weldr/lorax/blob/master/test/README.md#github-integration. 

mind you that the internal URL for 7.8 is gone now so use 7.9 image

PR #986, opened 2020-04-09
Update datastore for VMware testing

chris1984:


bcl:
> Probably we should disable the tests altogether. @bcl what do you think ?

I think we should continue to make sure we have working tests until the transition is finished.

bcl:
Merged to rhel8-branch with the bug added.

PR #738, opened 2019-05-29
Fix for Azure test

atodorov:
--- Description of proposed changes ---

the version coming from RPM is older and prevents installing
ansible[azure] via pip. OTOH removing with yum also removes
lorax-composer and breaks tests when running against the RPM
instead of git checkout (which will be the default with Cockpit CI).

Related: rhbz#1715003

--- Merge policy ---

- [ ] Travis CI PASS
- [ ] `*-aws-runtest` PASS
- [ ] `*-azure-runtest` PASS
- [ ] `*-images-runtest` PASS
- [ ] `*-openstack-runtest` PASS
- [ ] `*-vmware-runtest` PASS
- [ ] For `rhel8-*` and `rhel7-*` branches commit log references an approved
  bug in Bugzilla. Do not merge if the bug doesn't have the 3 ACKs set to `+`!

--- Jenkins commands ---

- `ok to test` to accept this pull request for testing
- `test this please` for a one time test run
- `retest this please` to start a new build

PR #990, opened 2020-04-15
Do not merge - test commit

jikortus:
Test commit for cockpit-project/bots #738


jikortus:
Most of the tests are OK now that the fixes for openstack and vmware tests have been merged. A fix for live-iso test is in #994. The only remaining failing test is the `ext4-filesystem` one, however I haven't found out yet what exactly is wrong. I can see an error message related to using some locale that has not been installed in the system, but this doesn't seem to cause the failure. It looks like there's something (slightly?) wrong with `subscription-manager attach --auto`, as the system looks like not subscribed afterwards, however the rest of the test passes fine. I tried to reproduce this locally, but without any luck so far.

jikortus:
All of the issues should be gone by now, and the remaining one (live iso) has a proposed fix, therefore closing the PR.

PR #992, opened 2020-04-16
Change --make-pxe-live --no-virt to use a fsimage

bcl:
Instead of a partitioned disk image. This will allow a single pass
of lmc to be used to create an ostree live PXE image.

Resolves: rhbz#1802591


bcl:
Merged to rhel7-branch for loax-19.7.27-1

PR #930, opened 2019-12-19
[7.9] Add test for running composer with --no-system-repos option

atodorov:
- add test coverage for rhbz#1650363
- cherry-picked from 5322664432631f61af831d1eef5f400ffb65b3f9

Related: rhbz#1785159

PR #993, opened 2020-04-17
Add --no-system-repos to lorax-composer

bcl:
Running lorax-composer --no-system-repos will prevent it from copying
the yum repositories from /etc/yum.repos.d/ into the lorax-composer repo
directory. It will *only* use repositories setup using the sources api
or written to /var/lib/lorax/composer/repos.d/

If lorax-composer has previously been run without this switch the system
repos will need to be removed from the composer/repos.d/ directory. It
would also be a good idea to remove the cached metadata in
/var/tmp/composer/

Resolves: rhbz#1785388

### DIFF
--- a/src/sbin/lorax-composer
+++ b/src/sbin/lorax-composer
@@ -79,6 +79,8 @@ def get_parser():
                         help="Top level temporary directory")
     parser.add_argument("--proxy", default=None, metavar="PROXY",
                         help="Set proxy for DNF, overrides configuration file setting.")
+    parser.add_argument("--no-system-repos", action="store_true", default=False,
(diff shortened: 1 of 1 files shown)


atodorov:
@bcl looks good to me but I can't proceed with tests b/c ext4 is failing (subscription issues, should be resolved with 7.9 image refresh). I've cherry-picked your commit in #930 to try the tests but all failed. It's up to you if you want to wait until we can get the tests passing again or merge as-is.

bcl:
I've tested it locally, and given the size of the change my confidence is high that the failures are unrelated, so I'll merge this.

PR #995, opened 2020-04-20
Use only BOS repo for Beaker repos in vm.install

jikortus:
The redirection from download.devel for Beaker repos doesn't work properly if run from a BRQ location. We can just directly use BOS repos in this case, as it shouldn't break anything and the amount of data transferred from this repo shouldn't cause any significant delay even from a BRQ site.

@atodorov if you think this is OK, I'll create a BZ to reference.

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -20,7 +20,7 @@ fi
 
 if ! rpm -q beakerlib; then
     if [ $(. /etc/os-release && echo $ID) = "rhel" ]; then
-        (cd /etc/yum.repos.d; curl -O -L http://download.devel.redhat.com/beakerrepos/beaker-client-RedHatEnterpriseLinux.repo)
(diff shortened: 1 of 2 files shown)


jikortus:
> The redirection does work, but results in a 404 for me. This is clearly a regression and needs to be addressed as such. I'm not sure who's responsible for this system, could be IT, RCM or Beaker team but you need a ticket for that.

Yes, the 404 was what I was getting at. Are you sure it really worked for a BRQ location in the past? The URL it redirects to looks fairly strange to me indeed.

jikortus:
FTR, it turned out that the best solution would be to remove the redirect (and serve the content from the original location). I created a ticket for that and will close this PR, however we still have to bear this issue in mind when running `vm.install` from a BRQ location, until this is fixed in the infrastructure.

PR #996, opened 2020-04-21
lorax: Add --skip-branding cmdline argument

bcl:
Also document how branding currently works. See docs/lorax.rst

Resolves: rhbz#1826479


fepitre:
Is this going to fall into FC32 branch too?

bcl:
> Is this going to fall into FC32 branch too?

Sure, no reason not to. It will have to wait until after the release though, we're currently in final freeze.

PR #997, opened 2020-04-22
Backport --skip-branding to RHEL8.2-branch

bcl:


bcl:
Pushed to rhel8.2-branch, built as part of lorax-28.14.42-2

PR #1000, opened 2020-04-24
rsyslog: Disable journal ratelimits during install

bcl:
Every log entry is sacred

Resolves: rhbz#1752754


bcl:
Tests passed, tar couldn't update github due to an error 502 Bad Gateway

PR #1001, opened 2020-04-27
test: Work around invalid fedora baseurls

marusak:
In the newest Fedora 31 refresh (and in Fedora 32) `baseurl` in repos
contains `download.example`. That is of course not a valid domain.

Fixes https://github.com/cockpit-project/bots/pull/771


marusak:
This should be possibly reported? No idea though what component I should report this against. If anyone has some idea, please let me know (or just feel free and go ahead with this reporting and just post here the report url). Thanks!

PR #1003, opened 2020-04-27
lmc-no-virt: Add requirement on anaconda-install-env-deps

bcl:
This makes sure that anaconda has all the expected modules available,
eg. libblockdev-plugins-all


bcl:
Merged into rhel8-branch

PR #1004, opened 2020-04-28
RHEL-7 download servers URLs fix in tests

jikortus:
The structure of URLs at the download servers had to be updated a bit, and also it's not possible to use latest repos based only on major version (minor version is needed as well). Using `7.9` for the latest repos seemed OK to me at this point, otherwise I think it would be fairly complicated to supply the major-minor versions somewhat dynamically.

On a somewhat related note, I also added a minor change to the `test_repos_sanity.sh` test to use both rel-eng/nightly repos depending on repos available in the system.


bcl:
Looks like glusterfs is now 6.* not 3.* could you include a commit to change that and get the tests passing?

jikortus:
@bcl, I ironed out the remaining issues, and the test suite should now work fine with the new URLs and use rel-eng/nightly repos as appropriate. As for the `test_repos_sanity.sh` test failure in RHEL-7.9, it's caused by missing repos in lorax-composer `repos.d` directory, for which I filed https://bugzilla.redhat.com/show_bug.cgi?id=1830897. I don't know why the sanity tests failed for RHEL-7.8, not sure if it's worth investigating as we have to switch to 7.9 anyway.

atodorov:
The failure in repo_sanity.sh is unrelated but should be fixed separately.

PR #1007, opened 2020-04-30
Master osbuild composer backend

atodorov:
This is the same as #917 but rebased on top of master branch.

My intention was to make sure composer-cli is using the osbuild-composer backend so we can have coverage for the scenarios where compose types are compared against a hard-coded list, see discussion here:
https://github.com/osbuild/osbuild-composer/pull/503#discussion_r408913577

Arguably most of the tests that we have here should be disabled b/c they are already covered in osbuild-composer. Only the cli specific ones should remain here. I'll update this PR with a few more commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ $(VM_IMAGE): srpm bots
 		--upload $(CURDIR)/test/vm.install:/var/tmp/vm.install \
 		--upload $(realpath tests):/ \
 		--run-command "chmod +x /var/tmp/vm.install" \
-		--run-command "cd /var/tmp; /var/tmp/vm.install $$srpm" \
+		--run-command "cd /var/tmp; BACKEND=$(BACKEND) /var/tmp/vm.install $$srpm" \
 		$(TEST_OS)
 	[ -f ~/.config/lorax-test-env ] && bots/image-customize \
 		--upload ~/.config/lorax-test-env:/var/tmp/lorax-test-env \
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ mandir ?= $(PREFIX)/share/man
 DOCKER ?= podman
 DOCS_VERSION ?= next
 RUN_TESTS ?= ci
+BACKEND ?= lorax-composer
 
 PKGNAME = lorax
 VERSION = $(shell awk '/Version:/ { print $$2 }' $(PKGNAME).spec)
@@ -173,10 +174,10 @@ vm-local-repos: vm
 		$(TEST_OS)
 	bots/image-customize -v \
 		--upload $(REPOS_DIR):/etc/yum.repos.d \
-		--run-command "yum -y remove composer-cli lorax-composer" \
+		--run-command "yum -y remove composer-cli $(BACKEND)" \
 		--run-command "yum -y update" \
-		--run-command "yum -y install composer-cli lorax-composer" \
-		--run-command "systemctl enable lorax-composer" \
+		--run-command "yum -y install composer-cli $(BACKEND)" \
+		--run-command "systemctl enable $(BACKEND).socket" \
 		$(TEST_OS)
 
 vm-reset:
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -127,7 +127,7 @@ class ComposerTestCase(VirtMachineTestCase):
         # Upload the contents of the ./tests/ directory to the machine (it must have beakerlib already installed)
         self.machine.upload(["../tests"], "/")
 
-        print("Waiting for lorax-composer to become ready...")
+        print("Waiting for backend to become ready...")
         curl_command = ["curl", "--max-time", "360",
                                 "--silent",
                                 "--unix-socket", "/run/weldr/api.socket",
@@ -149,7 +149,7 @@ class ComposerTestCase(VirtMachineTestCase):
         return local_dir
 
     def runCliTest(self, script):
-        extra_env = []
+        extra_env = ["BACKEND=%s" % os.getenv('BACKEND', 'lorax-composer')]
         if self.sit:
             extra_env.append("COMPOSER_TEST_FAIL_FAST=1")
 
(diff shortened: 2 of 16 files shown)


atodorov:
@bcl I see `2020-05-01 07:17:52,596: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)` as a failure which happens when the 2nd of these lines is executed:

```
        rlRun -t -c "$CLI blueprints push $(dirname $0)/lib/test-http-server.toml"
        UUID=`$CLI compose start test-http-server tar`
```

Do you have any idea why we get a timeout ?

bcl:
> Do you have any idea why we get a timeout ?

With osbuild-composer running or lorax?
I can't think of anything that would cause that, other than the API crashing, for either backend. In lorax it's making a git commit and in osbuild-composer it's adding it to the store (which is in memory).

atodorov:
> With osbuild-composer running or lorax?

osbuild-composer.

FTR I am trying to reproduce this locally but I'm hitting a different error.

composer-cli compose start passes (doesn't fail like here in the logs) but the actual compose exits immediately with status FAIL and the subsequent `cancel` fails!

atodorov:
The failing `composer-cli compose cancel` is a genuine issue, see https://github.com/osbuild/osbuild-composer/pull/180

bcl:
> The failing `composer-cli compose cancel` is a genuine issue, see [osbuild/osbuild-composer#180](https://github.com/osbuild/osbuild-composer/pull/180)

Right, the backends aren't 100% identical. Another place where they differ is uploading which is implemented differently than what we did in Fedora.

atodorov:
Note: osbuild-composer v15 still doesn't seem to be in the Fedora repositories.

atodorov:
@bcl, @larskarlitski this is ready for review. You also have to adjust the branch settings of this repository to require the Fedora32 test jobs instead of the Fedora31 ones.

atodorov:
Also let me know if I should start removing some of these test jobs instead b/c the origin of this PR goes way back to when osbuild-composer didn't have its own test suite and we were trying to adopt the current one to be backend independent.

Makefile
```diff
@@ -5,6 +5,7 @@ mandir ?= $(PREFIX)/share/man
 DOCKER ?= podman
 DOCS_VERSION ?= next
 RUN_TESTS ?= ci
+BACKEND ?= osbuild-composer
```
bcl, line 8: I think this should default to lorax-composer, the code isn't dead yet, and this is what it should be tested against by default.

I think it should run the tests twice, once against the default, and again against osbuild-composer.
And osbuild-composer's gating tests should also run this with the osbuild-composer backend set.

larskarlitski: Yeah, agreed.

atodorov:
@bcl can you check the `osbuild-composer` log. I still see:
```
{"api":"1","backend":"lorax-composer","build":"33.6-1","db_supported":true,"db_version":"0","msgs":[],"schema_version":"0"}
```

and I'm not sure what am I missing

Makefile
```diff
@@ -5,7 +5,10 @@ mandir ?= $(PREFIX)/share/man
 DOCKER ?= podman
 DOCS_VERSION ?= next
 RUN_TESTS ?= ci
-BACKEND ?= lorax-composer
+ifeq ($(BACKEND),)
```
bcl, line 8: How is this different from `BACKEND ?= lorax-composer` ?

bcl:
> @bcl can you check the `osbuild-composer` log. I still see:

It doesn't look like the vm-local-repos target is being run. It looks like it builds the lorax rpm and installs and enables it.

Makefile
atodorov: IDK what the other syntax does, my intention here is to have the BACKEND variable default to "lorax-composer" if it is not set. It looks like that is the case without the if condition, right ?

atodorov:
> repos

We don't execute `make vm-local-repos` here, that was intended for manual testing on RHEL snapshots only. 

The script should be doing `make vm`, in the case of the osbuild-composer scenario doing this twice (2nd time after we adjust the value of the BACKEND variable). I can see make vm being executed twice in the log but the second time also uses lorax-composer. That's what I can't figure out.

Makefile
bcl: Right, the ?= syntax in a makefile sets it to that default if there is no environmental variable set

atodorov:
@bcl I found the missing piece which was not letting us start osbuild-composer from the test script:
https://github.com/weldr/lorax/pull/1007/files#diff-daf79794fc7fc030d0d6a03e740a7311R152

In the fedora32/osbuild-composer log I can now see that osbuild-composer has been used for testing instead of lorax-composer.


This PR is ready for review.

@bcl: do you want me to cherry-pick your changes which also assert that the backend which was started is the expected one ? (in composertest.py)

Note: this PR was based onto #1050 but now that we got it working I would question if we really need to merge 1050. Your call.

bcl:
I think PR #1050 is the right direction to go, even with the osbuild-composer backend running the tests won't all pass and will need tweaks to make the results work for both. I think we can just close this unless someone has strong objections. 

*thanks* for all your work on it.

atodorov:
> I think PR #1050 is the right direction to go, even with the osbuild-composer backend running the tests won't all pass and will need tweaks to make the results work for both. I think we can just close this unless someone has strong objections.
> 

@bcl I do object to closing this PR. Even if we don't execute the image build tests with the new backend we still have the cli tests and they are passing with the current changes.  The changes related to not using the example blueprints are also needed internally for the gating tests. It did take a long time to figure that one out but seeing how the test suite is finally working IMO it will be better not to ditch it entirely.

bcl:
> @bcl I do object to closing this PR. Even if we don't execute the image build tests with the new backend we still have the cli tests and they are passing with the current changes. The changes related to not using the example blueprints are also needed internally for the gating tests. It did take a long time to figure that one out but seeing how the test suite is finally working IMO it will be better not to ditch it entirely.

Ah! Thanks for the explanation, I hadn't realized it was still useful. I'll take one more look at it today. Looks like the tests are passing :)

PR #1009, opened 2020-05-04
Dockerfile.test: Switch to using F32 for testing

bcl:


bcl:
Merged into f32-branch

PR #1011, opened 2020-05-05
Fix repos files handling in test_repos_sanity.sh

jikortus:
As there are no system repos copied over to lorax-composer's repos.d directory, we need to reflect that in the `test_repos_sanity.sh` test.

### DIFF
--- a/tests/cli/test_repos_sanity.sh
+++ b/tests/cli/test_repos_sanity.sh
@@ -10,9 +10,9 @@ CLI="${CLI:-./src/bin/composer-cli}"
 
 rlJournalStart
     rlPhaseStartSetup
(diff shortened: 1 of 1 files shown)


bcl:
I can't tell why those 2 tests are failing in `compose start`, restarted rhel-7-9 and it failed with a different error the 2nd time.

jikortus:
Hmm, this is strange. Might that be some transient network/server issue? I could see the error that was present before, composer-cli returned 1 for some reason (I suspect slow network connection to RHN and timeout during waiting for response from lorax-composer, but I may be wrong). As for the vmware problem, I'm just discussing it with @chris1984.

PR #1017, opened 2020-05-19
drop 32-bit support from ppc live image grub.cfg

sharkcz:
Seems petitboot can't properly parse the live image grub config on ppc, thus
booting fails on bare-metal. Fix the problem by removing the obsolete 32-bit
entries.


bcl:
Test failures are unrelated to this change.

PR #1020, opened 2020-05-21
Keep /etc/default/useradd in install.img

jstodola:
Anaconda creates a user for connecting via ssh during the installation when the sshpw kickstart command is used. The user is created with "/sbin/nologin" shell when /etc/default/useradd is missing, which prevents the user from log into the running installation.

Resolves: rhbz#1838677


bcl:
I've cherry picked this to master so it doesn't get lost.

PR #1023, opened 2020-05-28
Dummy PR

martinpitt:
I just want to see how tests behave on the current rhel-7-9 image, to
see if https://github.com/cockpit-project/bots/pull/908 is a regression.

### DIFF
--- a/README
+++ b/README
@@ -1,3 +1,5 @@
+pwned
+
 I am the Lorax.  I speak for the trees [and images].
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Ack, fails. I *think* I know why -- it just creates a ginormous VM overlay, which overflows our /tmp/ tmpfs. I'll send a fix soon.

martinpitt:
Fixing in #1027, and will backport once that landed.

PR #1027, opened 2020-06-01
test: Put VM image overlay into /var/tmp

martinpitt:
At least in our CI, the default place to store VM runtime overlays
(testvm.get_temp_dir()) points to a tmpfs file, so that tests can put VM
overlays into RAM and are not affected by slow I/O. But that doesn't
work for composer, as it tends to produce huge overlays due to real-life
OS composed trees.

Use /var/tmp/ instead, which is meant for large files.

 - [x] Depends on https://github.com/cockpit-project/bots/pull/926

### DIFF
--- a/test/composertest.py
+++ b/test/composertest.py
@@ -35,10 +35,11 @@ class VirtMachineTestCase(unittest.TestCase):
 
     def setUpTestMachine(self, image=testvm.DEFAULT_IMAGE, identity_file=None):
         self.network = testvm.VirtNetwork(0)
-        if identity_file:
-            self.machine = testvm.VirtMachine(image, networking=self.network.host(), cpus=2, memory_mb=2048, identity_file=identity_file)
-        else:
-            self.machine = testvm.VirtMachine(image, networking=self.network.host(), cpus=2, memory_mb=2048)
+        # default overlay directory is not big enough to hold the large composed trees; thus put overlay into /var/tmp/
+        self.machine = testvm.VirtMachine(image, networking=self.network.host(),
+                                          cpus=2, memory_mb=2048,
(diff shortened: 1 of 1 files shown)


martinpitt:
This is meant to fix the tar timeouts like [this](https://logs.cockpit-project.org/logs/pull-1024-20200601-182816-4b1a4f8f-fedora-31-tar/log.html).

Once this lands, we'll also need this on the rhel-X branch.

martinpitt:
[travis failure](https://travis-ci.org/github/weldr/lorax/builds/693610566?utm_source=github_status&utm_medium=notification) looks odd, like a Fedora repo fluke. Retrying.

martinpitt:
Travis failure repeats, but it also happened in e. g. #1024, so unrelated.

bcl:
> [travis failure](https://travis-ci.org/github/weldr/lorax/builds/693610566?utm_source=github_status&utm_medium=notification) looks odd, like a Fedora repo fluke. Retrying.

I think the rawhide repos are currently useless until everything is rebuilt for 3.9, or the docker image for rawhide needs to be updated for 3.9. Neither of which are failures we can do anything about but wait.

martinpitt:
Nice, [that works](https://logs.cockpit-project.org/logs/pull-1027-20200602-055227-21d49bf2-fedora-31-tar-bots-926/log.html)! The bots PR landed. Unblocking, re-running all tests.

martinpitt:
@bcl: Travis is a required status, so this can't be merged yet. But you said it's been broken for a while already; apparently #1024 was merged with some kind of "admin override"? Can you please land that? Then I'll backport it to the stable branches, so that CI will finally work again.

PR #1031, opened 2020-06-05
Master size ostree

bcl:
This currently doesn't have any tests for the new features. But I think this is working, I can pass --size and observe it using that size in the results json. And if I do a start-ostree fedora-iot-commit I can see the ref and parent in the json.


bcl:
PR #1033 is the RHEL8 version

PR #1035, opened 2020-06-16
Ensure /proc exists before trying to touch /proc/modules

AdamWill:
The last few Rawhide composes failed because, when we try to
touch /proc/modules here (in treebuilder.py), /proc does not
exist in the inroot. I tried and failed all afternoon to figure
out *why* it suddenly doesn't exist; I'd still like to know, but
this at least should make it not an issue. `exist_ok=True` means
that if the dir already exists we'll do nothing and carry on.

### DIFF
--- a/src/pylorax/creator.py
+++ b/src/pylorax/creator.py
@@ -256,6 +256,7 @@ def rebuild_initrds_for_live(opts, sys_root_dir, results_dir):
         raise Exception("No initrds found, cannot rebuild_initrds")
 
     # Hush some dracut warnings. TODO: bind-mount proc in place?
+    os.makedirs(joinpaths(sys_root_dir, "/proc"), exist_ok=True)
     open(joinpaths(sys_root_dir,"/proc/modules"),"w")
 
     if opts.ostree:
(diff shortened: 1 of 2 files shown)


AdamWill:
@bcl @nirik @mohanboddu

AdamWill:
The mystery here sure is Very Mysterious, BTW. I can reproduce it locally so long as I run lorax inside mock. If I run lorax directly on my system it works fine. If I run it in mock, bug happens. It happens with both mock 2.0 and mock 2.3. It happens whether I use the contents of the last stable compose or the contents of current Koji - both for the mock root itself (`mock -r fedora-rawhide-x86_64` vs `mock -r fedora-rawhide-x86_64-koji`) and for the lorax run (`-s` args passed to lorax). It happens with both `--old-chroot` and `--new-chroot`. `/proc` is part of the `filesystem` package, which definitely creates and contains it, and `filesystem` is definitely installed into the inroot. I can even `chroot` into the inroot during the lorax run and run `rpm -V filesystem` and it shows no errors...but there is no `/proc`. It's weird.

bcl:
Figured out what was happening, ends up it is mock writing a .rpmmacro inside the mock. It sets netsharedpath to /proc:/sys which make rpm skip creating them, even when inside a different root.

Filed a bug with mock - https://bugzilla.redhat.com/show_bug.cgi?id=1848201

PR #1038, opened 2020-06-17
composer-cli: Change timeout to 5 minutes

bcl:
Related: rhbz#1844649

In some cases osbuild-composer with an empty cache can take longer than
a minute to depsolve and start a compose.


jikortus:
Thanks for the fix, @bcl, hopefully it will help. I tried to prepare a VM for you for some debugging, but wasn't able to reproduce the related bug again on a different host. Maybe I can reproduce it with the local VM image I used earlier, but don't know if it would be of any use to you.

bcl:
> Maybe I can reproduce it with the local VM image I used earlier, but don't know if it would be of any use to you.

Thanks, I think I've got a good handle on it now. It's easy to reproduce if you lower the timeout to something crazy like 5s

PR #1041, opened 2020-06-22
Skip cloud-init check for tests on F32

atodorov:
I will kick off tests with Fedora 32 manually


atodorov:
@bcl, @larskarlitski the test log for Azure shows a failure for `systemctl status cloud-init` which causes the job to fail. Have you seen this before ?

bcl:
I haven't noticed that one before, and the error log there doesn't really say much other than what looks like part of a ssh key image and `cloud-init.service: Failed with result 'exit-code'.`

atodorov:
FYI osbuild/osbuild-composer#698

bcl:
> FYI [osbuild/osbuild-composer#698](https://github.com/osbuild/osbuild-composer/issues/698)

The lorax-composer template for vhd has it enabled though, and from the logs it is trying to start, it just doesn't log why it failed to start.

`services --enabled=sshd,chronyd,waagent,cloud-init,cloud-init-local,cloud-config,cloud-final`

atodorov:
IDK what the new failures are from but definitely not related to the switch to Fedora32 or the skipping of the cloud0init check. Can we get this merged so we can default to the fedora-32 test images and proceed to leaving only composer-cli related test jobs in this repository ?

PR #1042, opened 2020-06-22
Skip creating empty /proc/modules for dracut

bcl:
Creating /proc/modules just quiets one dracut warning, it doesn't
actually effect the initrd that is created.

Now that mock is being run with netsharedpath set to exclude /proc is
causes problems for running lorax inside a mock, so just remove this to
prevent a crash when it doesn't exist.

Related: rhbz#1848201

### DIFF
--- a/src/pylorax/creator.py
+++ b/src/pylorax/creator.py
@@ -255,9 +255,6 @@ def rebuild_initrds_for_live(opts, sys_root_dir, results_dir):
     if not kernels:
         raise Exception("No initrds found, cannot rebuild_initrds")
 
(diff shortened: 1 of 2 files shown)


AdamWill:
seems fine to me, this has caused us trouble more than once before so dropping it seems reasonable. I guess we could log that the dracut warning is expected and probably harmless just to prevent us forgetting about this later and worrying about it?

atodorov:
FYI https://github.com/osbuild/osbuild-composer/issues/698

bcl:
>  I guess we could log that the dracut warning is expected

Probably a good idea.

PR #1048, opened 2020-07-16
Install weak dependencies of anaconda-install-env-deps

poncovka:
It looks like lorax doesn't install weak dependencies of anaconda-install-env-deps
on the boot.iso, so let's install these packages explicitly.

Related changes:
https://fedoraproject.org/wiki/Changes/SwapOnZRAM
https://fedoraproject.org/wiki/Changes/RemoveDeviceMapperMultipathFromWorkstationLiveCD

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -103,6 +103,9 @@ installpkg xfsdump
 ## extra storage packages
 # hostname is needed for iscsi to work, see RHBZ#1593917
 installpkg udisks2 udisks2-iscsi hostname
+%if basearch in ("i386", "x86_64"):
+    installpkg fcoe-utils
+%endif
 
 ## extra libblockdev plugins
 installpkg libblockdev-lvm-dbus
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -111,6 +111,9 @@ installpkg libblockdev-lvm-dbus
 installpkg volume_key
 installpkg nss-tools
 
+## enable swap on zram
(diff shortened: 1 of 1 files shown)


bcl:
Correct, Lorax doesn't install weak deps. But I have to wonder why these are weak deps? I'd have expected anaconda-install-env-deps to require everything needed.

cmurf:
Maybe this PR
https://pagure.io/fedora-comps/pull-request/515

will avoid the one for lorax? Another possibility is adding 'zram-generator-defaults' to the 'anaconda-tools' group. If so, nack PR515 and I'll resubmit.

poncovka:
> Correct, Lorax doesn't install weak deps. But I have to wonder why these are weak deps? I'd have expected anaconda-install-env-deps to require everything needed.

The fcoe-utils package depends on device-mapper-multipath and device-mapper-multipath shouldn't be installed on Live OS.
The zram-generator-defaults needs to be optional, because Fedora Cloud excludes it. We will revisit the idea of weak dependencies in anaconda-install-env-deps, but these packages need to be installed by lorax anyway.

poncovka:
> Maybe this PR
> https://pagure.io/fedora-comps/pull-request/515
> 
> will avoid the one for lorax? Another possibility is adding 'zram-generator-defaults' to the 'anaconda-tools' group. If so, nack PR515 and I'll resubmit.

I don't think that lorax installs groups.

bcl:
> The fcoe-utils package depends on device-mapper-multipath and device-mapper-multipath shouldn't be installed on Live OS.
> The zram-generator-defaults needs to be optional, because Fedora Cloud excludes it. We will revisit the idea of weak dependencies in anaconda-install-env-deps, but these packages need to be installed by lorax anyway.

Ah, of course :) Thanks for the explanation.

PR #1049, opened 2020-07-17
Fix starting osbuild-composer

bcl:
This is based on PR #1007 from @atodorov. and changes should be made to that PR. Do not merge this one.

PR #1050, opened 2020-07-21
tests: Disable non-cli test scenarios b/c osbuild-composer

atodorov:
the image building functionality is now tested by the
osbuild-composer test suite. We weren't successfull at making this
test suite work with the 2 backends so decided to disabled these
scenarios instead. In the future they will likely be deleted.

Proposing this PR instead of #1007 and #1049 after talking yesterday with Brian. The idea here is to disable as much as possible (I am still leaving the cli scenarios) and then work downstream to rely on reverse dependency gating tests to cover composer-cli. 

@bcl maybe we'd like to split composer-cli into a separate repo/package at some point. 

@larskarlitski should I go ahead and remove all of the disabled scenarios from `testmap.py` as well ?

### DIFF
--- a/test/check-api
+++ b/test/check-api
@@ -1,9 +1,13 @@
 #!/usr/bin/python3
 
+import unittest
+
 import composertest
 import requests
 import subprocess
 
+
+@unittest.skip("API is now tested by osbuild-composer test suite")
 class TestApi(composertest.ComposerTestCase):
     """Test Composer HTTP API"""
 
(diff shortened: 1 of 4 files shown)


atodorov:
> Yep, this makes sense. Thanks!

Can you merge b/c I don't have sufficient permissions to merge when Required test results are missing.

teg:
@bcl could you merge this, or is anything outstanding with the tests here?

PR #1053, opened 2020-07-24
tests: Allow skipping image build in compose sanity test

atodorov:
when $SKIP_IMAGE_BUILD is set the test skips 2 phases:
- start the compose again (after it has been cancelled) and
- waiting for it to finish & downloading the resulting image

This will allow to enable this test script in downstream gating
jobs.

### DIFF
--- a/tests/cli/test_compose_sanity.sh
+++ b/tests/cli/test_compose_sanity.sh
@@ -76,19 +76,12 @@ rlJournalStart
         fi
     rlPhaseEnd
 
+if [ -z "$SKIP_IMAGE_BUILD" ]; then
     rlPhaseStartTest "compose start again"
         UUID=`$CLI compose start test-http-server qcow2`
         rlAssertEquals "exit code should be zero" $? 0
 
         UUID=`echo $UUID | cut -f 2 -d' '`
-    rlPhaseEnd
-
-    rlPhaseStartTest "compose info"
-        if [ -n "$UUID" ]; then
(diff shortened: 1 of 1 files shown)


atodorov:
@bcl this is now ready for review. Here are logs from a commit which specified SKIP_IMAGE_BUILD=1:

https://logs.cockpit-project.org/logs/pull-1053-20200724-104526-d30cce34-fedora-32/log.html

https://logs-https-cockpit.apps.ci.centos.org/logs/pull-1053-20200724-104526-d30cce34-fedora-32-osbuild-composer/log.html

Search for:
```
:: [ 06:55:59 ] :: [   INFO   ] :: Skipping image build phases
```

tests/cli/test_compose_sanity.sh
```diff
@@ -69,28 +69,29 @@ rlJournalStart
         fi
     rlPhaseEnd
 
+    rlPhaseStartTest "compose info"
```
bcl, line 72: This isn't really needed, the previous block makes sure it is RUNNING (or fails).

atodorov: Right. That was explicitly designed as a phase to cover the `compose info` command, maybe because we didn't use it before when checking for the status or just duplicate.  Should I remove it  then ?

bcl: May as well remove it.

atodorov: Removed

PR #1057, opened 2020-08-04
test: Don't upgrade pip packages

marusak:
See https://github.com/cockpit-project/bots/pull/1115

As per comment `...but since py2.7 is dead it's probably not a good idea to upgrade it`.

Lets see if this was the problem or not really.

### DIFF
--- a/tests/cli/test_build_and_deploy_aws.sh
+++ b/tests/cli/test_build_and_deploy_aws.sh
@@ -45,7 +45,7 @@ rlJournalStart
         rlRun -t -c "virtualenv $VENV"
         source $VENV/bin/activate
 
-        rlRun -t -c "pip install --upgrade pip setuptools"
+        rlRun -t -c "pip install pip setuptools"
         rlRun -t -c "pip install awscli"
 
         # aws configure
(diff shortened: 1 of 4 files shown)


marusak:
Both tests with bots from #1115 passed, thus this fixes the problem we encountered.

atodorov:
I see a failure for AWS but it's not related.

marusak:
> Add Related: rhbz#1785159 to the commit log otherwise I can't merge.

Cannot that be done on squashing? Or is it preventing you from clicking the button? (just that tests are green :) )

atodorov:
> > Add Related: rhbz#1785159 to the commit log otherwise I can't merge.
> 
> Cannot that be done on squashing? 

It can but you could have also done this. You know that all commits on RHEL related branches need an approved BZ.

marusak:
> You know that all commits on RHEL related branches need an approved BZ

In dist-git, not in upstream. And I guess you just get tarballs into dist-git and create specific commit. Or you sync all commits somewhere else?

atodorov:
> In dist-git, not in upstream. And I guess you just get tarballs into dist-git and create specific commit. Or you sync all commits somewhere else?

@bcl can give you more information on the actual process but in this repo we've always followed the BZ rule, even for commits on GitHub.

bcl:
> In dist-git, not in upstream. And I guess you just get tarballs into dist-git and create specific commit. Or you sync all commits somewhere else?

lorax is built from a distro specific branch, with a new version. We require Related or Resolves in the commit messages upstream. Granted, this is in Extras so the rules are more loose, but I like to have valid BZ numbers anyway, makes things more consistent.

PR #1060, opened 2020-08-05
tests: Push example blueprints b/c they may not exist

atodorov:
Related: rhbz#1825190

### DIFF
--- a/tests/composer/test_blueprints.py
+++ b/tests/composer/test_blueprints.py
@@ -121,6 +121,16 @@ dict_name_results3 = "/root, /home/norm, /home/cliff"
 
 
 class BlueprintsTest(unittest.TestCase):
+    @unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Setup requires a running API server")
+    @classmethod
+    def setUpClass(cls):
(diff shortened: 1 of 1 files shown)


tests/composer/test_blueprints.py
```diff
@@ -121,6 +121,15 @@
 
 
 class BlueprintsTest(unittest.TestCase):
+    @unittest.skipUnless(os.path.exists("/run/weldr/api.socket"), "Setup requires a running API server")
+    @classmethod
+    def setUpClass(cls):
+        """
+            Make sure that example blueprints are pushed to the server!
+        """
+        rc = blueprints_push("/run/weldr/api.socket", 0, ["example-http-server.toml"], show_json=False)
```
bcl, line 130: It will also need example-development

atodorov: I had the impression example-development.toml is uploaded by one of the tests but I will add it.

bcl: It depends on which test, when I run the nose tests against osbuild-composer it complains that example-devel isn't installed.

PR #1062, opened 2020-08-10
tests: Fix diff test BACKEND check

bcl:
The environmental variable we are using is 'BACKEND' not
'COMPOSER_BACKEND'

Related: rhbz#1825190

### DIFF
--- a/tests/composer/test_blueprints.py
+++ b/tests/composer/test_blueprints.py
@@ -268,7 +268,7 @@ class BlueprintsTest(unittest.TestCase):
 
     # XXX MUST COME LAST
     # XXX which is what _z_ ensures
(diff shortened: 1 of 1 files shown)


tests/composer/test_blueprints.py
```diff
@@ -268,7 +268,7 @@ def test_workspace(self):
 
     # XXX MUST COME LAST
     # XXX which is what _z_ ensures
-    @unittest.skipIf(os.getenv("COMPOSER_BACKEND") == "osbuild-composer",
+    @unittest.skipIf(os.getenv("BACKEND") == "osbuild-composer",
```
atodorov, line 271: Good catch!

PR #1063, opened 2020-08-25
Adds oVirt/RHV as a provider to upload an image to.

dbrandonjohnson:


bcl:
Thanks for the patch! But we are deprecating lorax-composer in favor of osbuild-composer so will not be adding any more features to it. It will likely be removed in Fedora 34 or 35.

See https://github.com/osbuild/osbuild-composer/ for more information.

PR #1065, opened 2020-08-27
tests: Ignore W0707 raise-missing-from warnings

bcl:


bcl:
The rpmfluff failure is a bug in rpmfluff, fixed by v0.6.1 which isn't in Fedora 33 yet - https://bugzilla.redhat.com/show_bug.cgi?id=1873340

PR #1067, opened 2020-08-31
templates: Ensure nano is installed for the runtime environment

Conan-Kudo:
The expectation is that all environments where an editor might be
used should ship GNU nano by default and tools should activate it
when an "editor" is requested. This change should ensure that for
the install media runtime environment.

Reference: https://fedoraproject.org/wiki/Changes/UseNanoByDefault

Resolves: [rhbz#1874094](https://bugzilla.redhat.com/show_bug.cgi?id=1874094)

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -180,6 +180,7 @@ installpkg fpaste
 installpkg python3-pyatspi
 
 ## extra tools not required by anaconda
+installpkg nano nano-default-editor
 installpkg vim-minimal strace lsof dump xz less
 installpkg wget rsync bind-utils ftp mtr vconfig
 installpkg icfg spice-vdagent


AdamWill:
of course it's arguable whether we should drop vim-minimal at the same time. that might annoy people though :)

Conan-Kudo:
@AdamWill I'm not going to break people's muscle memory. 😜

share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -180,6 +180,7 @@ installpkg fpaste
 installpkg python3-pyatspi
 
 ## extra tools not required by anaconda
+installpkg nano nano-default-editor
```
bcl, line 183: What provides nano-default-editor? I'm not seeing it in kojo or dist-git...

bcl: Found it, it's a subpackage of nano.

PR #1070, opened 2020-09-05
Runtime install cleanup

AdamWill:
This bundles up a few changes I came up with to try and improve on netinst size - see https://pagure.io/releng/issue/9689 . With all these changes a test image I built from yesterday's Branched tree came out to *exactly* the size limit, an overall improvement of ~17MB. I'm still working on more, but I think these are good changes for now.


AdamWill:
It'd be good if someone else can go through https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/WHENCE and double-check the firmware changes.

AdamWill:
I did a simple default install test in a VM with the image I built, and it worked.

Conan-Kudo (approved):

share/templates.d/99-generic/runtime-cleanup.tmpl
```diff
@@ -240,10 +240,37 @@ removefrom libvorbis /usr/${libdir}/libvorbisenc.*
...
+removefrom linux-firmware /usr/lib/firmware/TDA7706*
+removefrom linux-firmware /usr/lib/firmware/tlg2300_firmware.bin
+removefrom linux-firmware /usr/lib/firmware/s5p-mfc*
+removefrom linux-firmware /usr/lib/firmware/go7007/*
+removefrom linux-firmware /usr/lib/firmware/intel/IntcSST2.bin
+removefrom linux-firmware /usr/lib/firmware/intel/fw_sst*
+removefrom linux-firmware /usr/lib/firmware/intel/dsp*
+removefrom linux-firmware /usr/lib/firmware/as102*
+removefrom linux-firmware /usr/lib/firmware/qcom/venus*/*
+removefrom linux-firmware /usr/lib/firmware/meson/vdec/*
```
lzap, line 268: I spotted Yamaha and Korg directories, these sound like sound firmware :-)

AdamWill: ah, you do have to cross-check with the Fedora package file lists; quite a lot of sound firmwares are already split out into `alsa-firmware` and not actually carried in our `linux-firmware` package. Some of the things in my list are sound firmwares that haven't been split out, though (the Intel `fw_sst` and `dsp` ones). We could probably report a bug on linux-firmware or alsa-firmware to have those split out I guess.

bcl:
@M4rtinK or @poncovka should also have a look at these to make sure there isn't something that will surprise anaconda/blivet/etc.

share/templates.d/99-generic/runtime-cleanup.tmpl
```diff
@@ -240,10 +240,37 @@ removefrom libvorbis /usr/${libdir}/libvorbisenc.*
 removefrom libxml2 /usr/bin/*
 removefrom libxml2-python /usr/${libdir}/python?.?/site-packages/libxml2mod.a
 removefrom libxml2-python /usr/${libdir}/python?.?/site-packages/libxml2mod.la
+removefrom linux-firmware /usr/lib/firmware/dvb*
+removefrom linux-firmware /usr/lib/firmware/*_12mhz*
+removefrom linux-firmware /usr/lib/firmware/v4l*
+removefrom linux-firmware /usr/lib/firmware/brcm/BCM-*
```
bcl, line 246: This looks like it may include some broadcom related wifi firmware, and some things related to Raspberry Pi.
I don't know how to tell for sure if they are important, do you?

AdamWill: The first three are all TV capture stuff. The fourth I meant to capture the very first driver in the `WHENCE` list, which is documented as "Cypress Bluetooth firmware for HTC Vive" - the Vive is a VR headset. I did stop removing *anything* to do with bluetooth, wifi, serial ports, PCMCIA etc. after that though, and I guess the wildcard might inadvertently capture a similarly-named thing added later. We could drop that line I guess.

share/templates.d/99-generic/runtime-cleanup.tmpl
```diff
@@ -240,6 +240,32 @@ removefrom libvorbis /usr/${libdir}/libvorbisenc.*
...
+removefrom linux-firmware /usr/lib/firmware/ttusb-budget/dspbootcode.bin
+removefrom linux-firmware /usr/lib/firmware/emi26/*
+removefrom linux-firmware /usr/lib/firmware/emi62/*
+removefrom linux-firmware /usr/lib/firmware/cpia2/*
+removefrom linux-firmware /usr/lib/firmware/dabusb/*
+removefrom linux-firmware /usr/lib/firmware/vicam/*
+removefrom linux-firmware /usr/lib/firmware/dsp56k/*
+removefrom linux-firmware /usr/lib/firmware/sun/*
```
poncovka, line 254: If I am not mistaken, it is an ethernet driver. Might be useful.

poncovka:
Otherwise, it looks good to me.

share/templates.d/99-generic/runtime-cleanup.tmpl
AdamWill: I mean. If we think anyone is still using a Sun PCI ethernet card from 2004...

share/templates.d/99-generic/runtime-cleanup.tmpl
bcl: Ah, right, looks like they named everything else with the brcm prefix. It can stay.

bcl (approved):
Looks ok to me.

AdamWill:
So my plan here is we see how this affects the next Rawhide compose, and if it seems OK, then apply to F33.

@mohanboddu what do we need to do for the changes to take effect in a Rawhide compose? New lorax build for Rawhide?

mohanboddu:
> @mohanboddu what do we need to do for the changes to take effect in a Rawhide compose? New lorax build for Rawhide?

@AdamWill Yes, once the build is in buildroot, we can start a compose.

PR #1073, opened 2020-09-09
Runtime install cleanup 33

AdamWill:
Same as https://github.com/weldr/lorax/pull/1070 but for F33. Ideally we'd get a working Rawhide compose and run it through openQA before merging this, but if Rawhide keeps being a pain we might have to just cross our fingers and do it.


sgallagher:
https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Server/x86_64/iso/ looks like Rawhide completed yesterday and it is definitely producing a smaller (though not yet small _enough_) netinstall image. Can we get this merged?

PR #1074, opened 2020-09-11
Also remove .opt-1.pyc and .opt-2.pyc, not just .pyo

hroncok:
.pyo is for Python < 3.5 only.

cc @AdamWill could you include this in your testing?

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -353,6 +353,7 @@ removefrom ${product.name}-logos /usr/share/{firstboot,gnome-screensaver,kde4,pi
 
 ## cleanup_python_files()
 runcmd find ${root} -name "*.pyo" -type f -delete
+runcmd find ${root} -name "*.opt-?.pyc" -type f -delete
 runcmd find ${root} -name "*.pyc" -type f -exec ln -sf /dev/null {} \;
 
 ## cleanup /boot/ leaving vmlinuz, and .*hmac files


bcl:
Looking at the current rawhide image I don't see any .pyc or .pyo files so I'm not sure we need this.

AdamWill:
What happens is that the very *next* line makes all `.pyc` files a symlink to `/dev/null`, and then later, [this bit](https://github.com/weldr/lorax/blob/master/share/templates.d/99-generic/runtime-cleanup.tmpl#L367) removes them.  You can see it in the logs:

    removing broken symbolic link /usr/lib/python3.9/site-packages/lvmdbusd/__pycache__/vg.cpython-39.pyc -> /dev/null
    removing broken symbolic link /usr/lib/python3.9/site-packages/lvmdbusd/__pycache__/utils.cpython-39.opt-1.pyc -> /dev/null
    removing broken symbolic link /usr/lib/python3.9/site-packages/lvmdbusd/__pycache__/lvmdb.cpython-39.pyc -> /dev/null
    removing broken symbolic link /usr/lib/python3.9/site-packages/lvmdbusd/__pycache__/state.cpython-39.pyc -> /dev/null
    removing broken symbolic link /usr/lib/python3.9/site-packages/lvmdbusd/__pycache__/fetch.cpython-39.opt-1.pyc -> /dev/null

etc. etc. etc. Why we do this odd indirect thing instead of just wiping all `.pyc` files directly like we do `.pyo` files, I don't know.

bcl:
> etc. etc. etc. Why we do this odd indirect thing instead of just wiping all `.pyc` files directly like we do `.pyo` files, I don't know.

Well, my guess would be that it's by accident. Since /dev/ isn't populated the attempt to symlink to /dev/null ends up making broken symlinks :)

The reason for linking to /dev/null was, IIRC, to keep python from trying to compile and write to the filesystem while running things...

AdamWill:
ah, I see, that makes sense. So we should keep `/dev/null`, I guess...

bcl:
> ah, I see, that makes sense. So we should keep `/dev/null`, I guess...

Yeah, it should move after the broken symlink cleanup though. I ran a build with that on Friday and it worked fine. Not totally sure there's much benefit to it, but it should save on a few writes to the overlay.

hroncok:
When Python sees a symbolic link, it will happily override it with the bytecode cache:

```
[tmp]$ touch module.py
[tmp]$ python3 -m module
[tmp]$ ls -l __pycache__/
.rw-rw-r--@ 121 churchyard 14 Sep 18:26 module.cpython-38.pyc
[tmp]$ rm __pycache__/module.cpython-38.pyc 
[tmp]$ ln -sf /dev/null __pycache__/module.cpython-38.pyc
[tmp]$ ls -l __pycache__/
lrwxrwxrwx@ 9 churchyard 14 Sep 18:27 module.cpython-38.pyc -> /dev/null
[tmp]$ python3 -m module
[tmp]$ ls -l __pycache__/
.rw-rw-r--@ 121 churchyard 14 Sep 18:27 module.cpython-38.pyc
```

So this is not going to work.

bcl:
> When Python sees a symbolic link, it will happily override it with the bytecode cache:

Thanks, ends up I didn't look at all of them when I tried that. See PR #1077 for my next iteration here.

bcl:
Closed by #1077

PR #1075, opened 2020-09-11
Drop firmware for Mellanox Spectrum

AdamWill:
Mellanox Spectrum devices are switches intended for data centers.
It is I guess feasible that someone might want to install Fedora
on one, but from the product pages and data sheets, I believe
they all have management interfaces that do not require this
firmware to work, and that's what you'd use if you needed a
network connection during OS deployment. The firmware is only
needed for the actual switched interfaces, and we don't need to
make those work during installation.


AdamWill:
@DanGoldbergMellanox - sorry for the random ping, but you seem to be a Mellanox-y person :) Can you confirm my understanding is accurate here? The effect of this change would be that the `mellanox/mlxsw_spectrum*` firmware files would not be included in the Fedora installer environment - the environment when you boot a Fedora netinst or DVD installer image. So whichever bits of the device need those firmware files will not work in that environment. Once the install is complete, the firmware files *will* be present on the installed system.

bcl (approved):
Makes sense to me (unless someone complains).

PR #1076, opened 2020-09-11
runtime-cleanup: big refresh of stale things

AdamWill:
I based this on the output of a recent installer image build:
https://kojipkgs.fedoraproject.org/compose/branched/Fedora-33-20200904.n.0/logs/x86_64/buildinstall-Everything-logs/pylorax.log
I looked at every runtime-cleanup related error there and tried
to make appropriate changes. In many cases this means just
removing a line that isn't needed any more because the package
in question just went away or is no longer pulled into the
installer environment. In other cases packages changed name or
files moved around, and I tried to make appropriate updates. In
a few cases files moved to another package but I wasn't sure
enough it would still be safe to remove them so I just left them
in place. Most of the changes here I'm pretty sure should be
safe, though there *could* be unforeseen fallout from e.g. fixing
the removals from procps to be removals from procps-ng - it's
been years since that package was renamed, so something *could*
have started using those binaries in the meantime. I did at least
check that anaconda itself does not.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -10,7 +10,7 @@ remove usr/share/i18n
 ## perl is needed by /usr/bin/rxe_cfg from libibverbs
 
 ## no sound support, thanks
-removepkg flac gstreamer-tools libsndfile pulseaudio* sound-theme-freedesktop
+removepkg flac-libs libsndfile pulseaudio* sound-theme-freedesktop
 ## we don't create new initramfs/bootloader conf inside anaconda
 ## (that happens inside the target system after we install dracut/grubby)
 removepkg dracut-network grubby anaconda-dracut
@@ -22,7 +22,7 @@ removefrom dracut --allbut /usr/lib/dracut/modules.d/30convertfs/convertfs.sh \
                   /usr/lib/systemd/* /usr/lib/dracut/modules.d/98dracut-systemd/*.service \
                   /usr/lib/dracut/dracut-initramfs-restore
 ## we don't run SELinux (not in enforcing, anyway)
-removepkg checkpolicy selinux-policy libselinux-utils
+removepkg selinux-policy libselinux-utils
 
 ## selinux checks for the /etc/selinux/config file's existance
 ## The removepkg above removes it, create an empty one. See rhbz#1243168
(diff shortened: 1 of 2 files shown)


AdamWill:
Note on the `madan-fonts` change: that package, AFAICT, contains fonts intended for Nepali. It is no longer installed by default in comps - it's only `optional` in the catch-all fonts group. I believe `lohit-devanagri-fonts` covers the same characters and is our current default for Nepali. I did check that you can select Nepali in the installer after this change and it appears to look fine, though I don't read Nepali at all.

AdamWill:
I ran an image built with these changes through openQA and it did fine. That doesn't cover everything that can happen in the installer environment at all, but it's something.

bcl (approved):
Looks ok to me, thanks! Let's also see what @poncovka says on monday

poncovka (commented):
Looks good to me. Thank you!

sharkcz:
Seems this broke s390x and ppc64le
````
verifying the installroot
2020-09-25 05:52:19,897: libtic.so.6, needed by /usr/bin/tic, not found
libtic.so.6, needed by /usr/bin/tic, not found
2020-09-25 05:52:19,898: libtic.so.6, needed by /usr/bin/captoinfo, not found
libtic.so.6, needed by /usr/bin/captoinfo, not found
2020-09-25 05:52:19,900: libtic.so.6, needed by /usr/bin/toe, not found
libtic.so.6, needed by /usr/bin/toe, not found
2020-09-25 05:52:19,901: libtic.so.6, needed by /usr/bin/infotocap, not found
libtic.so.6, needed by /usr/bin/infotocap, not found
2020-09-25 05:52:19,902: libtic.so.6, needed by /usr/bin/infocmp, not found
libtic.so.6, needed by /usr/bin/infocmp, not found
````
`libtic` is removed, but not the tools (their removal has been dropped)

see eg. https://koji.fedoraproject.org/koji/taskinfo?taskID=52204714 or https://koji.fedoraproject.org/koji/taskinfo?taskID=52204709

sharkcz:
looks like s390x and ppc64le include also `ncurses` while x86 has only `ncurses-libs` + `ncurses-base`, looking why it is so ...

AdamWill:
Ah, I apologize for that. I did base these changes off an x86_64 log; I tried to check for arch-specific things where appropriate, but didn't notice that one. Indeed, ncurses only being included on those arches would explain the problem.

sharkcz:
No problem, it allowed us to review the dep chains. And thanks for the cleanup, it's always useful :-)

PR #1078, opened 2020-09-14
runtime-cleanup: big refresh of stale things

AdamWill:
I based this on the output of a recent installer image build:
https://kojipkgs.fedoraproject.org/compose/branched/Fedora-33-20200904.n.0/logs/x86_64/buildinstall-Everything-logs/pylorax.log
I looked at every runtime-cleanup related error there and tried
to make appropriate changes. In many cases this means just
removing a line that isn't needed any more because the package
in question just went away or is no longer pulled into the
installer environment. In other cases packages changed name or
files moved around, and I tried to make appropriate updates. In
a few cases files moved to another package but I wasn't sure
enough it would still be safe to remove them so I just left them
in place. Most of the changes here I'm pretty sure should be
safe, though there *could* be unforeseen fallout from e.g. fixing
the removals from procps to be removals from procps-ng - it's
been years since that package was renamed, so something *could*
have started using those binaries in the meantime. I did at least
check that anaconda itself does not.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -10,7 +10,7 @@ remove usr/share/i18n
 ## perl is needed by /usr/bin/rxe_cfg from libibverbs
 
 ## no sound support, thanks
-removepkg flac gstreamer-tools libsndfile pulseaudio* sound-theme-freedesktop
+removepkg flac-libs libsndfile pulseaudio* sound-theme-freedesktop
 ## we don't create new initramfs/bootloader conf inside anaconda
 ## (that happens inside the target system after we install dracut/grubby)
 removepkg dracut-network grubby anaconda-dracut
@@ -22,7 +22,7 @@ removefrom dracut --allbut /usr/lib/dracut/modules.d/30convertfs/convertfs.sh \
                   /usr/lib/systemd/* /usr/lib/dracut/modules.d/98dracut-systemd/*.service \
                   /usr/lib/dracut/dracut-initramfs-restore
 ## we don't run SELinux (not in enforcing, anyway)
-removepkg checkpolicy selinux-policy libselinux-utils
+removepkg selinux-policy libselinux-utils
 
 ## selinux checks for the /etc/selinux/config file's existance
 ## The removepkg above removes it, create an empty one. See rhbz#1243168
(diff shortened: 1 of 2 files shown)


AdamWill:
this is the F33 branch version of #1076 . It's exactly the same.

sharkcz:
please don't merge this, rawhide got broken due it

share/templates.d/99-generic/runtime-cleanup.tmpl
```diff
@@ -267,16 +234,10 @@ removefrom linux-firmware /usr/lib/firmware/as102*
...
 removefrom lldpad /etc/*
-removefrom lua /usr/bin/*
-removefrom madan-fonts /usr/share/fonts/madan/*
 removefrom mdadm /etc/*
 removefrom mesa-dri-drivers /usr/${libdir}/dri/*_video.so
-removefrom module-init-tools /etc/* /usr/sbin/insmod.static /usr/sbin/weak-modules
-removefrom mt-st /etc/* /usr/sbin/*
+removefrom mt-st /usr/sbin/*
 removefrom mtools /etc/*
-removefrom ncurses /usr/bin/captoinfo /usr/bin/infocmp /usr/bin/infotocap
```
sharkcz, line 277: this part breaks s390x/ppc64le image creation, because they install `ncurses` rpm, so it leaves the tools, but their libs are removed

bcl: @AdamWill We could either restore this removal (I'm not sure why this line was added actually) or we could wrap the `removefrom ncurses-libs /usr/${libdir}/libpanel.* /usr/${libdir}/libtic*` line in an arch check for ppc64le and s390x

sharkcz: I'm trying to figure out why s390x and ppc64le add `ncurses` to the package set, it could be a bug somewhere, hopefully will know more on Monday

sharkcz: so the dep chain is following - powerpc-utils/s390utils -> perl stuff -> `perl-Term-Cap` -> `/usr/bin/infocmp` -> `ncurses`

`infocmp` call in `perl-Term-Cap` is a fall-back path and even this path has another fall-back, so my suggestion is to remove all of `ncurses` in the cleanup template with
`removefrom ncurses /usr/bin/*`

unrelated to this issues is a dependency on `ncurses` in the `@Core` group in `comps`, I believe this should be `ncurses-base` instead (the terminal definitions are needed, not the compiler/decompiler tools)

bcl: Thanks for tracking that down. May as well remove all of the ncurses files, see PR #1085

bcl:
This is in F34 and I don't see a good reason to change F33 at this point, so closing it. Thanks!

PR #1079, opened 2020-09-14
Drop firmware for Mellanox Spectrum

AdamWill:
Mellanox Spectrum devices are switches intended for data centers.
It is I guess feasible that someone might want to install Fedora
on one, but from the product pages and data sheets, I believe
they all have management interfaces that do not require this
firmware to work, and that's what you'd use if you needed a
network connection during OS deployment. The firmware is only
needed for the actual switched interfaces, and we don't need to
make those work during installation.


AdamWill:
This is the F33 branch version of #1075 .

PR #1082, opened 2020-09-14
Drop dpaa2 firmware on non-aarch64 arches

AdamWill:
AFAICS, the devices that need these firmwares - various boards
built by NXP, https://www.nxp.com - are all aarch64. So we don't
need to carry these firmware files in the installer env for other
arches.


AdamWill:
This is F33 branch version of https://github.com/weldr/lorax/pull/1080 .

bcl (approved):

PR #1083, opened 2020-09-14
Fix broken single-item tuples in a few places

AdamWill:
When we stopped caring about ppc and ppc64, we changed several
instances of three-item tuples:

("ppc", "ppc64", "ppc64le")

into...this:

("ppc64le")

which is not a single item tuple, but just the string "ppc64le"
in some extraneous braces. It so happens that the right thing
still happened in all relevant cases , we think, but it's wrong.
There's no need to be using an iterator at all for a single
item, so just change them all to == "ppc64le" or != "ppc64le" as
appropriate.


AdamWill:
F33 branch version of https://github.com/weldr/lorax/pull/1081 .

bcl (approved):

PR #1085, opened 2020-09-28
runtime-cleanup: Remove ncurses package

bcl:


bcl:
Merged into master

sharkcz:
looks like I'm late, but it fixes the problem, tested both locally and in regular composes :-)

bcl:
> looks like I'm late, but it fixes the problem, tested both locally and in regular composes :-)

Thanks!

PR #1086, opened 2020-09-29
Switch to using mutter as the window manager

bcl:
mutter is the modern replacement for metacity, this installs the
required packages to use it for running Anaconda.

NOTE: This also requires a change to Anaconda to start mutter instead of
metacity.


bcl:
See https://github.com/rhinstaller/anaconda/pull/2874 for the Anaconda changes.

bcl:
Looks like this is making good progress in anaconda land using gnome-kiosk.

PR #1087, opened 2020-09-30
Remove lorax-composer

bcl:


bcl:
Submitted change to cockpit/bots as well - https://github.com/cockpit-project/bots/pull/1273

test/run
```diff
@@ -2,9 +2,7 @@
 # This is the expected entry point for Cockpit CI; will be called without
 # arguments but with an appropriate $TEST_OS, and optionally $TEST_SCENARIO
 
-export BACKEND="${BACKEND:-lorax-composer}"
-
-if [ "$BACKEND" == "osbuild-composer" ] || [ "$TEST_SCENARIO" == "osbuild-composer" ]; then
+if [ "$TEST_SCENARIO" == "osbuild-composer" ]; then
     rm -rf ./test/images/*
     export BACKEND="osbuild-composer"
     make BACKEND=osbuild-composer vm
```
atodorov, line 8: This shouldn't be necessary now that the backend is changing, no ?

atodorov (changes requested):
The removed files as far as they relate to tests look good but I see failure to create the qcow2 image in in test_compose_sanity with osbuild-composer backend.

test/run
bcl: At some point I want to turn on the lorax scenario, so yes I think it needs to check since the lorax vm won't need osbuild-composer installed.

atodorov: Sure, the lorax tests don't need osbuild-composer but you have the BACKEND variable set to default to osbuild-composer in many places, Makefile for example, which will propagate to vm.install and it will install osbuild-composer anyway. 

The section in `test/run` which is guarded by the modified if condition was used to do a conditional make vm because the proper value of the backend variable didn't get propagated. But `vm.install` still installs osbuild-composer or lorax-composer. Once you merge these changes this will start failing for the lorax tests b/c there will be no lorax-composer available anymore.

bcl: > Makefile for example, which will propagate to vm.install and it will install osbuild-composer anyway.

Ah, right! I guess for now it's not worth worrying about and I can revisit that when I get the lorax only tests ready to be run.

> Once you merge these changes this will start failing for the lorax tests b/c there will be no lorax-composer available anymore.

Which ones? The whole point of this PR is to remove them all. I'm only leaving the composer-cli tests running against osbuild-composer and eventually those will get moved elsewhere when composer-cli is rewritten.

bcl:
@atodorov  do you know where this comes from? "fedora-32 Expected — Waiting for status to be reported "
I submitted a PR to the cockpit bots to only run the fedora-32/osbuild-composer test, and I don't see any place to tell GH what tests to expect so I'm not sure what's going on there.

atodorov:
> @atodorov do you know where this comes from? "fedora-32 Expected — Waiting for status to be reported "

That's probably on the branch protection settings page for the repo but I don't have access to it. GitHub will wait for required checks even if they are never reported.

bcl:
> That's probably on the branch protection settings page for the repo but I don't have access to it. GitHub will wait for required checks even if they are never reported.

found it, thanks!

test/run
atodorov: > Once you merge these changes this will start failing for the lorax tests b/c there will be no lorax-composer available anymore.

I'm wrong here. Ignore this.

atodorov (approved):

test/run
atodorov: Having this extra section still doesn't make any sense b/c `make vm` will do exactly the same, b/c the BACKEND variable now defaults to osbuild-composer. So you always get osbuild-composer installed even if you don't use it. But this doesn't block merging of this PR and proceeding forward with the cleanup.

bcl: Right, I'll revisit the vm setup when I'm ready to enable plain lorax testing. Thanks for the review!

PR #1091, opened 2020-10-28
Switch to using upstream mk-s390image for s390 cdboot.img creation

bcl:
mk-s390-cdboot has stopped working because the kernel outgrew the
hard-coded offset it used when creating cdboot.img. IBM now has a script
in s390utils that can do the same thing so use the upstream script
instead.

This drops mk-s390-cdboot script, switches the s390 templates to use
mk-s390image from s390utils.

It adds @ROOT@ to cdboot.prm, and sets inst.stage2 so that the installer
image will be found when booting the iso.

Resolves: rhbz#1891778


share/templates.d/99-generic/live/s390.tmpl
```diff
@@ -6,6 +6,7 @@ KERNELDIR=BOOTDIR
 INITRD_ADDRESS="0x02000000"
 LIVEDIR="LiveOS"
 LORAXDIR="usr/share/lorax/"
+MKS390IMAGE="/usr/share/s390-tools/netboot/mk-s390image"
```
sharkcz, line 9: This is now `/usr/bin/mk-s390image` since `s390utils-2.15.0-2.fc34`

bcl: Ok, weird. I thought I pushed that with the requires change. good catch.

PR #1094, opened 2020-11-11
[RHEL7] Fix 'local variable 'rc' referenced before assignment'

kierdavis:
This happens when --make-tar is selected and the virtual install
fails in such a way that the disk image contains no acceptable
partitions.

I've submitted this fix against the RHEL7 branch because that's the version I was using when I hit this, but it might affect other branches too. Shall I open PRs for all affected branches, or do you have a better procedure for this?


bcl:
Thanks, RHEL 7 is no longer receiving updates, so PRs for master and RHEL 8 would be the right way to do it.

PR #1097, opened 2020-11-18
Do not use '--loglevel' option when running Anaconda

vojtechtrefny:
This option has been removed, see https://github.com/rhinstaller/anaconda/pull/2864


poncovka (commented):
Looks good to me.

PR #1099, opened 2020-11-30
Don't remove libldap_r libraries during runtime-cleanup.tmpl

droideck:


bcl (approved):
Thanks! Looks good to me.

PR #1102, opened 2020-12-29
Update creator.py

wangchuanguo:
I failed to --make-ostree-live , and I found the reason is that Dir 'sysroot' does not exist.
My environment: 
OS: Centos 8.2.2004   
lorax: 34.6-1


bcl:
> My environment:
> OS: Centos 8.2.2004
> lorax: 34.6-1

You should be using the version of lorax/livemedia-creator provided by CentOS. There are differences between distributions and releases that can cause subtle (and not-so-subtle) problems like this.

PR #1105, opened 2021-01-07
test: Fix vm.install for non-LVM cloud images

martinpitt:
Some of our images, like centos-8-stream, are already built from the
official cloud images instead of virt-install with LVM. More images are
going to do that soon [1][2], so fix vm.install to only do the LVM grow
steps if the image actually uses LVM.

Also adjust the comment, as commit 6ddaa5e0dd0b0ce9db fixed this for
RHEL images.

[1] https://github.com/cockpit-project/bots/pull/1518
[2] https://github.com/cockpit-project/bots/pull/1527

### DIFF
--- a/test/vm.install
+++ b/test/vm.install
@@ -31,12 +31,15 @@ sed -i "s/download.fedoraproject/dl.fedoraproject/" /etc/yum.repos.d/*.repo
 # Example: `#baseurl=http://download.example/pub/fedora/linux/updates/$releasever/Everything/$basearch/`
 sed -i "s/download.example/dl.fedoraproject.org/" /etc/yum.repos.d/*.repo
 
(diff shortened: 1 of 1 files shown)


martinpitt:
This blocks https://github.com/cockpit-project/bots/pull/1527 . It'll need to be backported to rhel8-branch as well, together with commit 6ddaa5e0dd0b0ce9d.

martinpitt:
That doesn't quite work right yet, see https://github.com/weldr/lorax/pull/1106

martinpitt:
This is not sufficient yet to run tests on a cloud image, see PR #1106. But it is a necessary prerequisite, so marking as ready for review.

larskarlitski (approved):
Makes sense to me. Thanks!

PR #1107, opened 2021-01-18
templates: Copy missing files into ISO filesystem

dpward:
Restore the Fedora (or generic) license files, which have been missing from image builds since Fedora 30.

Copy the `livecd-iso-to-disk` script (if it is present) into the LiveOS/ directory. Currently, a post-install script in fedora-live-base.ks [modifies a Lorax template](https://pagure.io/fedora-kickstarts/blob/97dd0f7e4d9f5b7005ede9369df6784a0fc01a46/f/fedora-live-base.ks#_338-346) to do this; move that directly into Lorax.

Please backport to stable Fedora branches (applies cleanly).

### DIFF
--- a/share/templates.d/99-generic/aarch64.tmpl
+++ b/share/templates.d/99-generic/aarch64.tmpl
@@ -68,7 +68,7 @@ mkdir ${KERNELDIR}
 %>
 
 # Add the license files
-%for f in glob("/usr/share/licenses/*-release/*"):
+%for f in glob("usr/share/licenses/*-release-common/*"):
     install ${f} ${f|basename}
     <% filegraft += " {0}={1}/{0}".format(basename(f), outroot) %>
 %endfor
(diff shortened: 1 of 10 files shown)


bcl (changes requested):
Looks ok, but could you add the leading / back to the glob paths?  In practice they don't matter but it looks nicer.

dpward:
Actually, dropping the / matters. An image can be built with generic branding, i.e., with the generic-release package instead of the fedora-release package in the kickstart file.
However, the build may still occur in a mock chroot with the fedora-release package installed. This is a reason the license files need to be copied from the installroot, not from the mock chroot.

Also, that was the only place I could find in the templates that had a leading / in the path. The other globs look like this:
`%if exists("boot/efi/EFI/*/gcdx64.efi"):`

bcl:
> However, the build may still occur in a mock chroot where the fedora-release package is installed. This is a reason the license files need to be copied from the installroot (no leading / in the path), rather than the mock chroot (leading / in the path).

exists an glob always check inroot, not the host, eg.

`glob=lambda g: list(rglob(g, root=inroot)))`

is what's passed into the template.

BUT on the other hand, you're right everywhere else it's used without the leading / so I retract my request, let's make it consistent :)

Thanks!

bcl (approved):
Looks good!

PR #1108, opened 2021-01-20
Improve lmc support for other arches

bcl:


hadess:
Looks good to me. I'll try to spend some time on this soon, see if I can bring it the last few steps.

bcl:
Thanks, I'd like to get aarch64 working before I merge it. From looking at the Fedora aarch64 test docs it looks like it needs to use UEFI to boot at all, but without a graphics console I couldn't figure out what was going wrong. We may need to add or change some of the settings for it, if that's the case they should be put into the QEMU dict to make it easier.

bcl:
Some ideas to borrow from @thozza - https://github.com/thozza/osbuild-composer/blob/generate-all-test-cases/tools/test-case-generators/generate-all-test-cases#L455

PR #1110, opened 2021-01-28
tests: Include the fedora-updates repo when testing boot.iso building

bcl:
Might fix https://github.com/cockpit-project/bots/pull/1595


bcl:
@martinpitt Is there a good way to test this with https://github.com/cockpit-project/bots/pull/1595? Or should I just merge it?

martinpitt:
@bcl: We can run project PRs against a bots PR, but it's not necessary here -- the bots PR does not change any logic, just automatic test triggers. So good to land this, and we can retry the test in the bots PR. Note that here you tested fedora-32/osbuild-composer while in the bots PR you enabled fedora-32/lorax, i.e. a different scenario. That may have been a typo, or intended.

Either way, this PR makes sense of course. Thanks!

PR #1111, opened 2021-01-29
Use image dependencies metapackage

VladimirSlavik:
The difference between the anaconda-install-{env,img}-deps packages is how they treat dependencies. The -env package leaves some dependencies as weak to allow less featureful builds. The -img package hard-requires everything Anaconda could potentially use and ensures everything works.

For boot.iso, the latter is preferable.

Depends on https://github.com/rhinstaller/anaconda/pull/3126

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -2,7 +2,7 @@
 <%page args="basearch, product"/>
 
 ## anaconda package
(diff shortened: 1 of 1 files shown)


bcl (approved):
Thanks! I like that!

bcl:
New package is now available in rawhide, thanks!

VladimirSlavik:
Thank you too!

PR #1113, opened 2021-02-13
Use inst.rescue to trigger rescue mode

AdamWill:
anaconda in F34 and Rawhide recently stopped accepting params
without the inst. prefix, so 'rescue' does nothing except print
a warning now. We need to use `inst.rescue`. This has worked for
quite a long time so will be OK at least on all Fedoras and RHEL
8, not sure about RHEL 7.


AdamWill:
We need this on Rawhide and F34. Doesn't seem to be an f34 branch yet.

AdamWill:
test failure is rawhide package signature problems, nothing to do with the PR.

bcl (approved):
Looks good, thanks!

PR #1115, opened 2021-02-16
composer-cli: Add support for start-ostree --url URL

bcl:
Some ostree builds, like edge, require passing a url to it along with
the ref and parent arguments. This adds an optional --url argument to
the 'composer-cli compose start-ostree' command.

Related: rhbz#1929381


atodorov (approved):

bcl:
Merged to rhel8-branch

PR #1123, opened 2021-03-17
Change khmeros-base-fonts to khmer-os-system-fonts

pnemade:
This font got renamed last year.

Signed-off-by: Parag Nemade <pnemade@fedoraproject.org>


bcl (approved):
Thanks!

PR #1129, opened 2021-03-24
test: Fix URL to bots testmap

martinpitt:
It was moved in https://github.com/cockpit-project/bots/commit/7a3c71968d568


bcl (approved):
Thanks!

PR #1130, opened 2021-03-24
[rhel8] test: Fix URL to bots testmap, fix osbuild-composer tests

martinpitt:
It was moved in https://github.com/cockpit-project/bots/commit/7a3c71968d568

----

Backport of PR #1129. This is mostly an excuse to test rhel-8-4/osbuild-composer on the current rhel-8-4 image -- it fails in https://github.com/cockpit-project/bots/pull/1836 and I'd like to see if it's due to the new image or (more probably) due to changes in the compose.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -8,7 +8,7 @@ images and tools to manipulate and start virtual machines from them.
 
 Each test is run on a new instance of a virtual machine.
 Branch/test scenario matrix is configured in
-[testmap.py](https://github.com/cockpit-project/bots/blob/master/task/testmap.py).
+[testmap.py](https://github.com/cockpit-project/bots/blob/master/lib/testmap.py).
 
 ## Dependencies
 
--- a/tests/cli/test_compose_sanity.sh
+++ b/tests/cli/test_compose_sanity.sh
@@ -31,7 +31,7 @@ rlJournalStart
             TYPE_EXT4=""
             TYPE_PARTITIONED_DISK=""
             TYPE_TAR="tar"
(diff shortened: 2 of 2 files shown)


martinpitt:
Yep, same test_compose_sanity() failure. The log artifact isn't particularly insightful, maybe this one?
```
[/usr/lib/tmpfiles.d/journal-nocow.conf:26] Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
Failed to create file /sys/fs/selinux/checkreqprot: Read-only file system
```

atodorov:
This is the actual failure:
```
:: [ 01:52:07 ] :: [   FAIL   ] :: lists all supported types (Assert: 'ami openstack qcow2 rhel-edge-commit rhel-edge-container rhel-edge-installer tar vhd vmdk' should equal 'ami openstack qcow2 rhel-edge-commit tar vhd vmdk')
```

@bcl looks like we need a refresh of the test case, doesn't it? 

@martinpitt see `tests/cli/test_compose_sanity.sh` lines 25-50, especially line 34. I think you need to append the 2 extra rhel-edge compose types there.

bcl:
> @bcl looks like we need a refresh of the test case, doesn't it?

yeah, this runs against osbuild-composer now and they have added new types.

martinpitt:
Added an extra commit which does that. (I don't really know what I'm doing :grin: )

atodorov (approved):

PR #1132, opened 2021-04-02
Drop trying to install reiserfs-utils

nirik:
reiserfs-utils was retired in Fedora, so we don't want to try and
install it anymore.

Signed-off-by: Kevin Fenzi <kevin@scrye.com>


bcl (approved):
Test failures are unrelated to the change. ACK.

PR #1133, opened 2021-04-08
lorax-composer: Catch dnf unknown group error

bcl:
dnf-4.2.23-4 is raising a ValueError if the group doesn't exist, catch
that and add it to the error list when depsolving. Includes tests for
depsolve and freeze.

Resolves: rhbz#1943206


atodorov (approved):

bcl:
This fix is only for RHEL 8.3, 8.4 will need a different fix (hopefully in dnf).

PR #1138, opened 2021-04-23
Replace metacity with gnome-kiosk

rvykydal:
When replacing metacity with gnome-kiosk I started to build on @bcl's Draft for replacing metacity with mutter:
https://github.com/weldr/lorax/pull/1086

With vm dependency moved into anaconda in this PR:
https://github.com/weldr/lorax/commit/8602e653ff999942199faf9100e0a71604980469
the switch itself will be in anaconda dependencies (after the switch I will remove also https://github.com/weldr/lorax/blob/2762cf95da6bee708fe586e5b2e314c935a731c9/share/templates.d/99-generic/runtime-cleanup.tmpl#L110 in a PR).

I am not sure if for gnome-kiosk we need also the part in this PR (compared to the mutter PR) to not remove /usr/bin/* from glib2 anymore, I will ask Ray.
The files that used to be removed:
```
-rwxr-xr-x. 1 rvykydal rvykydal 23832 Apr  8 17:00 gapplication
-rwxr-xr-x. 1 rvykydal rvykydal 52800 Apr  8 17:00 gdbus
-rwxr-xr-x. 1 rvykydal rvykydal 98168 Apr  8 17:00 gio
-rwxr-xr-x. 1 rvykydal rvykydal 15568 Apr  8 17:00 gio-querymodules-64
-rwxr-xr-x. 1 rvykydal rvykydal 52776 Apr  8 17:00 glib-compile-schemas
-rwxr-xr-x. 1 rvykydal rvykydal 32208 Apr  8 17:00 gsettings
```

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -172,7 +172,7 @@ removefrom gdb-headless /usr/share/* /etc/gdbinit*
 removefrom gdisk /usr/share/*
 removefrom gdk-pixbuf2 /usr/share/locale*
 removefrom gfs2-utils /usr/sbin/*
-removefrom glib2 /usr/bin/* /usr/share/locale/*
+removefrom glib2 /usr/share/locale/*
 removefrom glibc /etc/gai.conf /etc/rpc
 removefrom glibc /${libdir}/libBrokenLocale*
 removefrom glibc /${libdir}/libSegFault* /${libdir}/libanl*


rvykydal:
@halfline Ray, to you think we need this patch for gnome-kiosk in installer environment ?

halfline:
> @halfline Ray, to you think we need this patch for gnome-kiosk in installer environment ?

gnome-kiosk does not directly call any of those binaries at runtime. So if you want to remove them to make the installer image more compact, that's probably fine.

rvykydal:
Thank you @halfline.
@bcl what do you think? I am not able to look up what was the exact reason for not removing the binaries.

bcl:
No idea why they were left in place, there may not have been a specific reason.

Is gnome-kiosk going to drag in a bunch of new packages? The reason I was trying to get mutter working was because the full gnome desktop brings in way too much for a 'small' boot.iso

Have you done a test run with all of this, and if so how much bigger (or smaller) is it from the current boot.iso?

halfline:
> Is gnome-kiosk going to drag in a bunch of new packages? The reason I was trying to get mutter working was because the full gnome desktop brings in way too much for a 'small' boot.iso

gnome-kiosk is a thin wrapper around mutter that adds a keyboard layout switching api. It doesn't bring in any extra dependencies over mutter other than itself.

rvykydal:
I am closing this PR, it seems there is really no reason to keep the binaries.

PR #1144, opened 2021-05-11
Add Requires: lorax-templates-rhel on ELN

sgallagher:
The ELN project provides the %{rhel} macro, so we can conditionalize
on it when building for that target. This will also make the specfile
simpler for future versions of CentOS Stream and RHEL.

Also update the comment describing the upstream location.

Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -1,10 +1,10 @@
-# NOTE: This specfile is generated from upstream at https://github.com/rhinstaller/lorax
+# NOTE: This specfile is generated from upstream at https://github.com/weldr/lorax
 # NOTE: Please submit changes as a pull request
 %define debug_package %{nil}
 
 Name:           lorax
 Version:        35.2
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Tool for creating the anaconda install images
 
 License:        GPLv2+
@@ -20,6 +20,9 @@ BuildRequires:  make
 BuildRequires:  systemd-rpm-macros
 
 Requires:       lorax-templates
+%if 0%{?rhel} >= 9
+Requires:       lorax-templates-rhel
+%endif
 
 Requires:       cpio
 Requires:       device-mapper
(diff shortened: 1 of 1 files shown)


sgallagher:
This is exactly how it is done in CentOS Stream 9... 

https://gitlab.com/redhat/centos-stream/rpms/lorax/-/blob/c9s/lorax.spec#L23

hanzz:
@sgallagher , Well, we need to get lorax-templates-rhel into root while generating the ELN task here:

https://koji.fedoraproject.org/koji/taskinfo?taskID=67865501

One option was to pull it in by "lorax", but if that's not right way, we can find out some other way how to do that. For example:

- Add it to `build` group so it is in ELN buildroot everytime. But this would also include it for RPM builds, but I believe it's not big issue.
- Do PR for Pungi so it can include extra packages for runroot tasks. So far the package names are hardcoded there and there is no configuration option to extend/change it. This would probably have to happen here: https://pagure.io/pungi/blob/master/f/pungi/runroot.py#_254.

bcl:
>     * Do PR for Pungi so it can include extra packages for runroot tasks. So far the package names are hardcoded there and there is no configuration option to extend/change it. This would probably have to happen here: https://pagure.io/pungi/blob/master/f/pungi/runroot.py#_254.

That seems like the right way to me. The idea is that lorax is just the tool, and whatever (pungi in this case) is running it decides whether to use the generic templates or to replace them with something else.

sgallagher:
> > Do PR for Pungi so it can include extra packages for runroot tasks. So far the package names are hardcoded there and there is no configuration option to extend/change it. This would probably have to happen here: https://pagure.io/pungi/blob/master/f/pungi/runroot.py#_254.
> 
> That seems like the right way to me. The idea is that lorax is just the tool, and whatever (pungi in this case) is running it decides whether to use the generic templates or to replace them with something else.

We decided to go the route of adding it to the eln-build `@build` group:
```
$ koji list-groups eln-build build
build  [f35]
  bash: None, mandatory  [f35]
  bzip2: None, mandatory  [f35]
  coreutils: None, mandatory  [f35]
  cpio: None, mandatory  [f35]
  diffutils: None, mandatory  [f35]
  fedora-release: None, mandatory  [f35]
  findutils: None, mandatory  [f35]
  gawk: None, mandatory  [f35]
  glibc-minimal-langpack: None, mandatory  [f35]
  grep: None, mandatory  [f35]
  gzip: None, mandatory  [f35]
  info: None, mandatory  [f35]
  lorax-templates-rhel: None, mandatory  [eln-build]
  make: None, mandatory  [f35]
  patch: None, mandatory  [f35]
  redhat-rpm-config: None, mandatory  [f35]
  rpm-build: None, mandatory  [f35]
  sed: None, mandatory  [f35]
  shadow-utils: None, mandatory  [f35]
  tar: None, mandatory  [f35]
  unzip: None, mandatory  [f35]
  util-linux: None, mandatory  [f35]
  which: None, mandatory  [f35]
  xz: None, mandatory  [f35]
```

PR #1148, opened 2021-05-19
Replace metacity with gnome-kiosk

bcl:
Related: rhbz#1961099

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -105,7 +105,6 @@ removefrom gtk2 /usr/${libdir}/gtk-2.0/*/{engines,printbackends}/*
 removefrom gtk2 /usr/share/themes/*
 removefrom gtk3 /usr/${libdir}/gtk-3.0/*/printbackends/*
 removefrom gtk3 /usr/share/themes/*
-removefrom metacity --allbut /usr/bin/* /usr/${libdir}/*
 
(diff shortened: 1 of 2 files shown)


share/templates.d/99-generic/runtime-cleanup.tmpl
```diff
@@ -343,9 +342,6 @@ removepkg geoclue2
 removepkg cdparanoia-libs opus libtheora libvisual flac-libs gsm avahi-glib avahi-libs \
           ModemManager-glib
 
-## metacity requires libvorbis and libvorbisfile, but enc/dec are no longer needed
-removefrom libvorbis --allbut /usr/${libdir}/libvorbisfile.* /usr/${libdir}/libvorbis.*
-
```
rvykydal, line 348: I am not sure about this change.
In the iso I built with metacity replaced by gnome-kiosk the libvorbis is still pulled in so removing the unneeded enc/dec is still valid and I'd keep these lines. Maybe we could update the comment with the information about what pulls the library in, but I don't insist (and I don't know).

rvykydal (approved):
Other than the comment looks good to me.

rvykydal (changes requested):
Err, removing the approval before we clear it up with the libvorbis.

rvykydal:
I think the dependency chain here is:
gnome-kiosk -> mutter -> libcanberra -> libvorbis

rvykydal (approved):
Looks good to me, thank you!

PR #1149, opened 2021-05-21
fix: POC for a sane dracut chroot environment

haraldh:
`bash` and various other utilities only work with a sane chroot
environment. At least `bash` needs `/dev/fd` to function properly.

Take this as a POC patch to make things work.

See https://bugzilla.redhat.com/show_bug.cgi?id=1962975#c2

With this patch:
```console
# lsinitrd ./isolinux/initrd.img|fgrep ca-bundle
-r--r--r--   1 root     root       216090 May 18 10:36 etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
lrwxrwxrwx   1 root     root           46 May 18 10:36 etc/pki/tls/certs/ca-bundle.crt -> ../../ca-trust/extracted/pem/tls-ca-bundle.pem
```

### DIFF
--- a/src/pylorax/treebuilder.py
+++ b/src/pylorax/treebuilder.py
@@ -328,14 +328,17 @@ class TreeBuilder(object):
                 # Construct an initrd from the kernel name
                 outfile = kernel.path.replace("vmlinuz-", "initrd-") + ".img"
             logger.info("rebuilding %s", outfile)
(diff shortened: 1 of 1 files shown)


bcl:
Thanks for tracking that down, I've expanded on this a bit in #1151 which should make it easier to add future requirements.

PR #1150, opened 2021-05-24
Add prefixdevname to Anaconda initramfs

rvykydal:
Related: rhbz#1958173


bcl:
Ah, shoot! Thanks for catching that.

PR #1151, opened 2021-05-24
Add a context manager for dracut

bcl:
dracut needs a reasonable chroot environment in order to run correctly.
This adds the DracutChroot context manager that sets up and tears down
mounts inside the root directory tree.

Switch to using it in creator.rebuild_initrds_for_live() and
TreeBuilder.rebuild_initrds()


bcl:
This solves #1149

bcl:
Tests only failed due to coverage, which can't be increased for this change because it involves mounting things and running dracut in a chroot. Tested manually on rawhide host with `lorax` and `livemedia-create --make-pxe-live`

PR #1152, opened 2021-05-24
replace nomodeset with inst.text

bcl:


kparal:
> nomodeset no longer does anything helpful

What do you mean? It allows people to boot with a basic video driver. I myself used it a couple of times in the past, with old GPUs/problematic GPU drivers. Has something changed recently?

bcl:
> Has something changed recently?

Yes, I can't seem to track down the discussion at the moment but the gist of it was that `nomodeset` is no longer helpful. You can still pass `inst.xdriver` if there is a specific driver needed.

kparal:
The [Anaconda User Interface Basic Video Driver](https://fedoraproject.org/wiki/QA:Testcase_Anaconda_User_Interface_Basic_Video_Driver) testcase passed for Fedora 34 GA on multiple machines in multiple variations (BIOS, UEFI). So at that point everything was working as expected. I tested it quickly now with `Fedora-Workstation-Live-x86_64-Rawhide-20210605.n.0.iso` just by booting into the Live image in standard and basic graphics mode, and I still see it working correctly. In the standard mode, graphics are accelerated, GNOME is running on Wayland, `xdriinfo` says `iris`, `glxinfo | grep render` says `Intel UHD`. In the basic graphics mode, graphics are choppy, GNOME is running on X11, `xdriinfo` says `swrast`, `glxinfo` says `llvmpipe`, and `journalctl` says that the `fbdev` driver is being used (I used the UEFI mode). So I don't understand why `nomodeset` would no longer be helpful. Can you elaborate / find that discussion you referred to? Everything seems to be working as expected in my testing.

kparal:
OK, I tested in BIOS mode as well, the same Rawhide image. Works exactly as described in the test case, i.e. launches X11 and uses the `vesa` driver. Seems fine.

AdamWill:
I'm +1 @kparal here. It still Does Something which is sometimes useful. I'm fine with removing it in favor of a better alternative, but not removing it in favor of nothing.

bcl:
Ok, looks like we have some wires crossed so I'm closing this for now.
Thanks for the double-check :)

bcl:
I finally found where the discussion was: https://bugzilla.redhat.com/show_bug.cgi?id=1961092

AdamWill:
Followed up there.

kparal:
I'll follow up here, just to make things more complicated :-) (and because it feels better to discuss Fedora here than under a RHEL bug).

Brian asked:
> Is this something that's different between the Fedora and RHEL kernels maybe?

I believe the reason is in Ajax's [reply](https://bugzilla.redhat.com/show_bug.cgi?id=1961092#c5):
> At root, the issue here is there is no longer a vesa driver in RHEL9.

We clearly still ship the vesa driver in Fedora, that's why it works for Fedora.

Another difference is that we have a Fedora Workstation Live image, where the basic graphics mode is the most helpful. Replacing `nomodeset` with `inst.text` is of course useless in there, Live images don't support  a text installation.

And as Adam noted in bugzilla, the userbase, the expectations and the target hardware is surely very different between RHEL and Fedora.

Of course replacing "basic graphics mode" with a better term might be a good idea, thumbs up to that.

Also, if there's a consensus (probably including FESCo) that "basic graphics mode" should go away, or if there are good technical reasons for it, I don't actually object. I'm just saying that the original assumption "nomodeset no longer does anything helpful" seems untrue to me, unless something changed very recently.

PR #1153, opened 2021-06-09
Drop retired icfg

keszybz:
Rawhide compose failed because icfg was retired yesterday:
https://src.fedoraproject.org/rpms/icfg/c/64ef393a109dd468b1b6950bfef70b2c45b5b903?branch=rawhide
https://pagure.io/releng/failed-composes/issue/2557

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -173,7 +173,7 @@ installpkg python3-pyatspi
 installpkg nano nano-default-editor
 installpkg vim-minimal strace lsof dump xz less
 installpkg wget rsync bind-utils ftp mtr vconfig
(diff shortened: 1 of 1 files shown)


keszybz:
From #anaconda irc channel:

<jkonecny> zbyszek, I wonder why  icfg is in the install environment if it's there not because of Anaconda
<jkonecny> zbyszek, I think it is just a left-over https://bugzilla.redhat.com/show_bug.cgi?id=771733
<jkonecny> it should be pretty safe for removal
<jkonecny> it was added because someone asked but now users shouldn't use package like that to edit ifcfg-* files
<jkonecny> also not sure if that would even work because I don't know if the ifcfg-* files are compatible with icfg-* files

bcl (approved):
Thanks!

PR #1157, opened 2021-06-22
Use wildcard for grub2 modules in livemedia.ks example

bcl:
This will make it easier to use the kickstart on different arches.

Resolves: rhbz#1973530


bcl:
Merged

PR #1158, opened 2021-06-23
runtime-install: Remove gfs2-utils

bcl:
It is not needed on the installer media. Also remove references from
runtime-cleanup.tmpl

Related: rhbz#1975378


VladimirSlavik (approved):
Thank you!

PR #1170, opened 2021-09-07
Increase root partition of rhel-livemedia.ks to 8000MB

bcl:
Resolves: rhbz#2001942

### DIFF
--- a/docs/rhel-livemedia.ks
+++ b/docs/rhel-livemedia.ks
@@ -33,7 +33,7 @@ bootloader --location=none
 clearpart --all --initlabel
 rootpw rootme
 # Disk partitioning information
-part / --size=6656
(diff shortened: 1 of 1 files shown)


jstodola:
I would prefer if the size was increased more than ~350 MB. We already need like 120 MB (based on bug 2001942), so I'm afraid that sooner or later we will hit this problem again due to the increasing sizes of packages.
Is there any reason for not increasing the size to let's say 8000 MB, which would hopefully be enough for the rest of RHEL-8 release.

bcl:
> Is there any reason for not increasing the size to let's say 8000 MB,

Nope. I just picked the next round looking number :)

jstodola (approved):
It looks good to me, thanks.

PR #1175, opened 2021-10-06
Firmware cleanup 2021

AdamWill:
This PR includes a few commits to drop various probably-unnecessary firmware files in runtime-cleanup, to save about 30M on network install images. This will help with them being oversize for F35 and Rawhide currently.

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -17,12 +17,16 @@ installpkg grubby
 %if basearch != "s390x":
     ## skip the firmware for sound, video, and scanners, none of which will
     ## do much good for the installer. Also skip uhd-firmware which is not
-    ## even a kernel firmware package.
+    ## even a kernel firmware package. liquidio and netronome firmwares are
+    ## for enterprise switch devices, netinst deployment does not work on
+    ## these so there is no point shipping them - see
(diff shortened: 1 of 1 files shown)


AdamWill:
@nirik @bcl

AdamWill:
thanks. they're pretty tiny, though, they barely add up to a megabyte, so might not be worth bothering with!

AdamWill:
Thanks! As I'm intending these changes to help us get the F35 images under size, can you do a new release/build/update for F35? I can just backport the patches if you'd prefer, just let me know :)

bcl:
> Thanks! As I'm intending these changes to help us get the F35 images under size, can you do a new release/build/update for F35? I can just backport the patches if you'd prefer, just let me know :)

That probably needs a bug and an exception doesn't it? Since we're past the final freeze date.

bcl:
> FYI, I have a private TODO to re-examine the packages on boot.iso. I looked at that a few months ago and there are just so many things that make zero sense at first glance that it's hard to believe. I think the buildup to F34/RHEL9 brought a lot of stuff through some kind of churn. However I didn't check how many of these are cleaned up by the templates.

Inertia is a better description :) I tend to make minimal changes once something has been added since it's hard to know what really depends on it. IIRC the majority of the growth over time has been the firmware blobs but I'm sure there's room for improvement in other areas.

PR #1179, opened 2021-10-11
Handle all possible dnf group_install errors

bcl:
dnf has changed what it will raise when trying to group_install a
nonexistent group, this adds handling for all of them.

Related: rhbz#1947958


lukash:
LGTM

PR #1185, opened 2021-10-30
Update runtime-install/cleanup for Marvell Prestera fw split

AdamWill:
Marvell Prestera firmware has been split into its own subpackage,
so instead of stripping the files from linux-firmware, exclude
the package from the globed install command.


AdamWill:
Note, @nullr0ute sent the same change to F34 and F35, so it'd be technically correct to backport it to those branches too, just for anyone building updated installer images I guess.

nullr0ute:
LGTM

bcl (approved):
Thanks!

PR #1197, opened 2021-12-07
Fix bugs and make improvements in mkksiso

bcl:


src/sbin/mkksiso
```diff
@@ -165,7 +165,7 @@ class Mkisofs_aarch64(MakeISOTool):
 
     def run(self, tmpdir, grafts, volume_name, output_iso, efimode=NO_EFI):
         cmd = ["mkisofs", "-o", output_iso, "-R", "-J", "-V", volume_name,
-               "-T"]
+               "-joliet-long", "-T"]
```
VladimirSlavik, line 168: The man page states that "Oversized names get truncated." Will that result in a failure mode better than what the BZ states?

bcl: Good question... Actually it silently ignores anything longer than 253 bytes. So I guess I'll add a check for file length and raise an error.

bcl:
Merged into master.

PR #1201, opened 2021-12-13
runtime-cleanup: remove pipewire-related packages

AdamWill:
We aim to remove all sound support from the installer root, but
this was never updated for Pipewire. This just started causing
compose failures because of a dep chain from pipewire to lilv.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -10,7 +10,7 @@ remove usr/share/i18n
 ## perl is needed by /usr/bin/rxe_cfg from libibverbs
 
 ## no sound support, thanks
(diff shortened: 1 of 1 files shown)


AdamWill:
Note: we don't remove `pipewire*` because that includes `pipewire-libs`, and both `mutter` and `gnome-kiosk` are directly linked against libpipewire, so we can't take that out.

AdamWill:
I've tested that a lorax with this patch applied successfully builds an installer image (i.e. it doesn't choke on the dependency test) and the installer image boots successfully. openQA is [in the middle of](https://openqa.stg.fedoraproject.org/tests/1492183#live) testing it can actually install successfully.

bcl (approved):
Looks Good!

VladimirSlavik (commented):
Can't really approve, not enough data, but I had my sights on that for some time. Thanks!

PR #1202, opened 2021-12-13
cleanup: remove binaries from lilv

AdamWill:
Since we're leaving pipewire-libs in, it'll still pull in liblilv
from 0.3.41 onwards. Currently liblilv is in the same package as
some binaries which we don't need, and one actually requires the
removed libsndfile, so we need to trim it.


bcl (approved):
Looks good. I built a test boot.iso and installed using new the pipewire.

AdamWill:
This could potentially be reverted when https://bugzilla.redhat.com/show_bug.cgi?id=2032115 is addressed...

PR #1203, opened 2021-12-17
Add inst.rngd cmdline option

bcl:
rngd should no long be needed, recent kernels have enough entropy out of
the box. This leaves it in the boot.iso but only enables it if inst.rngd
is passed on the kernel cmdline.

Resolves: rhbz#2028720


bcl:
Merged to master.

PR #1204, opened 2021-12-21
Master image minimizer

bcl:


bcl:
Not normally, no. It is a script that came from livecd-creator originally.

PR #1205, opened 2022-01-12
Make the iso only bootable on UEFI

bcl:


Conan-Kudo:
You know, you don't _have_ to drop BIOS support for ISO, just drop isolinux and use GRUB for both.

rathann:
I think the consensus from the discussion at [F37 Change: Deprecate Legacy BIOS (System-Wide Change proposal)](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/K5YKCQU3YVCTMSBHLP4AOQWIE3AHWCKC/) is that it's too early to drop BIOS support.

bcl:
I wasn't planning on it, it has been replaced by #1226

PR #1213, opened 2022-02-03
cleanup: handle RPM database move to /usr

AdamWill:
In F36, the RPM database is being moved from /var/lib/rpm to
/usr/lib/sysimage/rpm:

https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr
https://bugzilla.redhat.com/show_bug.cgi?id=2042099

So we need to empty that location in runtime-cleanup. Seems fine
to just list both locations, for a while at least.


bcl (approved):
Looks ok to me.

PR #1215, opened 2022-02-09
runtime-cleanup: fix warnings from old or changed packages

AdamWill:
This addresses most "no files matched!" and "no files to remove!"
errors in current F36/Rawhide. They mostly relate to packages
that are no longer pulled in at all, or whose layout has changed.
The entries for audit haven't been fixed for the /usr merge years
ago. In linux-firmware, all files have been xz-compressed at some
point, so the entries were not matching any more. The usbdux/
subdir contains only sources for the actual firmwares that are a
level above, and we don't include those in the package any more;
the actual firmwares are useless in the installer env, so this
removes them. util-linux was split into util-linux and
util-linux-core, so we have to add an entry for util-linux-core
and move the relevant excludes to that entry.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -11,8 +11,6 @@ remove usr/share/i18n
 
 ## no sound support, thanks
 removepkg flac-libs libsndfile pipewire pulseaudio* rtkit sound-theme-freedesktop wireplumber*
-## lv2appy requires removed libsndfile, we don't need the rest either
-removefrom lilv /usr/bin/*
 ## we don't create new initramfs/bootloader conf inside anaconda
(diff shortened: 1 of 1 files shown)


bcl:
Thanks for doing all this! I think the tests are possibly failing due to branching for f36 not having propagated far enough, I see dnf repo errors in the logs.

AdamWill:
Yeah, I looked and it doesn't look like the failures are related to the PRs.

AdamWill:
Note, I think it's likely to do with branching somehow, but possibly not propagation but rather something in the CI setup not being updated? The key error is `Unable to detect release version (use '--releasever' to specify release version)`; because it can't detect the release version DNF uses an invalid URL for mirrormanager, which causes the 404.

bcl (approved):
Looks good. Did a local iso and install with all the new patches and everything worked fine.

PR #1219, opened 2022-02-16
runtime-cleanup: Remove ncurses package

sharkcz:
Remove the ncurses package -- on ppc64le and s390x it was pulled in and
the library check would fail because the library files have been
removed.

Re-adding commit 46ba3d541d623062794c44857ac65f3e575ef863 after a template cleanup based on usage on x86.


bcl (approved):
Thanks for fixing that.

PR #1222, opened 2022-03-25
runtime-cleanup: keep 'unshare' binary present from util-linux-core

KKoukiou:
This is being used by 'cockpit-desktop' [1] script that the new anaconda UI is
using.
Calling `unshare` in the installer environment is not really
necessary. However anaconda team prefers [2] to cleanly depend on
'cockpit-desktop' script instead of maintaining a modified copy of it.

[1] https://github.com/cockpit-project/cockpit/blob/main/src/ws/cockpit-desktop.in
[2] https://github.com/rhinstaller/anaconda/pull/3967#issuecomment-1076470675


KKoukiou:
CC @M4rtinK @VladimirSlavik

M4rtinK:
> CC @M4rtinK @VladimirSlavik

Looks good to me, just hoping this change is sufficient. Eq. if `unshare` uses another tool or library that gets cleaned up, it might not be enough. But I guess we will see. :)

bcl (approved):
Looks good to me. Seems to run in my test boot.iso

PR #1224, opened 2022-04-01
fedora-livemedia.ks: Use livesys-scripts

bcl:
Switch to using livesys-scripts instead of the embedded %post scripts.


VladimirSlavik:
Nice! It might be useful to have a comment with explanation and link to where the other parts of setup are happening. I see no way to guess where that code is now...

VladimirSlavik:
Thanks!

PR #1226, opened 2022-04-13
Use grub2 for BIOS boot support instead of syslinux

bcl:


share/templates.d/99-generic/efi.tmpl
```diff
@@ -21,9 +21,6 @@ install boot/grub2/fonts/unicode.pf2 ${EFIBOOTDIR}/fonts/
 
 ## actually make the EFI images
 ${make_efiboot("images/efiboot.img")}
-%if domacboot:
-    ${make_efiboot("images/macboot.img", imgtype="apple")}
-%endif
```
Conan-Kudo, line 24: Why is this being removed?

share/templates.d/99-generic/live/live-install.tmpl
```diff
@@ -18,8 +18,9 @@
     installpkg shim-ia32 grub2-efi-ia32-cdboot
 %endif
 %if basearch in ("i386", "x86_64"):
-    installpkg biosdevname syslinux
+    installpkg biosdevname
     installpkg grub2-tools grub2-tools-minimal grub2-tools-extra
+    install grub2-pc-modules
```
rathann, line 23: Why `install` instead of `installpkg`?

share/templates.d/99-generic/efi.tmpl
Conan-Kudo: That is, without this, we don't get in the Mac EFI menu the fancy "Fedora" icon that makes it obvious what we are to the firmware when booting the stick.

davdunc: +1 to @Conan-Kudo 's question here. It doesn't seem necessary to remove the integration with the intel mac efi firmware.

share/templates.d/99-generic/live/live-install.tmpl
bcl: Because it's a typo :) I haven't tested building a live iso yet. Thanks for catching that.

share/templates.d/99-generic/efi.tmpl
bcl: Because, at least in my testing, it isn't needed. And I like thinks to be as simple as possible. It's not too hard to tell what it is when booting, even without the logo.

Conan-Kudo: Does it work when the plist and logo files are installed in the regular ESP image?

bcl: No idea, but I would guess not. IIRC plist neeed to be 'blessed' in order for it to be loaded. There are so many variations to all these things that I'm not going to explore all of them just to get a logo on there.

Conan-Kudo: Ahh, yeah, so `grub2-macbless` only works with HFS+ ESP, rather than FAT32 ESP.

Conan-Kudo: I'm not actually sure if removing this will break on my 2009 MacBook Pro (which I use as a Linux laptop), I will need to test this... Last I tried it only showed me stuff with the HFS+ ESP...

Conan-Kudo: Yeah, confirmed, my 2009 MBP isn't super-happy with the FAT ESP...

AdamWill:
So just scanning this briefly - does this ultimately require us to boot via BIOS from a GPT-labelled disk? Because we did try doing that for Fedora installs for one release, and found enough systems that refused to do that that we reverted it.

That was some time ago, and of course a lot of systems were fine with it. So we might consider that a small enough problem to live with. Just wanted to flag it up.

Conan-Kudo:
It shouldn't require GPT on BIOS boot, but I _am_ preparing a separate Change proposal for doing that.

bcl:
> So just scanning this briefly - does this ultimately require us to boot via BIOS from a GPT-labelled disk? Because we did try doing that for Fedora installs for one release, and found enough systems that refused to do that that we reverted it.

No, this only effects the boot.iso boot process.

PR #1228, opened 2022-05-03
mkksiso: Rewrite to use xorriso features

bcl:
xorriso is able to extract files from an iso, list files on the iso,
update the iso to change existing files and add new ones. It also has
a -replay option that preserves the bootloader settings. This vastly
simplifies the operation of mkksiso and removes the requirement to run
it as root.


bcl:
Merged to master.

PR #1229, opened 2022-05-17
Firmware cleanups 20220509

AdamWill:
Several additional firmware file trims to save back space lost with upstream 20220509 firmware release. This saves 11M in my local testing, nearly all of the 13M that we lost with the update.


bcl (approved):
Looks good to me. Thanks for the excellent commit messages!

nullr0ute (commented):
You're probably better off just removing all of the mellanox directory rather than removing random bits, unless you can explicitly test you may well end up with devices in weird states if parts of firmware are missing. All or nothing is likely a better experience.

PR #1231, opened 2022-06-03
Update spec for syslinux->grub2 switch

AdamWill:
Since we're now using grub2 not syslinux on ix86 and x86_64, we
need to require the right things in the spec file.


bcl (approved):
Oops. Thanks for catching that.

PR #1237, opened 2022-06-08
Add branding package name support to templates

bcl:


bcl:
Includes patches from pr #1236

PR #1241, opened 2022-06-27
templates: adjust for mellanox firmware split to subpackage

AdamWill:
In Rawhide, Mellanox firmwares have been split to a subpackage,
so as usual we need to exclude that from the *-firmware install
and drop the stripping of it from linux-firmware.


bcl:
Failed tests are due to pylint not supporting python 3.11 yet.

PR #1242, opened 2022-06-28
Support other arches in lmc example kickstarts

bcl:


bcl:
Pushed to rhel9-branch

PR #1250, opened 2022-07-13
mkksiso: Set u+rw permission on extracted files and directories

bcl:
Some iso images have restrictive permissions, this makes sure that the
user running mkksiso can read and write to the files and directories
extracted from the original iso.

Related: rhbz#2088631


bcl:
Merged to master. Tests pass locally, something is up with the github action and running podman inside it (see https://github.com/containers/podman/issues/14920)

PR #1263, opened 2022-09-02
fonts packages syncup in template files

tagoh:
Remove bitmap-fangsongti-fonts because it isn't used at all
and add vazirmatn-vf-fonts according to
https://fedoraproject.org/wiki/Changes/EnhancePersianFontSupport


tagoh:
I wonder if this unit test failure are really related to this change.

bcl:
> I wonder if this unit test failure are really related to this change.

No, they should be fixed by #1260

bcl (approved):
Looks good, thanks.

PR #1264, opened 2022-09-07
Add release info flags for livemedia-creator

korewaChino:
This PR allows livemedia-creator to store release information as command line flags, similar to how vanilla allows you to create images with release information,

### DIFF
--- a/src/pylorax/cmdline.py
+++ b/src/pylorax/cmdline.py
@@ -213,6 +213,15 @@ def lmc_parser(dracut_default=""):
     parser.add_argument("--extra-boot-args", default="", dest="extra_boot_args",
                         help="Extra arguments to add to the bootloader kernel cmdline in the templates")
 
+    parser.add_argument("--release", default="", dest="release",
+                        help="Release for product information")
+    parser.add_argument("--variant", default="", dest="variant",
+                        help="Variant for product information")
+    parser.add_argument("--bugurl", default="", dest="bugurl",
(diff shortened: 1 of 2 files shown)


bcl (changes requested):
Hi, thanks for the PR! Could you make the options match their counterparts from lorax? eg. If they have a short form include that, set the same metavar, etc. But leave out `required` of course. Also, final should be `opts.isfinal` to match as well.

korewaChino:
I have copied the flags from lorax, please check again.

src/pylorax/cmdline.py
```diff
@@ -213,14 +213,14 @@ def lmc_parser(dracut_default=""):
...
 
-    parser.add_argument("--release", default="", dest="release",
-                        help="Release for product information")
-    parser.add_argument("--variant", default="", dest="variant",
-                        help="Variant for product information")
-    parser.add_argument("--bugurl", default="", dest="bugurl",
-                        help="Issue tracker URL for product information")
-    parser.add_argument("--final", action="store_true",
-                        help="Mark the image as a final release in the product information")
+    parser.add_argument("-r", "--release", help="release information", required=True, metavar="RELEASE")
```
bcl, line 216: don't include `required`

bcl (changes requested):
Other than that one change, this looks good.

bcl:
Fixed up, updated tests, merged to master. Thanks!

PR #1269, opened 2022-09-29
Drop 32-bit ARM and x86 support

AdamWill:
We don't build 32-bit ARM or x86 images in Fedora any more (we stopped doing x86 images after F30, we are no longer doing 32-bit ARM images after F36). So we can drop all of this stuff from master branch at least.


AdamWill:
@nirik for reference, am I right that we can ditch all of this as far as you're aware?

nirik:
> @nirik for reference, am I right that we can ditch all of this as far as you're aware?

Yep. As far as I know we can.

bcl (approved):
Looks good to me.

PR #1271, opened 2022-10-03
Drop anaconda auditd replacement

VladimirSlavik:
- [ ] Please do not merge yet, subject to further discussion.
- [x] anaconda PR: https://github.com/rhinstaller/anaconda/pull/4358

---

This means: Keep all things from the audit package as-is, do not erase or overwrite.

Previously:
- clean up most things audit in the template
- install a custom replacement for auditd from the anaconda packages
- in anaconda, run the replacement manually

Now:
- keep audit things as they are
- do not install the custom replacement
- in anaconda, run `auditctl -e 0` instead of the custom binary

Overall goals remain the same and met:
- keep boot.iso small: this adds +200 KiB-ish
- do not run auditd to lower runtime memory requirements: still ok
- do not spam journal/syslog with audit messages: still ok

Note: Most audit messages come from dracut, where it is not turned off anyway.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -116,9 +116,6 @@ removefrom gsettings-desktop-schemas /usr/share/locale/*
 removefrom NetworkManager-libnm /usr/share/locale/*/NetworkManager.mo
 removefrom nm-connection-editor /usr/share/applications/*
 removefrom atk /usr/share/locale/*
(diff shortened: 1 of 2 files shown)


poncovka (approved):

bcl:
The title is a bit confusing :) I read 'keep as-is' as keeping the anaconda auditd since that's the current state.
I'd change the title to something like: "Drop anaconda auditd replacement" to make it clearer what's happening.

bcl (approved):
Looks good to me, thanks!

PR #1274, opened 2022-10-05
Update kdump addon package name

VladimirSlavik:
It was pulled in via provides which is not well findable.

edit: see https://src.fedoraproject.org/rpms/kdump-anaconda-addon/blob/rawhide/f/kdump-anaconda-addon.spec


bcl (approved):
Looks fine :) Let me know when the package is available and I'll merge this.

VladimirSlavik:
This has been available for quite some time, at least 2 last Fedora releases but I suspect more. You can merge it immediately.

PR #1279, opened 2022-10-28
Improvement of unmounting /proc, /dev and binds

soksanichenko:
- Replace direct call of `umount` by function-wrapper `umount`
- Add retries and pauses for unmounting binds, because some mounts  in `/var/tmp/lorax` can be buzy at the moment unmounting in case parallel building of two or more images

### DIFF
--- a/src/pylorax/imgutils.py
+++ b/src/pylorax/imgutils.py
@@ -498,12 +498,14 @@ class DracutChroot(object):
         return self
 
     def __exit__(self, exc_type, exc_value, tracebk):
-        runcmd(["umount", self.root + "/proc" ])
-        runcmd(["umount", self.root + "/dev" ])
+        umount(self.root + '/proc')
+        umount(self.root + '/dev')
 
         # cleanup bind mounts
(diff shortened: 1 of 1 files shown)


src/pylorax/imgutils.py
```diff
@@ -498,12 +498,14 @@ def __enter__(self):
...
-        runcmd(["umount", self.root + "/proc" ])
-        runcmd(["umount", self.root + "/dev" ])
+        umount(self.root + '/proc')
+        umount(self.root + '/dev')
 
         # cleanup bind mounts
         for _, d in self.bind:
-            runcmd(["umount", self.root + d ])
+            # In case parallel building of two or more images
+            # some mounts in /var/tmp/lorax can be buzy at the moment of unmounting
```
bcl, line 507: `buzy` should be `busy`

soksanichenko: Thank you! Fixed.

bcl:
Thanks, I squashed the commits together and pushed to master locally.

PR #1280, opened 2022-11-01
setup.py: Set the version number

### LINKED ISSUE #1278: lorax has an invalid version devel
If you install lorax on Fedora 36, it will use the version `devel` which is not a valid version according to [pep440](https://peps.python.org/pep-0440/). This is causing issues for tools like poetry, that require your versions to be valid according to PEP 440. Could you please use a valid version?

bcl:
Previously this was trying to import the version from pylorax. This cannot work if the dependencies have not yet been installed, resulting in a 'devel' version showing up in unexpected places. It actually never really worked right.

Lorax uses tito for releases, and it understands how to update setup.py version numbers, so drop the use of 'devel' and set the version to a string.

Fixes #1278


VladimirSlavik (approved):
Looks good to me, less things to break :)

PR #1283, opened 2022-11-07
runtime-postinstall: Disable sshd.socket

bcl:
It interferes with inst.sshd and anaconda's sshd config.

Resolves: rhbz#2136916


jkonecny12:
Thanks for creating this @bcl . Could you please add a comment that it's added because of https://bugzilla.redhat.com/show_bug.cgi?id=2136916 and also to make the environment more robust.

Thanks!

M4rtinK (approved):
Looks good to me & should fix the issue. Thanks! :)

PR #1287, opened 2022-11-28
Use unicode.pf2 from /usr/share/grub/

bcl:
grub2 wants to remove it from /boot/grub2/fonts/ so use it from a more stable location instead.

Also add grub2-tools on aarch64 to make sure that it is installed.

Related: rhbz#2146545


AdamWill:
Yeah, makes sense to me.

bcl:
Merged to master.

PR #1291, opened 2022-12-09
Firmware cleanups 2022-12

AdamWill:
This saves us about 11M, and will solve the Rawhide netinsts being oversize (which was caused by webkitgtk pulling in more gstreamer bits).


AdamWill:
@nullr0ute just in case you have any notes, I think these are all safe.

bcl (approved):
Looks ok to me, but I'll give people the weekend to comment.

PR #1294, opened 2022-12-15
Firmware cleanup 202212 2

AdamWill:
We don't need to exclude bfa-firmware any more as I got it retired - https://bugzilla.redhat.com/show_bug.cgi?id=2152202 . I confirmed it's not in today's Rawhide compose at all. Checking lorax output from today's Rawhide I also found a couple more *-firmware packages we can exclude as they're not for hardware that's likely to be used during install. Probably these don't save much space, but every little helps.


bcl (approved):
ACK

PR #1299, opened 2023-01-10
rsyslog.conf: Set WorkDirectory to /var/lib/rsyslog

bcl:
Without this set rsyslogd will try to use / for things like /imjournal.state.tmp, which will fail due to SELinux.

Resolves: rhbz#2158731


poncovka (approved):
Thank you! I was just investigating this flood of logs on Fedora.

bcl:
@poncovka It also looks like it is time to update the rsyslog.conf format, we're currently still using the 'obsolete legacy' format :) 
I'll take a look at doing that in Fedora first.

jstodola (approved):
Thanks

Issue #1300, opened 2023-01-10
Update rsyslog.conf


bcl:
The config file currently uses the legacy format. It should be updated to use the current style, and workDirectory should be set to avoid SELinux errors generated by it attempting to write to `/imjournal.state.tmp`

bcl:
Closed by #1301

PR #1301, opened 2023-01-27
rsyslogd.conf: Rewrite using current config syntax

bcl:
The rsyslogd.conf file was written a long, long time ago. The format that is was using is now called the the obsolete legacy format, so it is time to update it.

It supports several levels of config commands, and they can mixed. So I have converted the module loading and global config settings to advanced, and left the templates and actions alone for now.

This also include setting the workDirectory, which fixes an error when it tries to write the state file, and replaces the '~' action with 'stop'.


bcl:
This will fix #1300

PR #1306, opened 2023-02-09
templates.d/99-generic/live: Enable automatic persistence for live media

Conan-Kudo:
This change alters the live environment operating mode to default to operating with full root system persistence using the functionality recently added to Dracut.

It also adds menu entries so that it is possible to reset the persistence or start without persistence at all.

Reference: https://fedoraproject.org/wiki/Changes/ModernizeLiveMedia

Co-authored-by: @iammattcoleman

### DIFF
--- a/share/templates.d/99-generic/live/config_files/aarch64/grub2-efi.cfg
+++ b/share/templates.d/99-generic/live/config_files/aarch64/grub2-efi.cfg
@@ -26,16 +26,32 @@ set timeout=60
 search --no-floppy --set=root -l '@ISOLABEL@'
 
 ### BEGIN /etc/grub.d/10_linux ###
-menuentry 'Start @PRODUCT@ @VERSION@' --class red --class gnu-linux --class gnu --class os {
-	linux @KERNELPATH@ @ROOT@ @EXTRA@ rd.live.image quiet rhgb
+menuentry 'Start @PRODUCT@ @VERSION@ with automatic data persistence' --class fedora --class gnu-linux --class gnu --class os {
+	linux @KERNELPATH@ @ROOT@ @EXTRA@ rd.live.image rd.live.overlay.overlayfs rd.live.overlay=LABEL=fedorapersist quiet rhgb
 	initrd @INITRDPATH@
 }
-menuentry 'Test this media & start @PRODUCT@ @VERSION@' --class red --class gnu-linux --class gnu --class os {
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
cc: @dcavalca @michel-slm @tdawson

Conan-Kudo:
cc: @davide125 @michel-slm @tdawson

share/templates.d/99-generic/live/config_files/aarch64/grub2-efi.cfg
```diff
@@ -26,16 +26,32 @@ set timeout=60
 search --no-floppy --set=root -l '@ISOLABEL@'
 
 ### BEGIN /etc/grub.d/10_linux ###
-menuentry 'Start @PRODUCT@ @VERSION@' --class red --class gnu-linux --class gnu --class os {
-	linux @KERNELPATH@ @ROOT@ @EXTRA@ rd.live.image quiet rhgb
+menuentry 'Start @PRODUCT@ @VERSION@' --class fedora --class gnu-linux --class gnu --class os {
```
bcl, line 29: I think this should be changed to make it clear it will start with persistence.

bcl (changes requested):
Also, how does this behave on a cdrom -- Does it transparently fall back to non-persistence or does the user need to boot the other entry?

Conan-Kudo:
It is supposed to transparently fail and fallback to in-memory persistence.

https://github.com/dracutdevs/dracut/blob/28f37724441a0d63fb1505deefbf5d4838b9818c/modules.d/90dmsquash-live-autooverlay/create-overlay.sh#L54-L60

share/templates.d/99-generic/live/config_files/aarch64/grub2-efi.cfg
Conan-Kudo: I'm not sure what to word here for what amounts to the default experience.

bcl: How about `Start ... with automatic persistance`?

bcl (approved):
Looks good, thanks!

Conan-Kudo:
Thanks for merging! 🥇

bcl:
This needs to be re-opened, and tested fully before it can be merged. See https://bugzilla.redhat.com/show_bug.cgi?id=2170544

PR #1307, opened 2023-02-14
RFC: creator.py: Allow explicit plymouth module requests for initrd.

FGrose:
Adjust DRACUT_DEFAULT to allow plymouth if expressly requested.

Also, refactor to remove redundant calls to dracut_args(). Correct a typo.

***********************************
`DRACUT_DEFAULT` can be overridden by `dracut-conf` or `dracut-args` as options to _livemedia-creator_, but
is there a way to set `dracut-conf` or `dracut-args` in the kickstart?
 * If so, this change is not needed...

This adjustment is needed for Fedora Sugar on a Stick (SoaS spin).

### DIFF
--- a/src/pylorax/creator.py
+++ b/src/pylorax/creator.py
@@ -36,7 +36,7 @@ from pykickstart.version import makeVersion
 # Use the Lorax treebuilder branch for iso creation
 from pylorax import DEFAULT_RELEASEVER, ArchData
 from pylorax.base import DataHolder
-from pylorax.executils import execWithRedirect
+from pylorax.executils import execWithRedirect, runcmd_output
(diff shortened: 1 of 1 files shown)


bcl:
Can you explain a bit more about what this is trying to fix? There is no way to control dracut configs in the kickstart, and even if there were those would not be used for the initrd on the iso.

As for the patch, two problems I have with it are -- running a cmdline tool to do something that can be done in python directly, and the other is modifying the DRACUT_DEFAULT value which is supposed to be treated like a constant. Maybe we can just drop `--omit plymouth`? I don't remember exactly what that's in there.

bcl:
See #1308 :) I removed it, and at least in a VM things look fine on BIOS and UEFI boot.

PR #1308, opened 2023-02-14
livemedia-creator: Do not omit plymouth module from dracut

bcl:
This was added in the initial version of livemedia-creator, copied over from the boot.iso dracut cmdline. This was changed in commit 6d1e637731da5eb9f89bb35c4da9c6de3785dc07, but livemedia-creator wasn't updated.


Conan-Kudo (approved):

src/pylorax/creator.py
```diff
@@ -49,7 +49,7 @@
 
 # Default parameters for rebuilding initramfs, override with --dracut-arg or --dracut-conf
 DRACUT_DEFAULT = ["--xz", "--add", "livenet dmsquash-live dmsquash-live-ntfs convertfs pollcdrom qemu qemu-net",
```
FGrose, line 51: include `dmsquash-live-autooverlay` to address https://bugzilla.redhat.com/show_bug.cgi?id=2170544

bcl: Yeah, that should be a separate set of patches since it's not directly related.

PR #1309, opened 2023-02-16
Strip some things from gtk4

AdamWill:
gtk4 is now getting sucked into installer images because mutter depends on it (since 44-beta). Strip some stuff the installer env currently doesn't need. The /usr/bin/* removal may not be safe long-term, but it should be OK for now.


AdamWill:
I'm backporting this to Rawhide as we need it to get a compose through; /usr/bin/gtk4-launch requires libtiff, which we strip elsewhere in runtime-cleanup, so unless we resolve that we can't make a compose work.

AdamWill:
The failures have nothing to do with the PR.

bcl (approved):
Looks like there's only 3 files under bin right now, so as long as they aren't really needed it's ok.

AdamWill:
at some point one of them might be and we might have to tweak that, but for right now, the only reason mutter needs gtk4 is it added a [new binary that links against it](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2175). I'm pretty sure that shouldn't call any of the binaries.

If we want to be "safer" we could only drop `/usr/bin/gtk4-launch`, but OTOH, that sort of seems like the binary that's most likely to eventually be needed if any of them are.

The only other option is to stop stripping libtiff.

PR #1312, opened 2023-02-17
Don't strip gtk4 binaries or libtiff (#2170716)

AdamWill:
It turns out that, somehow, stripping the gtk4 binaries caused: https://bugzilla.redhat.com/show_bug.cgi?id=2170716 at least, I tested a scratch build with this change and that bug didn't seem to happen any more. This puts the gtk4 binaries back, and also stops stripping libtiff, because gtk4-launch depends on it so we need to keep it in. This doesn't seem to cause too much of a size difference; I see 723417088 bytes for the netinst with this change, compared to 723197952 bytes for today's nightly.


bcl (approved):
ugh. thanks for catching that.

AdamWill:
I'm like alcohol - the cause of, and solution to, all your problems!

PR #1317, opened 2023-03-22
Pass vga=791 for live basic graphics mode on BIOS (#2176782)

AdamWill:
As per https://bugzilla.redhat.com/show_bug.cgi?id=2176782 , this seems to be the best way to make Wayland work on basic graphics mode on BIOS.

We don't need this for non-live, at least currently, as in the traditional installer images, anaconda runs on X.org. We may need to revisit this if we change the traditional installer env to run on Wayland.


AdamWill:
It'd be good to have some input from the folks on https://bugzilla.redhat.com/show_bug.cgi?id=2176782 . I mainly posted the PR to show how potentially simple the change is.

bcl:
This seems like the safest approach, along with documentation in the release notes so that if this specific value doesn't work they can pick something else.

PR #1319, opened 2023-04-06
lorax conversion to use libdnf5

bcl:


bcl:
Something is broken in the current rawhide container, switching it to Fedora 39 passes -- but I'm leaving it as rawhide so that it doesn't get left on 39  by accident :)

PR #1325, opened 2023-05-17
monitor: Skip repo errors involving CDROM/file source

bcl:
When using the rhsm command anaconda first tries the cdrom, and when it fails then tries the CDN. This causes lorax-composer to see the repo failure and exit too early, preventing it from working with rhsm.

This should never show up during normal lmc usage, so it's safe to skip. Includes new tests in LogMonitorTest


VladimirSlavik (commented):
LGTM, but I can't really say I am the authority on this.

You probably want @M4rtinK or @poncovka to confirm this is right and you are supposed to ignore the cdrom error...

ondrejbudai (commented):
I'm not an anaconda expert but the change looks overall good.

PR #1326, opened 2023-06-07
Remove retired paktype-naskh-basic-fonts from runtime-install

AdamWill:
See https://pagure.io/releng/failed-composes/issue/5031 . I'm hoping the remaining fonts, particularly all the Noto ones, should provide at least sufficient coverage for the language(s) we had this font in the installer environment for.


AdamWill:
Note, removing this font *does* change rendering of some languages, at least Urdu and one variant of Punjabi. I am not sure if the new rendering is reasonable. We're following up on that at https://pagure.io/releng/failed-composes/issue/5031 . I'm filing this PR so it's conveniently available for merge if we do decide the new rendering is OK (or if we don't have any other better options).

AdamWill:
Fonts folks decided to unretire the package - https://pagure.io/releng/issue/11466 - so closing this.

PR #1327, opened 2023-06-09
Add nvme-cli to the install image

cmadamsgit:
Include the nvme command in the install image.  This goes with other disk utilities like smartmontools.  Having it in the install image allows switching an NVMe device to 4K sectors, which requires formatting (so needs to be done before installing an OS).

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -140,6 +140,7 @@ installpkg rng-tools
 %if basearch in ("x86_64", "aarch64"):
 installpkg dmidecode
 %endif
+installpkg nvme-cli
 
 
 ## fonts & themes


VladimirSlavik:
I'd suggest instead adding it to anaconda-img-deps, since we want to do that. It's a side effect of a PR for NVMe over Fabrics. That is stuck on having the dependencies in place, but planned.

https://github.com/rhinstaller/anaconda/pull/4514/files
https://github.com/rhinstaller/anaconda/pull/4514#issuecomment-1525673014

cmadamsgit:
I brought this up on fedora-devel first, and Brian Lane said nvme-cli had already been added to the RHEL 9 branch, so I looked at how it was done there. If it gets pulled in through another path that'd be fine with me... I just noticed it missing when I was trying to reformat a new NVMe drive to 4K sectors.

bcl:
> I'd suggest instead adding it to anaconda-img-deps,

When this was added in RHEL9 we ran into a problem with it installing machine specific files, so it ends up you still need to change lorax. See commits:
f031f2850b3823fca0360ee9668702ab4139ec5f and 0099d618be177325487d2c23ec2d8a765d03ba1d

VladimirSlavik:
Wow, oof. How did I miss that? In that case yes, it's better to keep it as a lorax thing, if there's more to do about it...

PR #1328, opened 2023-06-17
runtime-cleanup: don't strip device-mapper-event

AdamWill:
The openQA install_blivet_lvmthin test - which, as the name implies, does an LVM thinp install - is consistently failing (only on ppc64le, for some reason). The logs show that lvm2 is trying to run dmeventd as part of some kind of check of the pool, but it's failing because dmeventd isn't there:

Failed to call the 'LvCreate' method on the '/com/redhat/lvmdbus1/ThinPool/0' object: GDBus.Error:org.freedesktop.DBus.Python.dbus.exceptions.DBusException: ('com.redhat.lvmdbus1.Lv', 'Exit code 5, stderr = /usr/sbin/dmeventd: stat failed: No such file or directory, /usr/sbin/dmeventd: stat failed: No such file or directory, Check of pool fedora/00 failed (status:64). Manual repair required!, Failed to activate thin pool fedora/00.')

So, apparently we do need this now. Whether to use dmeventd or not seems to be a build time check/config option for lvm2, but I don't think we can turn that off just to save a bit of space in the installer image - the package is only about 50K.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -34,8 +34,6 @@ removefrom shadow-utils --allbut /usr/bin/chage /usr/sbin/chpasswd \
 ## no services to turn on/off (keep the /etc/init.d link though)
 removefrom initscripts /usr/sbin/* /usr/share/locale/* /usr/share/doc/* /usr/share/man/*
 
-## no storage device monitoring
-removepkg device-mapper-event
 ## logrotate isn't useful in anaconda
 remove /etc/logrotate.d
 ## anaconda needs this to do media check


bcl:
It looks like this starts the service when it is installed so I'm a bit cautious about adding it. I'll try to do some tests on a new iso soon, but it might also be a good idea to ask @poncovka or @M4rtinK what Anaconda thinks about this.

bcl:
Adding this turns on the dm-event.socket listener, and during the install I see logging about dmeventd trying to dlopen `libdevmapper-event-lvm2thin.so` which fails because we are removing `lvm2-libs`. It seems to fail gracefully, on x86_64 at least, and the install using lvm thinp finishes and boots. 

But if we also add `dm-event.socket dm-event.service` to the `systemctl disable` in `runtime-postinstall.tmpl` it seems to be happy on x86 at least. The lvm2-libs package adds about 2M so if we can avoid adding that back in it would be nice.

bcl:
@AdamWill is there any way you could test https://github.com/weldr/lorax/pull/1343 with ppc64le?

AdamWill:
Yes, but it requires a bit of bodging and it's not my number 1 priority right now, so I can't promise I'm gonna get to it. I will try, though!

AdamWill:
BTW, the bug is actually fixed in Rawhide already since 2023-09-01, because Marian's fix for device-mapper-persistent-data landed that day ( device-mapper-persistent-data-1.0.6-2.fc40 ).

AdamWill:
So, since the install fail is already fixed a different way on Rawhide, I can't immediately confirm that this would also have fixed it...but it at least doesn't *break* install again: https://openqa.stg.fedoraproject.org/tests/3141546

I can re-run the process on F39 tomorrow, I guess. F39 doesn't have the other fix yet, so we can see if this fixes the install there.

bcl:
> I can re-run the process on F39 tomorrow, I guess. F39 doesn't have the other fix yet, so we can see if this fixes the install there.

Thanks!

bcl:
Merged the monitoring=0 change in #1343

bcl:
Hrm. So the error about dmeventd isn't there anymore. I see:

`ERROR:anaconda.modules.storage.installation:Failed to create storage layout: Failed to call the 'LvCreate' method on the '/com/redhat/lvmdbus1/ThinPool/0' object: GDBus.Error:org.freedesktop.DBus.Python.dbus.exceptions.DBusException: ('com.redhat.lvmdbus1.Lv', 'Exit code 5, stderr = Check of pool fedora_fedora/00 failed (status:64). Manual repair required!, Failed to activate thin pool fedora_fedora/00.')`

In the storage log for the failed test. But I can't tell what caused it.

AdamWill:
I guess that *could* still simply be caused by the bug the device-mapper-persistent-data update fixes?

PR #1332, opened 2023-07-04
runtime-install: exclude renamed iwl firmware packages

AdamWill:
@pbrobinson renamed the Intel wireless firmware packages in the 20230625 build of linux-firmware. This causes installer image builds to fail because they want to pull in the old packages from the release repo because they fall in the "*-firmware" glob, but those packages are now obsoleted by the newer ones from updates/updates-testing, and dnf doesn't like that. This should resolve the problem by specifically excluding all the renamed packages from the glob. Of course, this change must go along with the newer linux-firmware package, or else you'll get an image that's missing a lot of wireless firmware.


AdamWill:
Note, we need this on F37 and F38 branches too. I am doing backport builds with the patch downstream to edit into the linux-firmware updates.

bcl (approved):
Ack.

bcl:
Also cherry picked to f38-branch and f37-branch

PR #1335, opened 2023-08-06
runtime-install: excluded renamed olpc firmware package

AdamWill:
Similar to c7c6dc79d97777d00d4d1b23b8807734a56935e5 , this specifically excludes a firmware package that was renamed in the recent linux-firmware update so image builds work with both packages present in the available set (e.g. stable releases with the older firmware in the release repo and the newer one in the updates repo). Must go along with the linux-firmware update.


bcl (approved):
Ack.

AdamWill:
we'll need this on f37 and f38 too, as @nullr0ute wants to send the update to all three.

PR #1337, opened 2023-08-09
Exclude more obsoleted libertas firmware packages

AdamWill:
openQA test was showing an error only for the first of these it encountered, but actually, there used to be four different libertas firmware packages, we need to exclude all of them.


AdamWill:
again we'll need this on f38 and f37 branches. sorry for the extra work.

bcl (approved):
ACK

PR #1338, opened 2023-08-11
runtime-install: only pull in qcom-firmware on aarch64

AdamWill:
See https://bugzilla.redhat.com/show_bug.cgi?id=2178852 . This should save a chunk of space on the installer images for x86_64.

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -28,6 +28,8 @@ installpkg grubby
     ## https://bugzilla.redhat.com/show_bug.cgi?id=2011615
     ## bfa-firmware contains only obsolete files - see
     ## https://bugzilla.redhat.com/show_bug.cgi?id=2152202
+    ## qcom-firmware we pull in again lower down but *only* on aarch64, it is
+    ## no use on other arches - https://bugzilla.redhat.com/show_bug.cgi?id=2178852
     ## various iwl package names were changed in linux-firmware-20230625-151
     ## so need to be excluded or else dnf gets sad - see
     ## https://pagure.io/releng/issue/11511 . These exclusions can
@@ -53,7 +55,8 @@ installpkg grubby
                            --except iwlax2xx-firmware \
                            --except libertas-sd8686-firmware --except libertas-sd8787-firmware \
                            --except libertas-usb8388-firmware \
-                           --except libertas-usb8388-olpc-firmware
+                           --except libertas-usb8388-olpc-firmware \
+                           --except qcom-firmware
     installpkg b43-openfwwf
 %endif
 
@@ -67,6 +70,7 @@ installpkg glibc-all-langpacks
     installpkg grub2-tools>=${GRUB2VER}
     installpkg shim-aa64
     installpkg uboot-tools
+    installpkg qcom-firmware
 %endif
 %if basearch == "x86_64":
     installpkg grub2-tools-efi>=${GRUB2VER}


AdamWill:
@nullr0ute

share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -53,7 +55,8 @@ installpkg grubby
                            --except iwlax2xx-firmware \
                            --except libertas-sd8686-firmware --except libertas-sd8787-firmware \
                            --except libertas-usb8388-firmware \
-                           --except libertas-usb8388-olpc-firmware
+                           --except libertas-usb8388-olpc-firmware \
```
nullr0ute, line 58: this should be libertas-firmware now (combined they were all just 512kb)

share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -28,6 +28,8 @@ installpkg grubby
     ## https://bugzilla.redhat.com/show_bug.cgi?id=2011615
     ## bfa-firmware contains only obsolete files - see
     ## https://bugzilla.redhat.com/show_bug.cgi?id=2152202
+    ## qcom-firmware we pull in again later but *only* on aarch64, it is
```
nullr0ute, line 31: By later you mean later in the script or "we'll deal with this later" because we do need this as we support QCom devices where the lack of firmware will break them so just checking.

AdamWill: later in the script, in the other added line.

share/templates.d/99-generic/runtime-install.tmpl
AdamWill: no, the point of these lines is to exclude the *old* names. this is what I had to do to fix the tests failing on the linux-firmware updates because lorax would try to pull in *both* the old packages (from the original release repo) and the new packages (from the update), but the new packages obsolete the old ones, so it'd fail.

share/templates.d/99-generic/runtime-install.tmpl
AdamWill: agh, it looks like I didn't save the file again before sending the PR, so the new line is missing. :D will fix

share/templates.d/99-generic/runtime-install.tmpl
nullr0ute: ah, OK, it's hard for context sometimes in small commits :)

nullr0ute (approved):
LGTM

AdamWill:
changed "later" to "lower down" to be less ambiguous, and added the missing line to actually install qcom-firmware on aarch64 :)

share/templates.d/99-generic/runtime-install.tmpl
```diff
@@ -28,6 +28,8 @@ installpkg grubby
     ## https://bugzilla.redhat.com/show_bug.cgi?id=2011615
     ## bfa-firmware contains only obsolete files - see
     ## https://bugzilla.redhat.com/show_bug.cgi?id=2152202
+    ## qcom-firmware we pull in again lower down but *only* on aarch64, it is
+    ## no use on other arches - https://bugzilla.redhat.com/show_bug.cgi?id=2178852
```
bcl, line 32: 'not used' :)

bcl (changes requested):
Just a tiny comment correction

share/templates.d/99-generic/runtime-install.tmpl
AdamWill: well, I can change it if you like, but "no use" is a common phrase meaning "not of any use"?

bcl: Or 'of no use' :) But it doesn't look as goofy as it did pre-coffee so it's fine as is.

bcl (approved):

PR #1342, opened 2023-09-04
runtime-cleanup.tmpl: fix typo 'gschadow'

cjriches:
Commit fa2158c7a9 ("Drop non-determinism from default templates") made an attempt to ensure /etc/gshadow is reproducible by sorting it. However, there was a typo causing the sorted version to be saved as /etc/gschadow, leaving the original intact.

Fix the typo to ensure the file is overwritten.


cjriches:
I also noticed `/etc/passwd-` and `/etc/shadow-` files present with a few differences (the suffiexed version was missing `rpcuser` and `install` and had `root` login disabled). I couldn't find what code was creating these, and I'm not sure if they're an issue or not. Certainly a separate one if they are.

bcl (approved):
Good catch, thanks!

PR #1344, opened 2023-09-07
Explicitly pull in more filesystem packages

AdamWill:
Turns out we've been relying on udisks2 dependencies to pull in several filesystem tools packages that anaconda actually needs to create and read supported filesystems. udisks2 just changed all those dependencies to Recommends, which don't get pulled into the installer environment. So we need lorax to specifically list all of these. e2fsprogs was already being pulled in through some other dep chain, but let's explicitly list it just to make sure we don't hit the same problem in future. From a look through the blivet code, I don't think pulling in udftools, exfatprogs or nilfs-utils will be useful, but we can check that with @vojtechtrefny  .

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -118,7 +118,7 @@ installpkg systemd-sysv systemd-units
 installpkg rsyslog
 
 ## filesystem tools
-installpkg btrfs-progs jfsutils xfsprogs ntfs-3g ntfsprogs
+installpkg btrfs-progs jfsutils xfsprogs ntfs-3g ntfsprogs dosfstools e2fsprogs f2fs-tools
 installpkg system-storage-manager
 installpkg device-mapper-persistent-data
 installpkg xfsdump


AdamWill:
udisks2 change for reference: https://src.fedoraproject.org/rpms/udisks2/c/92fceb0c5dabd8f7845a1648d3e5ff3140e16edc?branch=rawhide

bcl (approved):
Looks good to me!

vojtechtrefny:
> I don't think pulling in udftools, exfatprogs or nilfs-utils will be useful, but we can check that with @vojtechtrefny .

We don't support NILFS at all in blivet and we don't support creating or resizing ExFAT or UDF so we don't need the tools for those either.

I am only wondering whether the `anaconda-install-env-deps` package would be a better place for this dependency. It already depends on hfsplus-tools.

AdamWill:
It probably would, yeah, but this is where most of the stuff is and I wanted it fixed pronto. In general we might want to move quite a few things from lorax's templates to there.

bcl:
> I am only wondering whether the `anaconda-install-env-deps` package would be a better place for this dependency. It already depends on hfsplus-tools.

That's the right way to do this. At some point I'll try to do a pass through the templates and come up with a PR.

VladimirSlavik:
Agreed with all of the above - better put it here now, then move it all, eventually.

PR #1348, opened 2023-09-19
docs: Update the quickstart example command

vojtechtrefny:
The updates repository is now split into 'Everything' and 'Modular' so the link is no longer valid for supported versions of Fedora.


vojtechtrefny:
The version update is not that important, but this like fifth time Lorax failed for me because changing the version in the example isn't enough for the updates repo to work :-)

bcl (approved):
Thanks!

PR #1352, opened 2023-10-25
Process exit code from `mksquashfs`:

soksanichenko:
- Lorax should process an exit code from `mksquashfs` and stop itself execution. Mksquashfs can fail in some cases, e.g. no space left on device. Lorax is failed with non-clear error message and exception in that case, like `OSError: nothing matching /var/tmp/lorax/lorax.1j_tpfz1/installroot/images/install.img in /`.
- That PR improves the error message and can help to a user find a root of issue


soksanichenko:
The PR was based on the old version of Lorax. Closed.

PR #1354, opened 2023-12-08
Adjust runtime-postinstall.tmpl for systemd config files move

keszybz:
Systemd config files were moved from /etc to /usr/lib/systemd, so this snippet fails. Instead of editing the config file, just create a drop-in snippet with the desired configuration.


keszybz:
Yikes. Updated.

bcl (approved):
Thanks!

PR #1366, opened 2024-01-22
Add support for --updates to mkksiso

jkonecny12:
It would be great way to avoid requirement for HTTP server for local development.

### DIFF
--- a/src/bin/mkksiso
+++ b/src/bin/mkksiso
@@ -434,7 +434,7 @@ def CheckDiscinfo(path):
             raise RuntimeError("iso arch does not match the host arch.")
 
 
-def MakeKickstartISO(input_iso, output_iso, ks="", add_paths=None,
+def MakeKickstartISO(input_iso, output_iso, ks="", updates_image="", add_paths=None,
                     cmdline="", rm_args="", new_volid="", implantmd5=True,
                     skip_efi=False):
     """
@@ -511,7 +511,11 @@ def MakeKickstartISO(input_iso, output_iso, ks="", add_paths=None,
         for p in add_paths:
             cmd.extend(["-map", p, os.path.basename(p)])
 
-        if CheckBigFiles(add_paths):
+        # include updates image which will be automatically loaded when on correct place
+        if updates_image:
+            cmd.extend(["-map", updates_image, "updates/updates.img"])
+
+        if CheckBigFiles(add_paths if not updates_image else add_paths + [updates_image]):
             if "-as" not in cmd:
                 cmd.extend(["-as", "mkisofs"])
             cmd.extend(["-iso-level", "3"])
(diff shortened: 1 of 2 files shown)


bcl:
You should already be able to do this by passing `--add /path/to/updates/` which will add the `updates/` directory and all of its content to the root of the iso.

Having a dedicated command is slightly nicer of course, but it also skips the CheckBigFiles function.

jkonecny12:
Yes, I know but it means that you have to document the approach of how to use this command and it's prone to typos. For that reason I think separate option is better and could be promoted as easy way to develop / debug Anaconda.

I'll add the CheckBigFiles support.

src/bin/mkksiso
```diff
@@ -511,6 +511,10 @@ def MakeKickstartISO(input_iso, output_iso, ks="", add_paths=None,
         for p in add_paths:
             cmd.extend(["-map", p, os.path.basename(p)])
 
+        # include updates image which will be automatically loaded when on correct place
+        if updates_image:
+            cmd.extend(["-map", updates_image, "updates/updates.img"])
+
         if CheckBigFiles(add_paths):
```
bcl, line 518: The updates_image also needs to be checked to see if it is gigantic :)

bcl (changes requested):

src/bin/mkksiso
jkonecny12: Done.

src/bin/mkksiso
```diff
@@ -555,6 +559,8 @@ def setup_arg_parser():
                         help="Do not run implantisomd5 on the ouput iso")
     parser.add_argument("--ks", type=os.path.abspath, metavar="KICKSTART",
                         help="Optional kickstart to add to the ISO")
+    parser.add_argument("-u", "--updates", type=os.path.abspath, metavar="KICKSTART",
+                        help="Optional updates image to add to the ISO")
```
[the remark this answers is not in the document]
jkonecny12: My bad - fixing.

src/bin/mkksiso
```diff
@@ -599,14 +605,14 @@ def main():
             log.error("Use either --ks KICKSTART or positional KICKSTART but not both")
             errors = True
 
-        if not any([args.ks or args.ks_pos, args.add_paths, args.cmdline, args.rm_args, args.volid]):
+        if not any([args.ks or args.ks_pos, args.updates, args.add_paths, args.cmdline, args.rm_args, args.volid]):
             log.error("Nothing to do - pass one or more of --ks, --add, --cmdline, --rm-args, --volid")
```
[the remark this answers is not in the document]
jkonecny12: Will be fixed.

jkonecny12:
Updated to resolve pending issues.

src/bin/mkksiso
```diff
@@ -511,7 +511,11 @@ def MakeKickstartISO(input_iso, output_iso, ks="", add_paths=None,
         for p in add_paths:
             cmd.extend(["-map", p, os.path.basename(p)])
 
-        if CheckBigFiles(add_paths):
+        # include updates image which will be automatically loaded when on correct place
+        if updates_image:
+            cmd.extend(["-map", updates_image, "updates/updates.img"])
+
+        if CheckBigFiles(add_paths + [updates_image]):
```
bcl, line 518: You need to check if updates_img is empty, otherwise you end up with `[""]` (which the tests caught, woo hoo!)

jkonecny12: Should be fixed. Feel free to say if you don't like to have this kind of expression. I like it but might be harder to read for someone.

bcl: I usually like those, but I think having it as an argument to a function call makes the function it harder to read. Even just moving it up and giving it a new variable:

```python
check_paths = add_paths if not updates_image else add_paths + [updates_image]
if CheckBigFiles(check_paths):
```
is easier, I think.

bcl:
I tweaked the CheckBigFiles call and rewrote the commit message to be a bit clearer about what this does and merged it locally.
Thanks!

PR #1370, opened 2024-01-26
runtime-install: wget2-wget has replaced wget

AdamWill:
wget has been retired:
https://src.fedoraproject.org/rpms/wget/c/ce69c17
in favor of wget2-wget:
https://fedoraproject.org/wiki/Changes/Wget2asWget

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -184,7 +184,7 @@ installpkg python3-pyatspi
 ## extra tools not required by anaconda
 installpkg nano nano-default-editor
 installpkg vim-minimal strace lsof dump xz less
-installpkg wget rsync bind-utils ftp mtr vconfig
+installpkg wget2-wget rsync bind-utils ftp mtr vconfig
 installpkg spice-vdagent
 installpkg gdisk hexedit sg3_utils
 


AdamWill:
as this broke rawhide and bcl doesn't seem to be around rn, I'm doing a backport build so we can get a rawhide out.

bcl:
Thanks for handling things. Why doesn't wget2-wget provide wget though? Seems like that's going to cause a whole bunch of confusion down the road.

AdamWill:
it does provide it, but IIRC, lorax needs exact binary package names, it can't find them via provides. comps is similar.

bcl:
> it does provide it, but IIRC, lorax needs exact binary package names, it can't find them via provides. comps is similar.

Lorax uses dnf to depsolve, it works fine with provides. Or should. Or did :) Maybe a dnf5 thing. I'm fine with the change, but I want to make sure there's not something else going on here first.

AdamWill:
oh, huh, then that's weird. This is the Rawhide compose failure we were dealing with: [failed-composes issue](https://pagure.io/releng/failed-composes/issue/5865), [failed task](https://koji.fedoraproject.org/koji/taskinfo?taskID=112390239), [the actual failure](https://kojipkgs.fedoraproject.org//work/tasks/239/112390239/root.log). The backported patch did seem to fix it.

bcl:
FTR I've filed #1372 to track this change in behavior.

bcl (approved):
ack

PR #1371, opened 2024-01-29
runtime-install: drop retired pcmciautils

AdamWill:
pcmciautils was just retired in Rawhide:
https://src.fedoraproject.org/rpms/pcmciautils/c/24639b0


bcl (approved):
About time... ;)

Issue #1372, opened 2024-01-29
Lorax installpkg no longer uses Provides


bcl:
The package names must match a package name for now. Working on a fix or workaround, but currently things like #1370 are needed.

Also asking for suggestions from dnf5 team - https://github.com/rpm-software-management/dnf5/issues/1212

bcl:
Fixed by #1373

PR #1373, opened 2024-01-29
ltmpl: Handle installing provides with filter_provides

bcl:
Switch to using resolve_pkg_spec function.


bcl:
Workaround for #1372

bcl:
Ends up there's a function for all this! So now this mostly is just a pass-thru to dnf5, including the version comparisons.

bcl:
Unit tests passed, but didn't get reported for some reason. coverage change isn't a big deal.

PR #1374, opened 2024-02-02
Add example of using livemedia-creator inside GitHub Actions

Cameronsplaze:


Cameronsplaze:
The action finally builds! I didn't realize the @anaconda-tools was required in packages, that took a bit to figure out.

The one catch is when I try to install the iso to virt-manager, it skips the install screen and jumps straight to asking for a user/pass. Is that expected? Is there a way to still force a install?

Cameronsplaze:
I think I found it, I was missing the 'firstboot' flag [here](https://docs.fedoraproject.org/en-US/fedora/f36/install-guide/appendixes/Kickstart_Syntax_Reference/#sect-kickstart-commands-firstboot). It's going through the build process now, and I'll try to install it tomorrow then mark the PR as ready. This is the first ks I've made, so any suggestions are welcomed!

Cameronsplaze:
I can't get the firstboot flag to work. Instead of taking you to the setup screen, it takes you to the user login screen. Any idea what I might be missing? I tried a few ideas locally, and nothing worked. I'm not sure if this is related, but the "Test this Media" option at startup would fail too, "checkisomd5@*.service failed to start".

bcl:
I'm not sure what's going on with firstboot and the checksum. I almost never use firstboot so I'm not sure of the current state of it in Anaconda. A few suggestions for this PR:

* Don't put the action into the repo workflow, put it in `./docs/github-action.yml` (I don't really want it running on every commit here)
* Add `overwrite: true` to the upload artifact action so that it doesn't accumulate a bunch of old iso builds.
* Document the action in comments at the top of the action
* Start with a known good kickstart, either the examples in `./docs/` or the official fedora spin kickstarts after flattening them. For the example it *should* be possible to just use the one from `./docs/fedora-livemedia.ks` instead of making a new one.

Cameronsplaze:
> Don't put the action into the repo workflow, put it in `./docs/github-action.yml` (I don't really want it running on every commit here)

The only trigger I have at the top is `workflow_dispatch`, so it can only run if you manually trigger it from this console GUI. And personally if I'm looking for action examples in repos, the first place I check is the `.github/workflows` dir, or seeing the list of actions in the `Actions` tab above. Would you still like it moved?

> Add `overwrite: true` to the upload artifact action

Good idea, done!

> Document the action in comments at the top of the action

I added something real basic, but wasn't sure what else to add. Is there any extra info/clarifying you'd like added?

> Start with a known good kickstart ... `./docs/fedora-livemedia.ks`

Cool that one works! I deleted the example I added. Originally I tried the `fedora-minimized.ks` file, but it doesn't have `dracut-live` installed (along with something else I forget), so I just assumed the examples were for lorax itself. I completely missed that one, thanks!

.github/workflows/example-livemedia-creator.yml
```diff
@@ -0,0 +1,79 @@
...
+            --iso-name Fedora-custom-example.iso \
+            --project Fedora \
+            --volid "Fedora-${{ inputs.fedora_release }}" \
+            --releasever "${{ inputs.fedora_release }}" \
+            --resultdir ./Results
+
+      - name: Upload ISO
+        uses: actions/upload-artifact@v4.3.1
+        with:
+          name: Fedora-custom-example.iso
```
bcl, line 64: I suppose including the fedora_release in the filename would be nice.

bcl (changes requested):
Looks pretty good, can you squash all the commits into one clean one with a description of the action and the variables. And add the release to the output filename. Then I think it will be good. Thanks!

.github/workflows/example-livemedia-creator.yml
Cameronsplaze: I'm down to try it. I was just worried it would be the literal "Fedora-custom-example-${{ inputs.fedora_release }}.iso" instead of "Fedora-custom-example-39.iso", and I'm out of upload quota to test that part.

Cameronsplaze:
> Can you squash all the commits into one clean one with a description of the action and the variables

Is this something I can do? Everything I can find says the option appears when you click the merge button. If it doesn't, there's some settings you can adjust under the branch protection rules to make it appear and/or force it to happen every time.

> ...one with a description of the action and the variables

If my theories right, here's something you can modify/paste into the box that appears on merging

```txt
Added example of how to use livemedia-creator in github actions, and still build in fedora.

Inputs:
- fedora_release: Both the release to build, and build on. Since we use '--no-virt', they must be the same.
- kickstart_path: The path to the kickstart file to build from.
```

bcl:
No, you should squash things locally then force-push to your branch. Something like this should work:

```
git rebase master
git rebase --interactive 55f8077e6649f7e4df6bf55c5306b87faa6d0149
```
You will be presented with a 'menu' of commits to apply and an action to take. Change all but the first to 'squash', save the file, and then edit the commit message -- it will have the text from all the squashed commits included, just change it to what you posted above.

Then you should be able to push to your branch, but using `--force` because it is changing the history instead of adding to it.
Github will then pick it up and change this PR.

Cameronsplaze:
I tried a few times but couldn't figure out why the squash wasn't working. I thought I was using the wrong commit, so I tried HEAD~<numer_of_commits> too, no luck. I updated my branch with your master a few times, so maybe those in the middle are what were added? Out of curiosity too, is there a down side to squishing on merge through GitHub's button?

Also if it works with you, I created a new branch real fast and just "created" the file in that. It's this PR: https://github.com/weldr/lorax/pull/1382. If it works, that's probably faster than trying to untangle that last commit above.

Thanks!

PR #1375, opened 2024-02-06
Some updates in template files regarding to the default fonts packages

tagoh:
Current font packages list in template is a bit outdated. This PR aims to update it.


bcl (approved):
Thanks!

PR #1379, opened 2024-02-14
dnf5: set releasever in the base object's vars

AdamWill:
With Fedora 40 now branched, openQA installer build tests started failing with an error:

  RuntimeError: cannot open file: (2) - No such file or directory [/etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-x86_64]

it seems like setting releasever via set_substitutions just does not fully work as expected, and libdnf goes looking for a file with a literal `$releasever` in it based on the gpgkey line in the repo file. Though it seems like the releasever substitution must work to *some* extent because we at least manage to resolve and download packages before the failure.

This seems to fix it, anyhow, and is based on code from libdnf5 itself in dnf5/main.cpp RootCommand::set_argument_parser() .

PR #1380, opened 2024-02-14
dnfbase: Fix substitutions

bcl:
Branching for Fedora 40 revealed a problem with using substitutions in the GPG key name, @AdamWill fixed this by setting them on the Base object which appears to be the correct way to do it.

Remove the per-repo kludge and set releasever and basearch on the Base variables so they are used everywhere.

### DIFF
--- a/src/pylorax/dnfbase.py
+++ b/src/pylorax/dnfbase.py
@@ -153,6 +153,10 @@ def get_dnf_base_object(installroot, sources, mirrorlists=None, repos=None,
     log.info("Using %s for module_platform_id", platform_id)
     conf.module_platform_id = platform_id
 
+    # Set variables used for substitutions
+    dnfbase.get_vars().set("releasever", releasever)
+    dnfbase.get_vars().set("basearch", basearch)
+
     # Add .repo files
     if repos:
         reposdir = os.path.join(tempdir, "dnf.repos")
@@ -209,14 +213,6 @@ def get_dnf_base_object(installroot, sources, mirrorlists=None, repos=None,
         log.error("No enabled repos")
         return None
     log.info("Using repos: %s", ", ".join(r.get_id() for r in rq))
-
-    # Add substitutions to all enabled repos
-    for r in rq:
-        # Substitutions used with the repo url
-        r.set_substitutions({
-                "releasever": releasever,
-                "basearch": basearch,
-        })
 
     log.info("Fetching metadata...")
     try:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Clone Repository"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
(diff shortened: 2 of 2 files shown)


bcl:
Fixes #1379

AdamWill:
I am not actually sure that setting basearch this way works. At least, the code in libdnf5 itself that sets releasever this way does *not* set basearch the same way, which seems like a warning flag. I'll try it out and see what happens.

AdamWill:
well, it seems to be working. I do kinda feel like I want to ask dnf5 folks what exactly is going on here and if this is the right way, though - @jan-kolarik @m-blaha ?

jan-kolarik:
> well, I tried it this way, and it seems to be working. I do kinda feel like I want to ask dnf5 folks what exactly is going on here and if this is the right way, though - @jan-kolarik @m-blaha ? edit: for more context, my original PR was #1379 which just added this approach for `releasever` but kept the substitutions , @bcl extended it to this.

The current situation looks as expected from the DNF point of view. You set up the variables, and then when the repositories are created, they are substituted with the real values in `sack.create_repos_from_reposdir()`.

However, I couldn't find proper documentation for this in the current API docs, so I've added it to our [existing issue](https://github.com/rpm-software-management/dnf5/issues/213), and we'll strive to provide better documentation soon.

AdamWill:
@jan-kolarik thanks! Just to be clear - when you say "The current situation looks as expected from the DNF point of view", do you mean the "current situation" as in "the changes in this pull request are correct"?

jan-kolarik:
> @jan-kolarik thanks! Just to be clear - when you say "The current situation looks as expected from the DNF point of view", do you mean the "current situation" as in "the changes in this pull request are correct"?

Yes, sorry for my cumbersome expressions 😛

src/pylorax/dnfbase.py
```diff
@@ -153,6 +153,10 @@ def sanitize_repo(repo):
     log.info("Using %s for module_platform_id", platform_id)
     conf.module_platform_id = platform_id
 
+    # Set variables used for substitutions
+    dnfbase.get_vars().set("releasever", releasever)
+    dnfbase.get_vars().set("basearch", basearch)
```
j-mracek, line 158: Looks like a good place for setting vars.

j-mracek:
> I am not actually sure that setting basearch this way works. At least, the two bits of code in dnf5 itself that set releasever this way - [here](https://github.com/rpm-software-management/dnf5/blob/e3e81756abf9ba97b33e08682ffdea95968b3fa0/dnf5/main.cpp#L505) and [here](https://github.com/rpm-software-management/dnf5/blob/e3e81756abf9ba97b33e08682ffdea95968b3fa0/dnf5daemon-server/session.cpp#L114) - do _not_ set basearch the same way, which seems like a warning flag. I'll try it out and see what happens.

The links refers to a commandline overrides from options. The patch looks OK.

bcl:
Can't figure out what's up with github. The tests pass, and I updated the checkout version like it suggested but it still says waiting for status. Merging it anyway.

PR #1381, opened 2024-02-16
Require lorax-templates-rhel when building for ELN, CentOS Stream and RHEL

sgallagher:


bcl (changes requested):
Is this needed for Fedora 40 or just rawhide?
No need to change the version numbers and commit message, I'm going to be doing a new build today or Monday. Just add the new condition.

bcl (approved):
Thanks!

PR #1382, opened 2024-02-18
Added example of how to use livemedia-creator in github actions

Cameronsplaze:
Added example of how to use livemedia-creator in github actions, and still build in fedora.

Inputs:
- fedora_release: Both the release to build, and build on. Since we use '--no-virt', they must be the same.
- kickstart_path: The path to the kickstart file to build from.


bcl (approved):
Looks good, thanks!

PR #1384, opened 2024-03-05
templates.d/99-generic/runtime-cleanup: Do not purge sound packages

Conan-Kudo:
Filtering out these packages causes the build to fail when porting Anaconda to Wayland for the boot.iso.

This is needed for https://github.com/rhinstaller/anaconda/pull/5498


Conan-Kudo:
CC: @M4rtinK @jexposit

PR #1385, opened 2024-03-06
runtime-cleanup: Do not purge pam binaries

jexposit:
In order to migrate from X.Org to Wayland [1], GNOME Kiosk needs to be run in a new PAM session.

After an update in systemd, the session is failing to start because the PAM binaries are not found with the following error:

    Starting user@0.service - User Manager for UID 0...
    pam_unix(systemd-user:account): helper binary execve failed: No such file or directory
    pam_unix(systemd-user:account): helper binary execve failed: No such file or directory
    PAM failed: Authentication service cannot retrieve authentication info
    user@0.service: Failed to set up PAM session: Operation not permitted

Do not remove the PAM binaries in /usr/sbin to avoid this issue.

[1] https://github.com/rhinstaller/anaconda/pull/5485

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -293,7 +293,7 @@ removefrom openssh /usr/libexec/*
 removefrom openssh-clients /etc/ssh/* /usr/bin/ssh-*
 removefrom openssh-clients /usr/libexec/*
 removefrom openssh-server /etc/ssh/* /usr/libexec/openssh/sftp-server
-removefrom pam /usr/sbin/* /usr/share/locale/*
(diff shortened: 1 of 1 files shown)


bcl (approved):
Looks ok to me. FWIW I'd prefer to do a new lorax build after whatever other changes are needed are in, instead of a new build for each PR. So you may need to patch locally until everything is sorted out.

bcl:
Tests still failing because new python3-coveralls hasn't hit all the mirrors yet (I assume).

jexposit:
Sure, for the GNOME Remote Desktop work @M4rtinK is doing, he will have to open another PR to preserve `libgsm.so*` and he'll need https://github.com/weldr/lorax/pull/1384 as well. I'll let you coordinate on building the new image.

PR #1386, opened 2024-03-06
Add prefixdevname support to the boot.iso

rvykydal:
Resolves: rhbz#2267227


rvykydal:
@bcl I think we can merge the PR, Anaconda seems to be ready for it: https://bugzilla.redhat.com/show_bug.cgi?id=2267227#c2

sgallagher:
When making changes to the lorax code that requires commensurate changes to the templates, it would be REALLY helpful if you could also send a merge request for https://src.fedoraproject.org/rpms/lorax-templates-rhel/

This broke our Fedora ELN composes in a non-obvious way.

PR #1388, opened 2024-03-21
runtime-cleanup: wget2-wget has replaced wget

yselkowitz:
wget2-wget is just symlinks to wget2, which provides the actual content. wget2 has no files in /etc though.

This is a follow-up to commit 9b703f53ef59601c8a0f362edae5b81bfe709a5a (#1370).

This is already in lorax-templates-rhel for EL10+: https://src.fedoraproject.org/rpms/lorax-templates-rhel/c/b78f09a8fd2a9fa58ecf3b3bac4e0b808ef968fa?branch=eln

/cc @AdamWill


bcl (approved):
Thanks!

PR #1395, opened 2024-05-21
Add prefixdevname to Anaconda image

msekletar:
Related: RHEL-30010

### DIFF
--- a/share/templates.d/99-generic/runtime-install.tmpl
+++ b/share/templates.d/99-generic/runtime-install.tmpl
@@ -134,6 +134,7 @@ installpkg tigervnc-server-module
 installpkg net-tools
 installpkg bridge-utils
 installpkg nmap-ncat
+installpkg prefixdevname
 
 ## hardware utilities/libraries
 installpkg pciutils usbutils ipmitool
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -348,7 +348,7 @@ class Lorax(BaseLoraxClass):
             for arg in user_dracut_args:
                 dracut_args += arg.split(" ", 1)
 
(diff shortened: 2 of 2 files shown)


bcl (changes requested):
Thanks for the PR, it needs a few changes:
* please don't edit the `lorax.spec`, that is done as part of the release process
* Refer the the Jira issue in the commit on a separate line as `Related: RHEL-30010`
* Also submit a template change to https://gitlab.com/redhat/centos-stream/rpms/lorax-templates-rhel which is the rhel/centos specific templates. No need to change anything else there, just the template and make sure the commit has the `Related` line.

msekletar:
> Thanks for the PR, it needs a few changes:
> 
> * please don't edit the `lorax.spec`, that is done as part of the release process

I've dropped spec changes

> * Refer the the Jira issue in the commit on a separate line as `Related: RHEL-30010`

Issue reference is now (and was previously) on the separate line. However, when I am looking at the commit in Github UI I don't see empty line (but it is there) between issue reference and subject line.

> * Also submit a template change to https://gitlab.com/redhat/centos-stream/rpms/lorax-templates-rhel which is the rhel/centos specific templates. No need to change anything else there, just the template and make sure the commit has the `Related` line.

Will do.

Thank you for the review.

msekletar:
> > * Also submit a template change to https://gitlab.com/redhat/centos-stream/rpms/lorax-templates-rhel which is the rhel/centos specific templates. No need to change anything else there, just the template and make sure the commit has the `Related` line.
> 
> Will do.

https://gitlab.com/redhat/centos-stream/rpms/lorax-templates-rhel/-/merge_requests/46

bcl (approved):
Looks good, thanks!

PR #1402, opened 2024-06-24
mkksiso: option --replace to replace arbitrary text in boot config.

sstark:
- It is a generalized version of the code that replaces the volume ID.

  - Test adjusted accordingly because signature changed slightly.

  - This can be used to augment/change the user visible boot menu to add more information about the kickstart installation that is going to happen.

  - Implemented for grub and isolinux, not s390 (don't know how to test).

### DIFF
--- a/src/bin/mkksiso
+++ b/src/bin/mkksiso
@@ -312,7 +312,7 @@ def GetCmdline(line, commands):
     return "", line
 
 
-def EditIsolinux(rm_args, add_args, new_volid, old_volid, tmpdir):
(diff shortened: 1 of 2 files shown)


bcl:
Thanks, this looks pretty good, but please also update it for S390 and don't worry about testing -- the volid change should be enough and the signature for the functions needs to be consistent.

sstark:
I added a commit on top for that, let me know if you would rather have that squashed into one.

bcl (approved):
Looks good, thanks!

PR #1403, opened 2024-06-28
Add erofs compressed rootfs support

bcl:
This adds support for creating the rootfs using the erofs compression format. https://erofs.docs.kernel.org/en/latest/quickstart.html

You can run it by checking out this branch from github (into /root/lorax in this example) and running:

```
#!/bin/bash

# Gather up the list of system repo files and use them for lorax
REPOS=$(for f in /etc/yum.repos.d/*repo; do echo -n "--repo $f "; done)
if [ -z "$REPOS" ]; then
    echo "No system repos found"
    exit 1
fi

PATH=/root/lorax/src/bin:/root/lorax/src/sbin:$PATH PYTHONPATH=/root/lorax/src/ \
/root/lorax/src/sbin/lorax --product="Fedora" --version=41  --release=41 \
--volid="Fedora-rawhide-test" \
--rootfs-type erofs \
--proxy http://proxy.notae.us:8123 \
$REPOS /var/tmp/lorax-fedora-iso/ || exit 1
```

There will be a boot.iso and an install.img in /var/tmp/lorax-fedora-iso/images/

The resulting boot.iso will boot and mount the erofs filesystem, but the overlay doesn't appear to work so it is read only. Dracut still needs some work to fully support this, but this should be a good starting point.

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -39,6 +39,7 @@ Requires:       isomd5sum
 Requires:       module-init-tools
 Requires:       parted
 Requires:       squashfs-tools >= 4.2
+Requires:       erofs-utils
 Requires:       util-linux
 Requires:       xz-lzma-compat
 Requires:       xz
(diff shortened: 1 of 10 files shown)


bcl:
This requires this dracut PR - https://github.com/dracut-ng/dracut-ng/pull/490

bcl:
Merged to master.

PR #1409, opened 2024-07-13
Prepare for the sbin merge

keszybz:
When packages are rebuilt in an environment with merged sbin, files move from /usr/sbin to /usr/bin. 'runtime-cleanup' was wreaking havoc on such installations, removing 'losetup', 'mkswap', and other crucial utilities. (https://bodhi.fedoraproject.org/updates/FEDORA-2024-3aafcac6a8#comment-3609189)

The rules for the change:
- for single names, /sbin/ is replaced by /*bin/. E.g. /usr/sbin/chpasswd → /usr/*bin/chpasswd. This is easy, because we want to match 'chpasswd' no matter where it is installed.
- in some cases, there was a glob, but the rpm only has files in /usr/sbin, so we can apply the same substition and match those files no matter where they are installed.
- in some other cases, there was a glob, but it only matched a single file in /usr/sbin, and the package has other files in /usr/bin. In those cases, we need to adjust the glob to be more specific, e.g. /usr/sbin/* → /usr/*bin/iconvconfig for glibc.

This should be safe with unmerged packages, and should unbreak builds on with merged packages.

### DIFF
--- a/share/templates.d/99-generic/runtime-cleanup.tmpl
+++ b/share/templates.d/99-generic/runtime-cleanup.tmpl
@@ -28,8 +28,8 @@ removepkg selinux-policy libselinux-utils
 append etc/selinux/config ""
 
 ## keep enough of shadow-utils to create accounts
-removefrom shadow-utils --allbut /usr/bin/chage /usr/sbin/chpasswd \
-                        /usr/sbin/groupadd /usr/sbin/useradd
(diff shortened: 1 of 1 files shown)


bcl (approved):
Looks ok, eventually I'd like to see the globs updated as packages are built.

keszybz:
Thanks.

https://src.fedoraproject.org/rpms/lorax-templates-rhel/pull-request/11 filed as requested.

PR #1421, opened 2024-09-16
livemedia: Add isomd5sum

bcl:
This is needed for the iso checksum check of the live media. On the boot.iso (and in Fedora) it is pulled in by anaconda-install-* packages, but since liveinst isn't supported on RHEL we don't use those. Add it to the kickstart to make sure the checksum service can run at boot time.

Resolves: RHEL-58896


jikortus (approved):
Thank you for the fix, @bcl! I can confirm that the media check (default boot option on the live media) works as expected now.

PR #1424, opened 2024-09-30
templates: Drop dnf install

bcl:
See https://src.fedoraproject.org/rpms/dnf/c/51ac29c

The dnf cmdline is no longer provided by dnf5. And anything needed similar functionality should have it as a requirement, not an installpkg line in the template.


bcl:
Fixed the commit message to be more accurate.

AdamWill:
per https://github.com/weldr/lorax/pull/1423#issuecomment-2383826319 this LGTM.

PR #1425, opened 2024-10-08
Fix --updates the updates image wasn't loaded

jkonecny12:
Seems that the original idea with autoloading in Anaconda is not working as it should. So to fix this and simplify everything, just use the same logic as what is used for the kickstart file.


jkonecny12:
Seems to me that the pylint issue is not related to this PR?

bcl:
> Seems to me that the pylint issue is not related to this PR?

Correct, looks like a bug in astroid that's fixed in the latest version (3.3.5) so we can ignore that.

bcl (approved):
ACK. Thanks!

PR #1429, opened 2024-10-09
Update template for anaconda webui to start as user

adamkankovsky:
This commit is about adding the pkexec command to the lorax build to make it possible to run anaconda webui as a nonroot user and creating a new user to run the installer under.

PR with patch: https://github.com/rhinstaller/anaconda/pull/5689


bcl (approved):
Looks good to me

PR #1436, opened 2024-11-22
runtime-cleanup: Newer glibc installs into /usr/lib64

bcl:
Prepare for this change so that it isn't unexpectedly deleted.


bcl:
Merged

PR #1444, opened 2024-12-13
mkksiso: Fix rebuilding the efiboot.img on some systems

### LINKED ISSUE #1442: mkksiso's RebuildEFIBoot() logic checks /EFI/BOOT/BOOT.conf, which doesn't exist on RHEL 9 installers
The RHEL/AlmaLinux 9 installers only provide `/EFI/BOOT/grub.cfg`, which appears to have the same contents as `/EFI/BOOT/BOOT.conf` on older OS versions. As a result, mkksiso does not correctly update `efiboot.img`.

Adding logic to copy `/EFI/BOOT/grub.cfg` to  `/EFI/BOOT/BOOT.conf` seems to work, but I'm unsure if this is the correct approach.

bcl:
Check for BOOT.conf or grub.cfg in /EFI/BOOT, and only raise an error in RebuildEFIBoot if the directory is missing.

Fixes #1442

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:rawhide
+FROM registry.fedoraproject.org/fedora:latest
 COPY test-packages .
 RUN dnf -y install $(cat test-packages) && touch /.in-container
 RUN useradd weldr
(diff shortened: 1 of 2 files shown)


bcl:
NOTE: Currently untested, but it might work :)

PR #1445, opened 2024-12-16
Remove leftovers from xorg drop

jkonecny12:
Forwardport of https://gitlab.com/redhat/centos-stream/rpms/lorax-templates-rhel/-/merge_requests/57


bcl (approved):
Thanks! tests failing because pylint/astroid are broken on py3.13 again so they can be ignored.

PR #1450, opened 2025-02-20
Switch RHEL10 default rootfs to erofs

bcl:
This switches the default to erofs, makes squashfs-tools a Recommends (need to find out if that is ok to do with an unsupported package), and checks for the required tools early so you don't waste time if they aren't installed.


bcl:
Deferred until RHEL11

PR #1453, opened 2025-03-10
runtime-postinstall: Remove systemd-gpt-auto-generator

bcl:
In some circumstances this causes a timeout when booting the installer iso. I have been able to reproduce it with a UEFI Secure Boot qemu test using the iso as -hda instead of -cdrom.

It is not needed for the installer environment so just remove the generator file. Note: This has no effect on the installed system.

Resolves: rhbz#2350605


keszybz (approved):
LGTM.

PR #1461, opened 2025-03-21
spec: update lorax-lmc-virt dependencies

yselkowitz:
This syncs changes from RHEL 10 into ELN.

* Use edk2-aarch64 on aarch64
* Use qemu-kvm on RHEL, not qemu
* Disable lorax-lmc-virt on RHEL ppc64le


bcl (approved):
Thanks, looks good!

PR #1463, opened 2025-03-26
runtime-install firmware package changes: drop now-unneeded exclusions related to package renames, leave out more audio/video firmwares, leave out a uboot firmware package, only do amd-ucode-firmware on x86_64

AdamWill:
See title and individual commit messages for details.


bcl (approved):
Looks good to me, moving in the right direction :)

PR #1467, opened 2025-03-29
runtime-postinstall: allow pipewire to run as root

AdamWill:
https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/17eaf83 prevented pipewire socket and service from running as root, but in the installer environment, we need them to, or else RDP install does not work.

Resolves: rhbz#2355207

### DIFF
--- a/share/templates.d/99-generic/runtime-postinstall.tmpl
+++ b/share/templates.d/99-generic/runtime-postinstall.tmpl
@@ -22,6 +22,13 @@ append etc/mdadm.conf "HOMEHOST <ignore>\n"
 remove etc/systemd/system/default.target
 symlink /lib/systemd/system/anaconda.target etc/systemd/system/default.target
 
+## Allow pipewire.service and pipewire.socket to run as root
+## https://bugzilla.redhat.com/show_bug.cgi?id=2355207
+mkdir etc/systemd/user/pipewire.service.d/
+install ${configdir}/systemd-allowroot.conf etc/systemd/user/pipewire.service.d/allowroot.conf
+mkdir etc/systemd/user/pipewire.socket.d/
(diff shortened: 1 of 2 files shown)


share/templates.d/99-generic/runtime-postinstall.tmpl
```diff
@@ -22,6 +22,13 @@ append etc/mdadm.conf "HOMEHOST <ignore>\n"
 remove etc/systemd/system/default.target
 symlink /lib/systemd/system/anaconda.target etc/systemd/system/default.target
 
+## Allow pipewire.service and pipewire.socket to run as root
```
bcl, line 25: I think it would be cleaner if we used `install` here instead of hard-coding them in the script.

bcl (changes requested):

share/templates.d/99-generic/runtime-postinstall.tmpl
AdamWill: install what? these files don't exist anywhere else. These are just little override snippets - the /usr/lib/systemd/user/pipewire.socket and pipewire.service in the pipewire package now have a line `ConditionUser=!root` that blocks them from running as root, these snippets override that.

I could put these snippets in a package and have lorax install them, but that seems like a lot of work...

bcl: `share/templates.d/99-generic/config_files/common` has a bunch of config files we install with `install`, take a look at https://github.com/weldr/lorax/blob/master/share/templates.d/99-generic/runtime-postinstall.tmpl#L62

AdamWill: ah, ok, cool, I'll look at doing it that way.

bcl (approved):
Looks good!

PR #1469, opened 2025-05-05
runtime-postinstall: Remove root password

bcl:
The setup package has recently (setup-2.15.0-23) started generating the /etc/shadow content instead of using it as the input. This has changed how system accounts are locked in a default install (root ends up being !unprovisioned and others use !*).

The replace command in runtime-postinstall.tmpl would not remove the root password in recent builds, leading to inst.sshd not working because root required a password that didn't exist.

This also expands the replace command tests to check for other possible ways to lock the root account and confirm they all result in an account with no password.

Resolves: rhbz#2364082


jkonecny12 (approved):
Looks great to me. Thanks!

M4rtinK (approved):
Looking good, thanks! :)

PR #1475, opened 2025-06-25
fix rootfs path in .bash_history and profile

bam80:
- For rpm-ostree systems, `/` is mounted to `/mnt/sysroot` instead, which is not the same as `/mnt/sysimage`.
- For usual systems, `/mnt/sysroot` is attached to the same file system as `/mnt/sysimage`, so it's indifferent.

So only `/mnt/sysroot` path suits for the both systems.

chroot to `/mnt/sysimage` on rpm-ostree systems produces an error:
```console
# HOME=/root chroot /mnt/sysimage bash -l -i
chroot: failed to run command ‘bash’: No such file or directory
```
More details:
https://github.com/rhinstaller/anaconda/blob/main/docs/developer/mount-points.rst

### DIFF
--- a/share/templates.d/99-generic/config_files/common/bash_history
+++ b/share/templates.d/99-generic/config_files/common/bash_history
@@ -6,6 +6,6 @@ tail -f /tmp/storage.log
 echo b > /proc/sysrq-trigger
 dmsetup table
 multipath -d
-HOME=/root chroot /mnt/sysimage bash -l -i
+HOME=/root chroot /mnt/sysroot bash -l -i
 less /tmp/anaconda.log
 grep -v _yum_lock /tmp/packaging.log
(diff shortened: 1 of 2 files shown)


bcl:
Thanks! It looks like `share/templates.d/99-generic/config_files/common/profile` also points to sysimage when it sets `PATH`, I could make a separate PR for that, or you could add it to this one :)

bcl (approved):
Thanks!

supakeen:
> As was discovered recently Fedora IoT doesn't use Lorax for it's images. So I'm wondering if we need symmetrical fix there? Could you help with that? @bcl @supakeen

If you file a separate issue describing the problem/behavior at https://github.com/osbuild/images I can investigate and take a look.

bam80:
@supakeen 
https://github.com/osbuild/images/issues/1708

PR #1477, opened 2025-07-02
config_files: Do not remove `chcon` in runtime cleanup

elkoniu:
chcon is being used by the bootc and recently added to bootc.spec file as an explicite requirement: https://github.com/bootc-dev/bootc/pull/1391 When it is not presented in the system the result is failure of the `bootc install to-filesystem` command used by Anaconda installer.


jkonecny12 (approved):
Looks good to me.

jkonecny12:
offtopic: I just wonder, I see `/usr/bin/[` on the same line and it doesn't look as a smart idea to remove this to me. @bcl do you know why it's removed?

bcl (approved):
Thanks!

PR #1479, opened 2025-07-10
postinstall: Setup /var/tmp as tmpfs

bcl:
In some cases /var/tmp is too small for the ostreecontainer kickstart install method. Switch to using tmpfs so that it will be able to take advantage of larger amounts of ram if it is available.


bcl:
Closing for now, /tmp is already tmpfs so it should be used instead of adding complexity.

PR #1483, opened 2025-08-01
templates: Remove explicit installation of anaconda-widgets

KKoukiou:
anaconda-widgets is a dependency of anaconda-gui and does not need to be listed separately.


bcl (approved):
Ack!

PR #1485, opened 2025-08-07
templates: Synchronize with lorax-templates-rhel

bcl:
Resolves: RHEL-54712


bcl:
This sync actually needs to go the other direction.

PR #1487, opened 2025-08-12
tests: Add some unicode to the executils tests

bcl:
Trying to debug what is going on with this report of unicode output causing runcmd to crash:

https://bodhi.fedoraproject.org/updates/FEDORA-2025-eb260cb695

Simply adding unicode to the output doesn't cause an error.

### DIFF
--- a/src/pylorax/__init__.py
+++ b/src/pylorax/__init__.py
@@ -205,7 +205,7 @@ class Lorax(BaseLoraxClass):
         logger.addHandler(sh)
 
     def init_file_logging(self, logdir, logname="pylorax.log"):
-        fh = logging.FileHandler(filename=joinpaths(logdir, logname), mode="w")
+        fh = logging.FileHandler(filename=joinpaths(logdir, logname), mode="w", encoding="UTF-8")
         fh.setLevel(logging.DEBUG)
         logger.addHandler(fh)
 
@@ -476,7 +476,7 @@ def setup_logging(logfile, theLogger):
     logger.addHandler(sh)
     theLogger.addHandler(sh)
 
-    fh = logging.FileHandler(filename=logfile, mode="w")
+    fh = logging.FileHandler(filename=logfile, mode="w", encoding="UTF-8")
     fh.setLevel(logging.DEBUG)
     fmt = logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s")
     fh.setFormatter(fmt)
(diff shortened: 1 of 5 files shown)


AdamWill:
I don't think this is correct. Per the subprocess docs, setting `encoding` implies setting `text` (aka `universal_newlines`):

"If encoding or errors are specified, or text is true, file objects for stdin, stdout and stderr are opened in text mode using the specified encoding and errors or the [io.TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) default. The universal_newlines argument is equivalent to text and is provided for backwards compatibility. By default, file objects are opened in binary mode."

So we ought *not* to set `encoding` when `binary_output` is `True`.

We can fix it (only) in `_run_program` *relatively* easily with something like this:
```
        partproc = partial(
            startProgram(
                argv,
                root=root,
                stdin=stdin,
                stdout=subprocess.PIPE,
                stderr=stderr,
                env_prune=env_prune,
                env_add=env_add,
                reset_handlers=reset_handlers,
                reset_lang=reset_lang
            )
        )
        if binary_output:
            proc = partproc()
        else:
            proc = partproc(encoding="utf-8")
```
but that doesn't fix it in `startProgram` if used directly, so I kinda don't like it. But to fix it in `startProgram` if used directly we'd have to check whether `text` *or* `universal_newlines` *or* `encoding` are already set in `kwargs`, I guess...

The other things in the upstream doc text I quoted are interesting too, noticeably the thing about the default encoding being the TextIOWrapper default. That says "In [UTF-8 Mode](https://docs.python.org/3/library/os.html#utf8-mode), this defaults to UTF-8. Otherwise, it defaults to [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding)." So that implies Python is *not* in "UTF-8 Mode" in our openQA environment at least?

bcl:
Nothing in the lorax codebase sets binary_output to true, so in the codepaths other than `execReadlines` or calling `startProgram` directly it is always setting universal_newlines. In `execReadlines` it does its own decode("utf8").

The template `runcmd` uses execWithCapture which calls `_run_program` which will set encoding to UTF-8

I only have the podman test environment (which I cannot get to fail) and my VM which does fail. This fixes it in the VM and doesn't make the tests fail in new ways.

Initially I did try setting encoding in `startProgram` but that makes the other codepaths that expect binary results fail in the tests.

I'd appreciate it if someone can try these patches with the new glib2 in other environments and make sure it works. I *think* it will, but since I can only reproduce it manually in my VM I'm not 100% sure.

AdamWill:
well, if `binary_output` is never used, maybe remove it? if we just go with this patch, it's a big footgun.

I didn't mean that this patch wouldn't fix the specific bug we're dealing with - it likely would - but that it might cause other problems (the `binary_output` issue) or not fix similar bugs on slightly different codepaths (the 'call `startProgram` directly' case).

PR #1488, opened 2025-08-18
support product version to contain blank character in templates.d/99-generic

DustMerlin:
When I use " lorax -v 'XX XX' ",finally grub.conf is not in accordance with expectation. So I find the ${product.version} need **'**

### DIFF
--- a/share/templates.d/99-generic/efi.tmpl
+++ b/share/templates.d/99-generic/efi.tmpl
@@ -36,7 +36,7 @@ ${make_efiboot("images/efiboot.img")}
     %endif
     install ${configdir}/grub2-efi.cfg ${eficonf}
     replace @PRODUCT@ '${product.name}' ${eficonf}
-    replace @VERSION@ ${product.version} ${eficonf}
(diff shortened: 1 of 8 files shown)


DustMerlin:
Could you elaborate on the details regarding $releasever? I'm not entirely familiar with it, as I don't have much experience with it in practice. What is the specific mechanism for mapping Lorax parameters to repository URLs?

bcl:
If you pass a dnf repo file to lorax using `--repo` and the file has the standard baseurl or mirrorlist entry in it, a `$releasever` with spaces will only work if the server has support for it. eg.

```
[fedora]
name=Fedora $releasever - $basearch
#baseurl=http://download.example/pub/fedora/linux/releases/$releasever/Everything/$basearch/os/
metalink=https://mirrors.fedoraproject.org/metalink?repo=fedora-$releasever&arch=$basearch
```

Fedora only uses a number as the release and version strings, so I'm assuming you are doing something custom :)

bcl (approved):
Looks good, thanks!

PR #1489, opened 2025-08-26
Drop all installs of shim-ia32 (it no longer exists)

AdamWill:
There is no shim-ia32 in shim-15.8-4, which was recently tagged for f44. This is because it was built after
https://github.com/rhboot/efi-rpm-macros/commit/5c7fcc9 , so it appears to be an intentional change - @vathpela may confirm.


vathpela:
Yeah - I don't believe it works or has in some time, and I don't think there is any actual user base either.  Certainly Fedora installs haven't worked in 2+ years.

PR #1491, opened 2025-09-15
mkksiso: make ISO reproducible with SOURCE_DATE_EPOCH

champtar:
When SOURCE_DATE_EPOCH is present, pass it to xorriso and also force
the files timestamps to make the output ISO reproducible. Tested on EL 10.


champtar:
@bcl thanks for the pointer, I've now added a test, please let the ci run

bcl (approved):
Looks good, thanks!

PR #1495, opened 2025-10-02
runtime-install: drop tigervnc

AdamWill:
anaconda dropped VNC support in
https://github.com/rhinstaller/anaconda/pull/5829 . There is no point to having these any more, and they pull Xorg and its deps into the image unnecessarily.


AdamWill:
This saves about 20M in openQA testing. Should be enough to get the aarch64 image under 1.2G again.

PR #1496, opened 2025-10-09
Do not remove SELinux from the runtime

elkoniu:
* Do not remove SELinux packages when ISO is created
* Update kernel parameters for with `selinux=1 enforcing=0`


jstodola (approved):
It looks good to me.

bcl:
> Hey @bcl just checking if this will be headed to the rhel10 refresh as well?

not at this time. That depends on Anaconda requesting a backport and filing an issue for it.

PR #1504, opened 2026-02-05
fix: split in reverse order when constructing the `*-logos` package name

MatthieuCoder:
When using releases containing more than two "-", this code breaks by splitting at the wrong delimiter.

Example:

With 'my-super-os-release', lorax is going to search for 'my-logos';
With this change lorax will search for 'my-super-os-logos' instead.


MatthieuCoder:
> Thanks, but we can't do that. We have packages that include the variant name (eg. fedora-release-workstation)

Could this be an option instead ?

bcl:
Lorax already has more than enough options. The best way to control this is package naming.

PR #1513, opened 2026-04-01
findkernels: Enable support for 64k kernel flavor

bcl:
In order to support new kernel flavors we need to add it to the list of them in treebuilder. The kernel will then be included in the kernels list passed to the templates, allowing it to be used with installkernel and installinitrd.

Resolves: RHEL-163728


bcl:
Replaced by #1514

PR #1516, opened 2026-04-27
lorax: Fall back to erofs if mksquashfs is not installed

brlane-rht:
This switches the requirement on squashfs-tools to a 'Recommends' so that if it is missing lorax will still be installed.

It checks that mksquashfs is available when the rootfs-type is squashfs or squashfs-ext4, if it is not it falls back to rootfs and checks to make sure mkfs.erofs is available.

The default rootfs-type is still squashfs, and can still be overridden using --rootfs-type=erofs.

Resolves: RHEL-68936

### DIFF
--- a/lorax.spec
+++ b/lorax.spec
@@ -42,7 +42,7 @@ Requires:       gzip
 Requires:       isomd5sum
 Requires:       module-init-tools
 Requires:       parted
-Requires:       squashfs-tools >= 4.2
+Recommends:     squashfs-tools >= 4.2
 Requires:       erofs-utils >= 1.8.2
 Requires:       util-linux
 Requires:       xz-lzma-compat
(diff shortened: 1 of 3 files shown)


brlane-rht:
@M4rtinK @yselkowitz what do you think?

yselkowitz:
erofs is supposed to be the first choice nowadays:  https://fedoraproject.org/wiki/Changes/EROFSforLiveMedia

So the spec file change makes sense, but I'm not sure about the code.  If  `--rootfs-type=squashfs` is specified but `mksquashfs` is not present, shouldn't that be an error?  Or does that only make sense if `erofs` is the default, which currently it isn't?  And why isn't erofs the default now?

brlane-rht:
That change is for kiwi, not lorax/livemedia-creator. releng still hasn't switched the boot.iso to erofs. So I'm trying to keep things working as-is, but also nicely deal with removing squashfs-tools in ELN.
Lorax doesn't set the policy, if someone wants to keep using squashfs they should be able to, until the package is dropped.

PR #1519, opened 2026-07-23
Increase root partition to 10 GiB for a live medium

jikortus:
The current value of 8 GiB is not sufficient anymore to build a live medium using livemedia.ks (see https://redhat.atlassian.net/browse/RHEL-214304).

I'm not sure if it would be worth increasing the size also on RHEL-10 and Fedora as a preventive measure, but at least RHEL-10 is not affected this problem.

### DIFF
--- a/docs/livemedia.ks
+++ b/docs/livemedia.ks
@@ -30,7 +30,7 @@ clearpart --all --initlabel
 rootpw rootme
 # Disk partitioning information
 reqpart
-part / --size=8000
+part / --size=10240
 
 %post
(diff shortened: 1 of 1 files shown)


brlane-rht (changes requested):
We bumped RHEL10 to 10240 for RHEL-142732 so we may as well do the same here.
Also, the Resolves just needs the RHEL-214304 not the full url.

jikortus:
Sure, thank you. Updated.

brlane-rht (approved):
Ack!

PR #1521, opened 2026-08-04
Keep the fb kernel module

jstodola:
The fb kernel module is needed for graphical installations on s390x in KVM. The module dropped from RHEL-10 initrd.img due to dracut switching the drm dracut module with simpledrm (RHEL-188254). Keep the fb kernel module in install.img.

Related: RHEL-224409


brlane-rht (approved):
Ack!

PR #1525, opened 2026-08-12
Handle relocation of packaged repo configs to /usr/share/dnf5

AdamWill:
In Fedora 45+, packaged repo configs are moved to /usr/share/dnf5:
https://fedoraproject.org/wiki/Changes/RelocateRpmRepoConfigsToUsr
This adjusts lorax appropriately. We drop the renaming of the repo directory entirely because it is no longer needed. It was originally introduced in https://github.com/rhinstaller/anaconda/commit/c51389518415d9182f01757314c7cfa272534557 to solve a problem involving upgrades. anaconda has not handled upgrades since 2012. anaconda will be updated to not expect this rename any more.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,6 @@ endif
 export TEST_OS
 VM_IMAGE=$(CURDIR)/test/images/$(TEST_OS)
 
-ifeq ($(REPOS_DIR),)
-REPOS_DIR = /etc/yum.repos.d
-endif
-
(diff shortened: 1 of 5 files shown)


AdamWill:
anaconda PR: https://github.com/rhinstaller/anaconda/pull/7233

brlane-rht (approved):
Looks good, thanks!

